PR #3, opened 2015-08-19
Start implementing API

pgrinaway:
Hi!
I have some empty classes with docstrings, as well as the base `.travis.yml` and dependency installation (but I still need to write the tests). This can be merged now if that makes it easier, but there are some caveats:

* `GeometryEngine`'s API is incomplete
* `NCMCSwitching`'s API still needs to be implemented. 
* Need to implement travis tests

I will actually have some time this afternoon to work on those things. So you can merge now if you'd like, and I'll make another PR, or we can wait a bit.


alchemical_elimination.py
```diff
@@ -0,0 +1,24 @@
...
+        pass
+
+    @property
+    def alchemical_system(self):
+         """
+         Returns a system with the unique atoms alchemically modified.
+         
+         Returns
+         -------
+         alchemical_system : simtk.openmm.System object
```
jchodera: Do we want to standardize the names of the `Context` parameters that should be switched here?  Do we want to have one universal `lambda` parameter, or maybe anything beginning with `lambda_` gets switched from 1 to 0 along some route?

calibration.py
```diff
@@ -0,0 +1,55 @@
...
+import simtk.openmm as openmm
+import simtk.openmm.app as app
+import simtk.unit as units
+import joblib
+
+class Calibrator(object):
+
+    """
+    Given a ligand system, generate new sets of charges
+    representing fluorination states, as well as biases for the expanded ensemble simulation in implicit solvent.
```
jchodera: Get rid of `fluorination states`.

This is presumably the scheme that generates initial guesses for the log weights?  Might be useful to update the docs for this.

calibration.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+
+    def __init__(self, ligand_oemol, ligand_prmtop, ligand_inpcrd, openmm_platform='CPU'):
```
jchodera: Do we need to generalize this besides ligands, or does the `Calibrator` class only apply to ligands?

flourine_scan.py
```diff
@@ -0,0 +1,61 @@
+__author__ = 'Patrick B. Grinaway'
+
+import simtk.openmm as openmm
+import simtk.openmm.app as app
+import gaff2xml
+import openeye.oechem as oechem
+from copy import deepcopy
+import sklearn.externals.joblib.memory as memory
+
+class FluorineScan(object):
```
jchodera: Not sure if this one is relevant anymore?

calibration.py
pgrinaway: Ah, this thing snuck in from a while back. I'll get rid of it.

pgrinaway:
Had a couple of extraneous files from a while back. Just eliminated them.

alchemical_elimination.py
pgrinaway: Hm, good point. Perhaps standardizing to `lambda_` would be a good idea. I'm not sure what, in the general case, this will be.

pgrinaway: Not sure whether I should add this as a note in the docstring, or try to enforce it somehow via the architecture.

pgrinaway:
Also, question: do we want `NCMCSwitching` base class to just accept `alchemical_system`, `alchemical_protocol`, and `initial_positions`? This would mean something like:
```python
#alchemical elimination of old atoms
old_alchemical_engine = AlchemicalEliminationEngine(old_system, topology_proposal)
old_alchemical_system = old_alchemical_engine.alchemical_system
ncmc_elimination = NCMCEngine(old_alchemical_system, elimination_alchemical_protocol, \ old_initial_positions)
ncmc_elimination.step() #maybe a better name than that
ncmc_old_positions = ncmc_elimination.positions
geometry_engine = GeometryEngine(topology_proposal, ncmc_old_positions)
geometry_proposal = geometry_engine.propose()
new_alchemical_engine = AlchemicalEliminationEngine(new_system, topology_proposal)
new_alchemical_system = new_alchemical_engine.alchemical_system
ncmc_introduction =  NCMCEngine(new_alchemical_system, introduction_alchemical_protocol, geometry_proposal.new_positions)
ncmc_introduction.step()
...
```

The alternative would be to have the NCMCEngine API perform the dimension switch itself, but I feel this is less general. On the other hand, with a merged topology, this is unnecessarily verbose.

Another point of curiosity here: should some of these things be `staticmethods`? That would avoid repeated instantiation (although a quick survey seems to indicate that instantiation is pretty fast relative to the other stuff we're doing here).

Thoughts?

jchodera:
> Also, question: do we want NCMCSwitching base class to just accept alchemical_system, alchemical_protocol, and initial_positions?  

That looks pretty nice to me!  

We may also need a method to retrieve the NCMC contribution to the acceptance probability.

> Another point of curiosity here: should some of these things be staticmethods? That would avoid repeated instantiation (although a quick survey seems to indicate that instantiation is pretty fast relative to the other stuff we're doing here).

I think if it makes sense to make a lot of these things objects to reduce mess, that would be better than static methods.

> `ncmc_introduction.step()`

Maybe we can change this to `.integrate()` instead, since it may run multiple steps?

> `ncmc_old_positions = ncmc_elimination.positions` 

We may also want to distinguish between initial positions and final positions for the NCMC scheme.

pgrinaway:
Ok, I've had a chance to work on this a bit more. I've mostly completed the API (I think...). You can take a look at what an iteration might look like in `run_example.py` (right now it obviously does nothing). I realized I need to create a base class for calculating the weight of a state, so that is next on the menu. I began to add travis tests, but that might be better for another PR once we decide this is ok (or if it's not, how we should change it). Thoughts, @jchodera @juliebehr ?

jchodera:
Checking this out now!

jchodera:
Any chance you could have `run_example.py` run, say, 10 iterations?  It's unclear from the current example which objects need to be created each step of the sampling process, and which ones persist through the whole sampling run.

pgrinaway:
Ok, I changed it to run 10 iterations. A couple things to keep in mind:

* As you can see, nearly every object is created on each iterations. We could certainly avoid that by having a `dict` that (say, for small molecules) would allow something like:
```python
if new_smiles_string not in alchemical_systems.keys():
     #...create new alchemical system
    alchemical_systems[new_smiles_string] = new_alchemical_system
```
and the like. I can add that in to the example later, but I wanted to quickly get something up now.

jchodera:
Tagging @juliebehr to take a look at this API.

jchodera:
Here is the example driver script in the PR that illustrates what this is like to use:
https://github.com/pgrinaway/expanded-ensembles/blob/master/run_example.py

pgrinaway:
Ok, so I've added the base API. There are a couple questions that I have at the moment.

* `GeometryEngine` takes `TopologyProposal` and `current_positions`:
 * This is doable, but may duplicate effort. Should it take a `System` as well?
 * This question relates to the implementation of `GeometryEngine`

* More generally than the above, what is the best way to deal with the confluence of topologies, parameters, and positions?
 * ParmEd seems nice, but it seems to only read files, and I'm not sure we want to have to write out files.
 * Is the `openmm.app.Topology` object useful here? (I'll look into this)
 * We can directly use the `openmm.System` object, and figure out the most efficient way to query it

On the ParmEd note, actually, writing (temporary) files is probably unavoidable at the moment due to the use of  `antechamber`, so maybe something like `TopologyProposal` -> `prmtop` [via antechamber] -> ParmEd -> openmm would actually be the easiest pipeline.

Thoughts, @juliebehr @jchodera?

jchodera:
Checking this out now!

jchodera:
Just checked out your `master` branch, but can't run the example:
```
mski1776:expanded-ensembles.pgrinaway choderaj$ python run_example.py 
  File "run_example.py", line 27, in <module>
    log_weight = state_weight_calculator.log_weight
AttributeError: 'StateWeight' object has no attribute 'log_weight'
```

pgrinaway:
Oops!

jchodera:
Any chance we could add some comments to the `run_example.py` to delineate the stages of the calculation too?

pgrinaway:
Yep, doing that now!

jchodera:
```
mski1776:expanded-ensembles.pgrinaway choderaj$ python run_example.py 
  File "run_example.py", line 29, in <module>
    new_system = new_system_generator.system
AttributeError: 'SystemGenerator' object has no attribute 'system'
```

jchodera:
It runs!
```
mski1776:expanded-ensembles.pgrinaway choderaj$ python run_example.py 
Proposal(old_topology=<Topology; 0 chains, 0 residues, 0 atoms, 0 bonds>, new_topology=<Topology; 0 chains, 0 residues, 0 atoms, 0 bonds>, logp=0.0, new_to_old_atom_map={0: 0})
<simtk.openmm.openmm.System; proxy of <Swig Object of type 'OpenMM::System *' at 0x10550e360> >
[ 0.  0.  0.]
0
[ 0.  0.  0.]
0
0.0
...
```

jchodera:
OK, I'll make a pull request against your `master`, @pgrinaway

pgrinaway:
Ok, sounds good.

pgrinaway:
While adding the comments (and doing some additional brainstorming), it strikes me that the order of things might be better differently. For instance, if I have a `SystemGenerator` create a `complex_system`, `ligand_system`, and `complex_structure` [ParmEd] or something like that, then the `StateWeight` could use the system that I've already generated, and the `GeometryEngine` could use the ParmEd structure.

jchodera:
I cleaned up the example and added lots of questions/suggestions as comments.

I also think we want to use an idiom where we create an object once and then repeatedly call a method during individual steps, when sensible to do so.  There's not really a point to creating an object and then calling a method once unless we need to access different aspects of the same thing repeatedly.

pgrinaway:
Ah! sorry for the clutter in this thread--I was trying to make line notes. Overall, I agree that objects are created a bit gratuitously, especially now that it is laid out. Perhaps, as we discussed a bit earlier (and I noted in the line notes) some of these things should become static methods? 

One of the main advantages I saw in using objects, even if there was a bit much instantiation, was that it becomes very easy to enforce inheritance/polymorphism/etc, and that perhaps some cleaner code would be worth the overhead.

pgrinaway:
Also, potentially for discussion tomorrow, what is the feeling about implementing `__call__`? It seems a bit odd perhaps, but then we could have something like
```python
transformation = Transformation(proposal_metadata)
new_proposal = transformation(sampler_state)
```
since a bunch of these classes only have one method anyway.

jchodera:
What's wrong with more traditional idioms?
```
myobject = MyClass(information)
...
new_result = myobject.doTheThing(old_result)
```
This could still let us use inheritance for classes, but we don't necessarily need to use static class method, only use objects once, or overload `__call__`.

jchodera:
```
[LSKI1497:choderalab/expanded-ensemble/expanded-ensembles.pgrinaway] choderaj% python run_example.py 
  File "run_example.py", line 114, in <module>
    run()
  File "run_example.py", line 42, in run
    transformation = Transformation(old_topology, old_system, metadata)    
TypeError: __init__() takes exactly 2 arguments (4 given)
```

pgrinaway:
That's fine. I can remove the static method decorators when I get in. I was looking at other idioms because most of the objects had no state. However, it might be useful to add state in derived classes, and static methods don't play as nicely with inheritance so I think you're right.

jchodera:
Agree completely!  Those objects may need state in derived classes for:
* storing what kinds of transformations to generate by specifying a set of
molecules in a library or wild type protein sequence
* tuning proposal probabilities to increase acceptance rates during burn in
* other things we haven't thought of yet

pgrinaway:
Fixing the example script to conform to the new API now. Can I move the imports to the top of the file? I agree that `from x import y` ends up looking cleaner.

jchodera:
> Can I move the imports to the top of the file?

Sure!

jchodera:
Is this getting close to merging?

pgrinaway:
Yeah. I have to rename a couple things + add the skeleton for the `BiasEngine`, which I'll do this afternoon. Then it should be good for merging.

jchodera:
Thanks!   I'd like to do some reorganization of the repo once things are tested and merged!

Go ahead and merge when ready.

pgrinaway:
Ok. The example runs on travis now, so I think I'm going to merge this.

PR #5, opened 2015-09-16
initial reorganization of repo

pgrinaway:
Started reorganizing the repo. Will fix example next (to accommodate the new organization) and add a setup.py.


jchodera:
Feel free to add an `attic/` to store things that don't seem to belong anywhere else.

pgrinaway:
Ok, that sounds good.

jchodera:
Looking better!  We're almost there!

Some observations about the current state:
* There are still some `.py` files at the top level directory; can we move those into appropriate subdirectories?
* We should move `multitopology/` into `perses/`.
* Add some description of the various subpackages to `README.md`
* Under `examples/`, let's have subdirectories `rjmc/`, `multitopology/`, `dualtopology/`, etc. so that the examples are segregated.  Also need a `README.md`.

jchodera:
Looking good!
* I think we still need a `setup.py`.  You can probably model this after the minimal one in [ThermoPyL](https://github.com/choderalab/thermopyl/blob/master/setup.py)
* The `devtools/` tree in [ThermoPyL](https://github.com/choderalab/thermopyl/tree/master/devtools) is also pretty modern and set up for the [docker travis environment](https://github.com/choderalab/thermopyl/blob/master/.travis.yml), so you may want to update to that too.

pgrinaway:
Ok. I'll add those after lunch then ping the thread to see if we're ready to merge

jchodera:
I'm hopping on a flight to London, so feel free to merge once you think things are as far along as they can get without updates from my end!

pgrinaway:
Going to merge. Will correct nosetest issue in a future PR (needs either test exclusion or openeye installation).

Issue #4, opened 2015-09-15
Plan going forward


pgrinaway:
Now that we have some base API, my imagination of the priorities are:

1a. Get the meat of instantaneous compound switching put together
1b. Get tests for this part in the repo and running on travis

2. Make conda-installable

3. Add in NCMC switching

Unless there are any objections, I'll start working on them in this order.

jchodera:
We should do a bit of prep work to get this in a shape where multiple people can contribute at once.

Here's what I think we need to do:
* Rename the repo to be a useful Python package name (`perses`?)
* Reorganize repo to create a directory/module structure, like:
```
perses/
   perses/
      annihilation/ - alchemical factories for annihilation
      dualtopology/ - direct dual-topology alchemical transformation
      multitopology/ - core-based multi-topology alchemical transformation
      rjmc/ - reversible-jump based methods
      tests/ - automated tests
   devtools/
   examples/
   setup.py
   travis.yml
   README.md
```
* Add basic documentation in `README.md` just to describe where stuff is
* Add `setup.py` to make this Python installable
* Add `devtools/` and `travis.yml` and enable automated testing [JDC can do this]
* Add `examples/` with most basic working example of each type

After that, I think we're all set to divide up the work and flesh out some details of the `rjmc` approach:
* geometry engine [PBG]
* small molecule topology proposal engine [PBG] based on simplest approach (picking similar compounds from a specified compound library using MCSS?)
* protein topology proposal engine [JMB]
* NCMC switching using alchemical protocols [JDC]

I'll also clean up the `multitopology` and `alchemical` scripts to the point where we can have Julie figure out how to use them for residue mutations.

jchodera:
Does this sound reasonable?

pgrinaway:
That definitely sounds reasonable to me. Do you want me to organize the repo as you described above, or did you want to do it as you indicated in the pull request? Either is fine with me, I just don't want to leave the repo in a weird state.

jchodera:
Can you take a stab at it?  Feel free to do some direct commits if that is easiest, and raise an issue if you want any input/feedback!

pgrinaway:
Sounds good

jchodera:
OK, we've stripped out the alchemical stuff and moved it to [`examol`](http://github.com/choderalab/examol).

We might need to add in some additional tests and automatic running of the examples to make sure they work.

After that, we can get to work on dividing up the actual implementation!

pgrinaway:
[...]

PR #7, opened 2015-09-21
Migrate multitopology alchemical stuff to examol

jchodera:
* Moved nosetests from conda recipe to `.travis.yml`
* Added encrypted `oe_license.txt` using the travis [encrypted files](http://docs.travis-ci.com/user/encrypting-files/) mechanism


jchodera:
* Changed travis-ci service to user `jchodera`

pgrinaway:
Hm, don't we need YANK for its alchemical stuff?

jchodera:
The plan in #4 was to have a separate `perses/annihilation/` module to handle alchemical annihilation.  That's probably the best plan for now, since everything can be self-contained.

I've actually been wanting to migrate `alchemy.py` *out* of YANK into its own repo so it can be more thoroughly tested on its own.  I wonder if it makes the most sense to have this in a separate `alchemy` repo or to put it here (and possibly rename `perses` to `alchemy`).  Alternatively, we can have an `alchemy` repo that also pulls in `dualtopology` and `multitopology`, and just keep `perses` for the RJMC stuff.

Thoughts?

pgrinaway:
>I've actually been wanting to migrate alchemy.py out of YANK into its own repo so it can be more thoroughly tested on its own. 

I think this makes a lot of sense. We may also have additional applications that need to be able to make alchemical modifications--there's no need to induce a dependency on YANK there.

As to the rest, I think it would actually make sense to have `perses` just contain the RJMC stuff, alchemy to be separate, and the multitopology stuff to be separate as well (since we have different collaborations for that). It might make planning for each of the projects a bit easier, too.

jchodera:
> As to the rest, I think it would actually make sense to have perses just contain the RJMC stuff, alchemy to be separate, and the multitopology stuff to be separate as well (since we have different collaborations for that). It might make planning for each of the projects a bit easier, too.

I think this sounds logical to me.

How about this for a plan:
* I'll set up a new `alchemy` repo as a new home for the working alchemical annihilation stuff from YANK.  Tests should all pass for this, and I can add even more testing since we can run travis for longer than when we had to test with the rest of YANK.
* I'll create a `multitopology` repo and migrate our (currently mostly broken) code from `perses` there.  We can work on getting that working again separately.
* That will leave the RJMC stuff inside `perses`, where all tests should pass.

pgrinaway:
Sounds good! What about putting multitopology stuff in `examol`?

Thanks!

jchodera:
> Sounds good! What about putting multitopology stuff in examol?

Great idea!  Working on that now.

jchodera:
Alchemy stuff is now going here, by the way: https://github.com/choderalab/alchemy

jchodera:
Multitopology stuff has now been migrated to [`examol`](http://github.com/choderalab/examol)

jchodera:
OK, looks like we're good again.  Merging.

Can you take over now, @pgrinaway?

pgrinaway:
Yep, sounds good!

PR #8, opened 2015-09-29
Implementing geometry engine

pgrinaway:
Beginning of initial implementation of direct switching between compounds/mutations. Currently getting together the `FFGeometryEngine` (so named because it uses forcefield data to propose geometries), followed by the proposal system for small molecules and then a script to run it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,8 @@ install:
 script:
   - conda config --add channels https://conda.binstar.org/${ORGNAME}
   - conda build devtools/conda-recipe
+  - conda install --yes --quiet pip
+  - pip install $OPENEYE_CHANNEL openeye-toolkits
   # Install and test (in separate directory)
   - conda install --yes --quiet --use-local ${PACKAGENAME}-dev
   # Run tests
@@ -31,7 +33,7 @@ script:
   - mkdir test && cd test && nosetests perses --nocapture --verbosity=2 --with-doctest
 env:
   matrix:
-    - python=2.7  CONDA_PY=27
+    - python=2.7  CONDA_PY=27 OPENEYE_CHANNEL="-i https://pypi.anaconda.org/openeye/channel/main/simple"
     #- python=3.3  CONDA_PY=33
     #- python=3.4  CONDA_PY=34
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,7 @@ env:
 
   global:
     - ORGNAME="omnia"
+    - OE_LICENSE="$HOME/oe_license.txt"
     - PACKAGENAME="perses"
     # encrypted BINSTAR_TOKEN for push of dev package to binstar
     - secure: "RRvLDPu9mPoNaRWIseaJdgShOXI+PaHPWKAIJvW7VYWcAS6iEN7W4Fj4zD5hkocQxc3ou97EtkgID+ApH10bSGKxCykyU0urSY9jsSOJX2m0AE19X0dVr6ySIQkwCWE6kUMVlvQYQo80fM2EMElD+btr4G9XBAhArAO7HvZHkoQ="
--- a/attic/geometry_attic.py
+++ b/attic/geometry_attic.py
@@ -1,0 +1,90 @@
+__author__ = 'patrickgrinaway'
+"""
+Old stuff goes here
+"""
+    def _cartesian_to_internal(self, atom, bond_atom, angle_atom, torsion_atom, positions):
+        """
+        Convert the cartesian coordinates of an atom to internal
+        """
+        a = positions[bond_atom] - positions[atom]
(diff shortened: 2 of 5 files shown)


pgrinaway:
Currently brainstorming ways of efficiently traversing the molecular graph to decide the order of position assignment.

jchodera:
The simplest approach here is to repeatedly go through the list of torsions and see if you can place atoms 1 or 4 in the 1-2-3-4 torsion list.  You can place atom 4 if positions for 4 are undefined, but you have positions for atoms 1, 2, and 3.  Similarly, you can place tom 1 if positions for 1 are undefined, and atoms 2, 3, and 4 have defined positions.  There are only O(N) torsions, so this scales well.

You can later make this more efficient by pruning the list of torsions used for this purpose if neither atoms 1 or 4 have undefined positions.

You can get the list of torsions from the OpenMM `System` object, `PeriodicTorsionForce` object.

pgrinaway:
sounds good! Thanks!

jchodera:
I forgot to mention that this will not work the same way for improper torsion since the bond and angle definitions will be between different atoms. you can probably just ignore impropers, but if you need to, there should be a way to create an equivalent atom placement scheme for impropers.

jchodera:
Is this PR ready to be merged yet?

pgrinaway:
I need to fix the computation of the determinant of the jacobian here.

pgrinaway:
Ok, current status:
* I couldn't really get `sympy` to do what I want.
* I went back and looked at the code that uses autograd -- turns out I had made a mistake. I fixed that, and now autograd provides jacobians whose determinants are not 0. So, I think this is now working. I suppose the hairy question is now how to test it...

jchodera:
> I suppose the hairy question is now how to test it...

There are a few tests that come to mind:
* Checking invariants.  For example, J(r,theta,phi) should be independent of phi for any (r,theta)
* Running a Monte Carlo simulation in (r,theta,phi) with a simple form of p(r,theta,phi) that should produce a uniform distribution in (x,y,z) within a sphere.  You might have to use something like p(r,theta,phi) \propto [r^2 sin(theta)] to get a uniform distribution.

Come to think of it, isn't the Jacobian determinant just `J(r,theta,phi) = r^2 sin(theta)`?

pgrinaway:
Ok, I can do those tests. The first in particular is pretty easy.

>Come to think of it, isn't the Jacobian determinant just `J(r,theta,phi) = r^2 sin(theta)`?

So, the same as spherical coordinates? Hm, maybe.

pgrinaway:
Empirically it seems to be the same (albeit with the opposite sign)

jchodera:
> Empirically it seems to be the same (albeit with the opposite sign)

The Jacobian determinant should definitely be positive.  `r^2 sin(theta)` should also be positive in the allowed domain `theta \in [0,\pi]`.

pgrinaway:
Well, the acceptance criterion uses the absolute value of dett(J), no? 

Point taken about the allowable domain of `theta`, though. I'll take a look.

jchodera:
> Well, the acceptance criterion uses the absolute value of det(J), no?

Ah, good point!  So it doesn't matter.

pgrinaway:
I think maybe it'd be good to get this merged and implement tests in the next PR? I've also got a couple other components about ready for PRs (proposal, biasing engine), so it might help to clean up.

jchodera:
Sure, that works.

Travis is currently failing on some imports (`numexpr` and `oechem`).  That means we'll have to add these as dependencies, right?

pgrinaway:
oh, let me add those before merging then!

jchodera:
If you want to incorporate the `openeye` tools into testing, here is an example from [openmoltools](https://github.com/choderalab/openmoltools):
* license file encryption: [encryption](http://docs.travis-ci.com/user/encrypting-files/) [travis changes](https://github.com/choderalab/openmoltools/blob/master/.travis.yml#L24-L25) (note that this should probably be done within the actual `choderalab/perses` checkout with a direct push because of the possibility for accidentally encrypting for your local branch instead)
* Installation and testing: [travis block](https://github.com/choderalab/openmoltools/blob/master/.travis.yml#L37-L43)
* License file environment variable: [travis block](https://github.com/choderalab/openmoltools/blob/master/.travis.yml#L56-L58)

pgrinaway:
Ok. I'll add that in after lunch and then merge when the tests pass.

pgrinaway:
I think the license file is actually already installed, but cannot be decrypted in a pull request. Otherwise, the `openeye` toolkits now install on travis.

jchodera:
Correct---the license can't be decrypted in a PR from your fork---only branches in this repo.  The only alternative is to skip those tests in PRs.

If it is ready to go, feel free to merge!

PR #10, opened 2015-10-17
[WIP] Started draft of NCMC deletion/insertion

jchodera:
I've started a draft of the NCMC deletion/insertion, but have a few questions about how we want `NCMCEngine` to work:
* Is the expectation that `NCMCEngine` is an abstract base class and we will implement subclasses that would have different behaviors?
* We also need the temperature to compute log acceptance probability contributions; perhaps we want to pass a `ThermodynamicState` to the constructor of these objects?
* The 'alchemical_protocol' isn't the same as the YANK discrete alchemical state protocol: 
Here, it's a time-dependent NCMC switching protocol for the alchemical parameters, which are encoded by a dict of switching functions, like `{ 'lambda_sterics' : 'lambda', 'lambda_angles' : 'lambda^2' }`, the number of switching steps `nsteps`, and the `timestep` for velocity Verlet steps.
Do we want the `alchemical_protocol` to store this information, or should the switching protocol be  encoded by which subclass of `NCMCEngine` we select?
* Do we want to also handle creation of the alchemical system here, or keep that separate?
* Does it make sense to split this into object creation followed immediately by integration,
or does it make more sense to configure the NCMCEngine with the switching functions at the beginning and then have 'integrate' start from the alchemical_system and initial_positions and return the `final_positions` and `log_ncmc`?
That would look more like this:
```python
# Initialization of run
ncmc_deletion  = NCMCEngine(thermodynamic_state, alchemical_deletion_protocol)
ncmc_insertion = NCMCEngine(thermodynamic_state, alchemical_insertion_protocol)
...
# Inside the main loop
...
[final_deletion_positions, log_ncmc_deletion] = ncmc_deletion.integrate(alchemical_deletion_system, initial_deletion_positions)
# do trans-dimensional RJMC here
[final_insertion_positions, log_ncmc_insertion] = ncmc_insertion.integrate(alchemical_insertion_system, initial_insertion_positions)
...
```

I'm also making a few changes to the `alchemy` package to support the alchemical switching of bonds, angles, and torsions in the way we discussed.

### DIFF
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -16,16 +16,16 @@ requirements:
     - openmm >=6.3
     - numpy
     - scipy
-    #- openmoltools
-    #- alchemy
+    - openmoltools
+    - alchemy
 
   run:
     - python
     - openmm >=6.3
     - numpy
     - scipy
-    #- openmoltools
-    #- alchemy
+    - openmoltools
+    - alchemy
 
 test:
   requires:
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -16,16 +16,16 @@ requirements:
     - openmm >=6.3
     - numpy
     - scipy
-    - openmoltools
-    - alchemy
+    - openmoltools-dev
+    - alchemy-dev
 
   run:
     - python
     - openmm >=6.3
     - numpy
     - scipy
-    - openmoltools
-    - alchemy
+    - openmoltools-dev
+    - alchemy-dev
 
 test:
   requires:
--- a/devtools/travis-ci/install.sh
+++ b/devtools/travis-ci/install.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
+pushd .
 
 # Install Miniconda
+cd $HOME
 MINICONDA=Miniconda-latest-Linux-x86_64.sh
 MINICONDA_MD5=$(curl -s http://repo.continuum.io/miniconda/ | grep -A3 $MINICONDA | sed -n '4p' | sed -n 's/ *<td>\(.*\)<\/td> */\1/p')
 wget http://repo.continuum.io/miniconda/$MINICONDA
@@ -8,10 +10,11 @@ if [[ $MINICONDA_MD5 != $(md5sum $MINICONDA | cut -d ' ' -f 1) ]]; then
     echo "Miniconda MD5 mismatch"
     exit 1
 fi
-bash $MINICONDA -b
+bash $MINICONDA -b -p miniconda
 
 # Configure miniconda
 export PATH=$HOME/miniconda/bin:$PATH
-conda install --yes conda-build jinja2 conda-server pip
-conda config --add channels http://conda.binstar.org/omnia
+conda install --yes conda-build jinja2 anaconda-client pip
+conda config --add channels omnia
 
+popd
--- a/examples/rjmc/run_example.py
+++ b/examples/rjmc/run_example.py
@@ -9,8 +9,8 @@ This could represent sampling of small molecules, protein mutants, or both.
 
 
 import numpy as np
-import simtk.openmm as openmm
-import simtk.openmm.app as app
+from simtk import unit, openmm
+from simtk.openmm import app
 from perses.rjmc.topology_proposal import Transformation, SamplerState
 from perses.bias.bias_engine import BiasEngine
 from perses.annihilation.alchemical_engine import AlchemicalEliminationEngine
@@ -22,6 +22,11 @@ def run():
     # Create initial model system, topology, and positions.
     # QUESTION: Do we want to maintain 'metadata' as well?
     # QUESTION: What if we incorporate the tuple (system, topology, metadata, positions) into a SamplerState object, or a named tuple?
+
+    # Run parameters
+    temperature = 300.0 * unit.kelvin # temperature
+    pressure = 1.0 * unit.atmospheres # pressure
+    collision_rate = 5.0 / unit.picoseconds # collision rate for Langevin dynamics
 
     # Specify alchemical elimination and introduction protocols.
     alchemical_elimination_protocol = dict()
@@ -47,11 +52,11 @@ def run():
     alchemical_engine = AlchemicalEliminationEngine(alchemical_metadata)
 
     #Initialize NCMC engines.
-    sitching_timestep = 1.0 * unit.femtoseconds
-    switching_nsteps = 10
-    switching_functions = {
+    switching_timestep = 1.0 * unit.femtosecond # timestep for NCMC velocity Verlet integrations
+    switching_nsteps = 10 # number of steps to use in NCMC integration
+    switching_functions = { # functional schedules to use in terms of `lambda`, which is switched from 0->1 for creation and 1->0 for deletion
         'alchemical_sterics' : 'lambda',
-        'alchemical_electrostatocs' : 'lambda',
+        'alchemical_electrostatics' : 'lambda',
         'alchemical_bonds' : 'lambda',
         'alchemical_angles' : 'lambda',
         'alchemical_torsionss' : 'lambda'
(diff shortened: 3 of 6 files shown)


jchodera:
The actual NCMC integrator is [here](https://github.com/choderalab/perses/pull/10/files#diff-8ea53d35fec62929093d48c801c10911R94), and is actually surprisingly simple!  I've added an option  to select whether we are deleting or inserting particles, in order to let us specify a single insertion alchemical protocol and use that for both directions if we want.

pgrinaway:
>Is the expectation that NCMCEngine is an abstract base class and we will implement subclasses that would have different behaviors?

That was my original thought.

>We also need the temperature to compute log acceptance probability contributions; perhaps we want to pass a ThermodynamicState to the constructor of these objects?

Ah! Good point! 

>Do we want to also handle creation of the alchemical system here, or keep that separate?

I was going to add a separate class for the creation of alchemical system.

* I do like your example of how we could do NCMC insertion and deletion.

* Regarding the switching protocol, upon further thought, I think it might be better for the `NCMCEngine` subclass to encode this information directly, since as you mentioned, this is different from the discrete `alchemical_protocol`

jchodera:
> I was going to add a separate class for the creation of alchemical system.

Did you intend the `AlchemicalEliminationEngine` to do this?  It seems to manufacture the alchemically-modified system from the unmodified system given a topology proposal that contains `new_to_old_atom_map`, `new_topology`, and `old_topology`.  I think that means it should be very easy to use the [`alchemy`](http://github.com/choderalab/alchemy) module to do this, so I can implement that---it should only take a few lines.

> Regarding the switching protocol, upon further thought, I think it might be better for the NCMCEngine subclass to encode this information directly, since as you mentioned, this is different from the discrete alchemical_protocol

I'm happy to go this route if you prefer!  But it might be easier to just have `alchemical_protocol` encode the `timestep`, `nsteps`, and `functions` dict if you want to make it more flexible than requiring swapping out subclasses in the code.

jchodera:
> I do like your example of how we could do NCMC insertion and deletion.

I'll go ahead and change the `NCMCEngine` and example to match the updated architecture:
```python
# Initialization of run
ncmc_deletion  = NCMCEngine(thermodynamic_state, alchemical_deletion_protocol)
ncmc_insertion = NCMCEngine(thermodynamic_state, alchemical_insertion_protocol)
...
# Inside the main loop
...
[final_deletion_positions, log_ncmc_deletion] = ncmc_deletion.integrate(alchemical_deletion_system, initial_deletion_positions)
# do trans-dimensional RJMC here
[final_insertion_positions, log_ncmc_insertion] = ncmc_insertion.integrate(alchemical_insertion_system, initial_insertion_positions)
...
```

pgrinaway:
Sounds good! I think `AlchemicalEliminationEngine` is what we want to do that, especially if it makes working with existing code easier.

jchodera:
OK, I've overhauled the `NCMCEngine` API to be more like we discussed above.

Can you check out the [updated example](https://github.com/jchodera/perses/blob/openeye-tests/examples/rjmc/run_example.py) to see if this looks good?

I'm still fixing up the [`alchemy`](http://github.com/choderalab/alchemy) package to support this, as well as debugging the example, but feedback on the updated API is still helpful at this point!

jchodera:
There seems to be a flaw in the `AlchemicalEliminationEngine` API since there is no way to tell which atom set we want to identify as the alchemical atoms---it's impossible to know if we want to select the atoms to be deleted or the atoms to be added.

The most conservative fix is to simply add another optional flag that specifies which branch (`create` or `delete`) we want:
```python
old_alchemical_system = alchemical_engine.make_alchemical_system(system, top_proposal, direction='delete')
...
new_alchemical_system = alchemical_engine.make_alchemical_system(new_system, top_proposal, direction='create')
```

pgrinaway:
I'll take a look at the rest of the API later, but does the direction matter from the `AlchemicalEliminationEngine`'s viewpoint? I had imagined that it would just make an alchemically modified system, and whether the atoms were being deleted or created would be determined at the NCMC stage. 

On the other hand, if that's unclear or messy, and the `AlchemicalEliminationEngine` would be better off doing that, I'm ok with it.

jchodera:
> I'll take a look at the rest of the API later, but does the direction matter from the `AlchemicalEliminationEngine`'s viewpoint? I had imagined that it would just make an alchemically modified system, and whether the atoms were being deleted or created would be determined at the NCMC stage.

The two alchemical transformations are different.  Recall we have three distinct sets of atoms:
* Set 1: atoms in both old system and new system
* Set 2: atoms in the old system that do not exist in the new system
* Set 3: atoms in the new system that do not exist in the old system

In the first transformation, we delete set 2.  In the second transformation, we delete set 3.  The atom numberings for these sets (if they exist) are also different depending on whether we are talking about the old system or the new system.

If we just call `alchemical_engine.make_alchemical_system(system, top_proposal`, there is no way to know which set (set 2 or 3) we are talking about or whether we are referring to the numbering of that set in the old or new system.



In the first case, we're alchemically modifying atoms that exist in `old_system` but do not exist in `new_system`, but the new atoms in `new_system` aren't present in `old_system`.  In the second case, we're alchemically modifying

jchodera:
I've made the appropriate API modifications, but I can't go any further to actually test since the `run_example.py` example actually has zero atoms in the proposed transformation.

Did you have a working example running, where atoms are actually being nominated for creation/deletion?  If so, could you create a PR to merge this in?  Once we get that merged, I can continue testing to make sure this works.  In the meantime, I'll focus on doctests and nosetests of the independent components---I just can't go any further with the `run_example.py` example.

pgrinaway:
Ah, yes, that is true. I suppose I was wondering whether it would also be convenient (and forgot that we don't do this at the moment) to give the AlchemicalElimination just a system and an atom map specific to that system, where its atoms are "new." However, I think that is confusing and perhaps just having the extra parameter would be easiest.

pgrinaway:
Yeah, let me clean it up and add it to the PR later so that we build a working example.

jchodera:
> Ah, yes, that is true. I suppose I was wondering whether it would also be convenient (and forgot that we don't do this at the moment) to give the AlchemicalElimination just a system and an atom map specific to that system, where its atoms are "new."

That's essentially what `AbsoluteAlchemicalFactory` does, but we need to generate that information somehow.  If this informational is already available, we can just dispense with the `AlchemicalElimination` class, but for now, it might be a fine layer of abstraction.

jchodera:
I've made the requisite changes to `alchemy` to support alchemically-softened bonds, angles, and torsions in [this PR](https://github.com/choderalab/alchemy/pull/3), but I am still tracking down some issues that prevent the nosetests from passing:
* With `openmm` 6.3, one of the test systems segfaults, which shouldn't happen; this may have been fixed recently, but...
* With `openmm-dev`, deep copying of forces is [fucked up](https://github.com/pandegroup/openmm/issues/1190), so I need to fix that first

Once those are working, I'll get the nosetests implemented for this PR.

jchodera:
The OpenMM `deepcopy` bug was fixed in https://github.com/pandegroup/openmm/pull/1211 but now there is some weird `segfault` bug with the CPU platform that I'm trying to track down.

jchodera:
Quick update: Tracking down an OpenMM bug that may be the cause of my `alchemy` test failures in https://github.com/pandegroup/openmm/issues/1214

Once that is solved, if `alchemy` tests pass, I will cut a new release and add unit tests and more doctests to this PR.

jchodera:
`alchemy` / `openmmtools` / `openmm` stuff is finally fixed and a new release of `alchemy` that supports softening bonds, angles, and torsions is out, so I'll be working to add tests to this PR next.

jchodera:
I'm ready to start implementing tests for this!  Any suggestions?  Does the `examples/run_example.py` work now, so that I can just add an NCMC variant of this and have it tested by nosetests?

I can also test the NCMC switching by itself on some very simple problems, like transforming one harmonic oscillator into another one over N steps, since the free energy difference can be estimate both analytically and from the resulting work distributions.

pgrinaway:
I think the NCMC test on simple problems is useful either way, since the quantities it produces can be compared to analytical results.

I think `run_example.py` will need to be modified a bit. I can do that.

jchodera:
Will merge and then @pgrinaway can fix example next.

Issue #11, opened 2015-10-25
Extend TopologyProposal with convenience methods to retrieve atoms being deleted and atoms being created


jchodera:
Perhaps we should make `TopologyProposal` a real class, since we could then define convenience methods that use `new_to_old_atom_map` along with `old_topology` and `new_topology` to return the lists of atoms being deleted or created.  This would be helpful for creating alchemically modified systems without having to recompute these lists.

pgrinaway:
I think this is a good idea. It might also make the proposal classes cleaner.

jchodera:
Do you want to take a stab at this?

pgrinaway:
Yep! Sounds good!

pgrinaway:
I'll make this the next PR after fixing the test failures.

pgrinaway:
Based on #22 and the discussion there, I've implemented a full class version of the `TopologyProposal`. I can open a PR now, though it will require a bit of a refactor to make the tests work (since the constructor is now different).

jchodera:
Go for it!

pgrinaway:
Implemented in #27

PR #12, opened 2015-10-27
[WIP] Topology proposal

pgrinaway:
This is a basic implementation of topology proposals with small molecules, where p(i | j) is proportional to tanimoto(i, j), which is symmetric. It currently lacks a map from the system/topology atoms to the small molecule atoms-- this may be a good reason to go with #11 and add more functionality to the proposal.

I'm going to write a basic script to run it as well, so that other parts can be tested.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,8 @@ install:
 script:
   - conda config --add channels https://conda.binstar.org/${ORGNAME}
   - conda build devtools/conda-recipe
+  - conda install --yes --quiet pip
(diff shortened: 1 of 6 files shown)


pgrinaway:
Hm, looks like I've imported the geometry stuff too...

jchodera:
Cool!

Would be good to get the geometry stuff in at some point, but maybe through a separate PR?

pgrinaway:
Ok, I think I'm going to close this PR since it somehow got tangled with the geometry stuff, and open a new one with the topology proposal stuff...

PR #13, opened 2015-11-02
[WIP] implementing topology proposal for small molecule switching

pgrinaway:
This is a simple topology proposal for sampling between different molecules. It still needs to be able to map atoms in a topology where there is more than just the small molecule.


jchodera:
Cool!

It's important to note that the concept of "Tanimoto similarity" only makes sense when you also specify which fingerprint bitstring is being compared.  Looks like you're using MACCS166.

Eventually, we may want to have a "plug-in" choice of similarity measure for this kind of "sampling over a bag of molecules" scheme.  I could also envision a scheme based on how many atoms need to be added/deleted compared to the common core to be much successful here, and eventually, we can train ML algorithms to predict thermodynamic length.  But for now, this is good!  Just think ahead to making things modular, if possible!

pgrinaway:
> the concept of "Tanimoto similarity" only makes sense when you also specify which fingerprint bitstring is being compared. Looks like you're using MACCS166.

Good point!

>Eventually, we may want to have a "plug-in" choice of similarity measure for this kind of "sampling over a bag of molecules" scheme.

I think this is ideal. I had envisioned that just being another subclass of `Transformation`, but I'm starting to see that in all cases there is duplicated code (atom mapping, for instance).

jchodera:
You could have a subclass hierarchy:
```
Transformation
--> SmallMoleculeTransformation
----> SmallMoleculeLibraryTransformation
------> TanimotoSmallMoleculeLibraryTransformation
------> LOMAPSmallMoleculeLibraryTransformation
------> MCSSSmallMoleculeLibraryTransformation
etc.
```
Obviously the names could be better, but this would mean you only have to implement the similarity computation in the sub-sub-classes.

jchodera:
Actually, you'd probably want to allow both the similarity computation and the atom mapping to be overridden, but compartmentalize as many useful helper functions as you can in `SmallMoleculeTransformation` to provide useful tools for simplifying that.

pgrinaway:
Yep, good point.

pgrinaway:
Ok, the biasing engine, as well as the combined proposal/system generators are implemented for small molecules. I'm going to merge this PR, then tomorrow morning fix the `run_example.py` to conform to the new API, and we'll have a working example!

PR #14, opened 2015-11-10
[WIP] Fixing run example, etc.

pgrinaway:
This PR fixes `run_example.py` to comply with the new API. It also tidies up some other miscellaneous details.

### DIFF
--- a/examples/rjmc/run_example.py
+++ b/examples/rjmc/run_example.py
@@ -14,12 +14,13 @@ from simtk.openmm import app
 import openeye.oechem as oechem
 import openeye.oeomega as oeomega
 import openmoltools
+import logging
 from perses.rjmc.topology_proposal import SingleSmallMolecule
 from perses.bias.bias_engine import BiasEngine
 from perses.annihilation.alchemical_engine import AlchemicalEliminationEngine
 from perses.rjmc.geometry import GeometryEngine
 from perses.annihilation.ncmc_switching import NCMCEngine
-from perses.rjmc.system_engine import SystemGenerator
+
 
 def generate_initial_molecule(mol_smiles):
     """
@@ -97,6 +98,7 @@ def run():
     system = initial_sys
     topology = initial_top
     positions = initial_pos
+    current_log_weight = bias_calculator.g_k('CC')
     for i in range(niterations):
         # Store old (system, topology, positions).
 
@@ -106,7 +108,7 @@ def run():
 
         # QUESTION: What about instead initializing StateWeight once, and then using
         # log_state_weight = state_weight.computeLogStateWeight(new_topology, new_system, new_metadata)?
-        log_weight = bias_calculator.generate_bias(top_proposal)
+        log_weight = bias_calculator.g_k(top_proposal.metadata['molecule_smiles'])
 
         # Perform alchemical transformation.
 
@@ -129,16 +131,16 @@ def run():
         print(ncmc_introduction_logp)
 
         # Compute total log acceptance probability, including all components.
-        logp_accept = top_proposal.logp + geometry_proposal.logp + ncmc_elimination_logp + ncmc_introduction_logp + log_weight - bias_calculator.g_k(top_proposal)
+        logp_accept = top_proposal.logp + geometry_proposal.logp + ncmc_elimination_logp + ncmc_introduction_logp + log_weight - current_log_weight
         print(logp_accept)
 
         # Accept or reject.
         if (logp_accept>=0.0) or (np.random.uniform() < np.exp(logp_accept)):
             # Accept.
-            (system, topology, positions, current_log_weight) = (new_system, top_proposal.new_topology, ncmc_new_positions, log_weight)
+            (system, topology, positions, current_log_weight) = (top_proposal.new_system, top_proposal.new_topology, ncmc_new_positions, log_weight)
         else:
             # Reject.
-            pass
+            logging.debug("reject")
 
 #
 # MAIN
--- a/examples/rjmc/run_example.py
+++ b/examples/rjmc/run_example.py
@@ -6,8 +6,7 @@ Example illustrating use of expanded ensembles framework to perform a generic ex
 This could represent sampling of small molecules, protein mutants, or both.
 
 """
-import sys
-sys.path.append("/Users/grinawap/PycharmProjects/perses")
+
 
 import numpy as np
 from simtk import unit, openmm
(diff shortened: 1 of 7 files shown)


pgrinaway:
Ok, I have seriously messed up my python installation here. Please hold...

pgrinaway:
@jchodera, Have you ever seen the `AttributeError: type object 'object' has no attribute '__getattr__'` problem with `CustomIntegrator`? For a minimal example, if I just create:

```python
import simtk.openmm as openmm
import simtk.unit as units
class TestIntegrator(openmm.CustomIntegrator):
    def __init__(self):
        super(TestIntegrator, self).__init__(2*units.femtoseconds)
        self.addGlobalParameter("thing", 3)

integrator = TestIntegrator()
```

I get the `AttributeError` above. This is now also happening for the `NCMCAlchemicalIntegrator` class (though I don't know what prompted it). I had thought perhaps I just messed up my python install, but I'm not sure. If you get it too, then I will raise an issue on the openmm tracker. Otherwise, I'll try to fix my python install.

pgrinaway:
Ok, as @maxentile pointed out, I was using the wrong method name. Oops...

pgrinaway:
Ok, the example now runs. However, with more than 2 NCMC steps, all positions are `nan`.

pgrinaway:
Ok, I'm wondering if the issue here is somewhere in the `NCMCAlchemicalIntegrator`:

* I commented out the computation of the `lambda` parameter in the integrator. This still resulted in `nan`s.

* I ran the `old_alchemical_system` for a few hundred steps with just a `LangevinIntegrator` or `VerletIntegrator`. The potential energy and positions were fine after this, so the system itself is not broken.

pgrinaway:
Have now also tried:

* `VelocityVerlet` `CustomIntegrator` from openmmtools--this works fine.

* An unrolled version of the above--this works fine.

pgrinaway:
Ok, I think I found what breaks the simulation.

This works fine:
```python

class VelocityVerletIntegrator(openmm.CustomIntegrator):

    """Verlocity Verlet integrator.

    Notes
    -----
    This integrator is taken verbatim from Peter Eastman's example appearing in the CustomIntegrator header file documentation.

    References
    ----------
    W. C. Swope, H. C. Andersen, P. H. Berens, and K. R. Wilson, J. Chem. Phys. 76, 637 (1982)

    Examples
    --------

    Create a velocity Verlet integrator.

    >>> timestep = 1.0 * simtk.unit.femtoseconds
    >>> integrator = VelocityVerletIntegrator(timestep)

    """

    def __init__(self, timestep=1.0 * unit.femtoseconds):
        """Construct a velocity Verlet integrator.

        Parameters
        ----------
        timestep : numpy.unit.Quantity compatible with femtoseconds, default: 1*simtk.unit.femtoseconds
           The integration timestep.

        """
        nsteps = 10
        super(VelocityVerletIntegrator, self).__init__(timestep)

        self.addPerDofVariable("x1", 0)

        # Constrain initial positions and velocities.
        self.addConstrainPositions()
        self.addConstrainVelocities()

        self.addUpdateContextState()
        self.addComputePerDof("v", "v+0.5*dt*f/m")
        self.addComputePerDof("x", "x+dt*v")
        self.addComputePerDof("x1", "x")
        self.addConstrainPositions()
        self.addComputePerDof("v", "v+0.5*dt*f/m+(x-x1)/dt")
        self.addConstrainVelocities()

        for step in range(nsteps):

            self.addUpdateContextState()
            self.addComputePerDof("v", "v+0.5*dt*f/m")
            self.addComputePerDof("x", "x+dt*v")
            self.addComputePerDof("x1", "x")
            self.addConstrainPositions()
            self.addComputePerDof("v", "v+0.5*dt*f/m+(x-x1)/dt")
            self.addConstrainVelocities()
```

This does causes the `nan`s:
``python
class VelocityVerletIntegrator(openmm.CustomIntegrator):

    """Verlocity Verlet integrator.

    Notes
    -----
    This integrator is taken verbatim from Peter Eastman's example appearing in the CustomIntegrator header file documentation.

    References
    ----------
    W. C. Swope, H. C. Andersen, P. H. Berens, and K. R. Wilson, J. Chem. Phys. 76, 637 (1982)

    Examples
    --------

    Create a velocity Verlet integrator.

    >>> timestep = 1.0 * simtk.unit.femtoseconds
    >>> integrator = VelocityVerletIntegrator(timestep)

    """

    def __init__(self, timestep=1.0 * unit.femtoseconds):
        """Construct a velocity Verlet integrator.

        Parameters
        ----------
        timestep : numpy.unit.Quantity compatible with femtoseconds, default: 1*simtk.unit.femtoseconds
           The integration timestep.

        """
        nsteps = 10
        super(VelocityVerletIntegrator, self).__init__((nsteps+1) * timestep)

        self.addPerDofVariable("x1", 0)

        # Constrain initial positions and velocities.
        self.addConstrainPositions()
        self.addConstrainVelocities()

        self.addUpdateContextState()
        self.addComputePerDof("v", "v+0.5*dt*f/m")
        self.addComputePerDof("x", "x+dt*v")
        self.addComputePerDof("x1", "x")
        self.addConstrainPositions()
        self.addComputePerDof("v", "v+0.5*dt*f/m+(x-x1)/dt")
        self.addConstrainVelocities()

        for step in range(nsteps):

            self.addUpdateContextState()
            self.addComputePerDof("v", "v+0.5*dt*f/m")
            self.addComputePerDof("x", "x+dt*v")
            self.addComputePerDof("x1", "x")
            self.addConstrainPositions()
            self.addComputePerDof("v", "v+0.5*dt*f/m+(x-x1)/dt")
            self.addConstrainVelocities()
```

So, it appears that calling `super(VelocityVerletIntegrator, self).__init__((nsteps+1) * timestep)` with that timestep (though I understand that is actually "time elapsed") is causing `nan`s. Will investigate to see what happens when I remove it on the NCMC class w/ switching

**Comment by jchodera** [comment]
> So, it appears that calling super(VelocityVerletIntegrator, self).__init__((nsteps+1) * timestep) with that timestep (though I understand that is actually "time elapsed") is causing nans. Will investigate to see what happens when I remove it on the NCMC class w/ switching

That must be setting the quantity `dt` used in the integrator, where `timestep = 1 fs` would work fine but `(nsteps+1) * timestep` would cause the system to explode.  I think we should be using `timestep` instead of `(nsteps+1) * timestep` in that call:
https://github.com/choderalab/perses/blob/master/perses/annihilation/ncmc_switching.py#L207

I wonder though how we advance the integration clock by the appropriate amount if we are using an integrator where the timesteps have been unrolled.  That might be a good question for the OpenMM issue tracker.

**Comment by jchodera** [comment]
Relevant:
your head a splode

**Comment by jchodera** [comment]
Oh, duh. I should have done something like this and defined a new internal timestep:
```
super(VelocityVerletIntegrator, self).__init__((nsteps+1) * timestep)

self.addPerDofVariable("dt_ncmc", timestep.value_in_unit_system(unit.md_unit_system))
```
and then use `dt_ncmc` instead of `dt` in the integration.

jchodera:
Let me know if you want me to make a PR to do this (into which I can also add some tests), or if you want to tackle it.

pgrinaway:
Ok, the example now runs, but there are a couple outstanding issues, which will be resolved in a future PR:

* Better geometry proposals

* NCMC switching tests

PR #20, opened 2015-11-12
fixing geometry engine test

pgrinaway:
Updating the geometry engine test.


jchodera:
The error here is
```
Error: No packages found in current linux-64 channels matching: alchemy-dev
```
Can you switch the recipe from `alchemy-dev` to `alchemy`?

pgrinaway:
I'm going to merge this. It's the openeye license thing that's failing, and I just realized it's because I obviously can't access the license from a PR.

PR #24, opened 2015-11-13
Ensure meta.yaml and setup.py have consistent requirements

jchodera:
This ensures `meta.yaml` and `setup.py` have consistent requirements, which hopefully fixes the travis issue.


pgrinaway:
huh, looks like theres still an issue with the license? are we sure that this one is the current license?

jchodera:
Nope, the problem is still getting my decryption logic to work correctly:
```
$encrypted_514505261e50_iv -in oe_license.txt.enc -out oe_license.txt -d; fi
/home/travis/build.sh: line 45: [: choderalab/perses: integer expression expected
```

jchodera:
OK, tests are working properly now, though @pgrinaway needs to fix a couple of them.

PR #26, opened 2015-11-13
[WIP] Topology proposal class

pgrinaway:
Ok, I've introduced a new class that contains the relevant information that we need from the proposal. I've also created a single subclass for small molecules, that adds a `molecule_smiles` property. I think this might be cleaner than using a `dict`, since it enforces what the property name should be. I've still left the `metadata` in, in case we need to otherwise use it.

### DIFF
--- a/perses/rjmc/geometry.py
+++ b/perses/rjmc/geometry.py
@@ -3,6 +3,7 @@ This contains the base class for the geometry engine, which proposes new positio
 for each additional atom that must be added.
 """
 from collections import namedtuple
+import perses.rjmc.topology_proposal as topology_proposal
 import numpy as np
 import scipy.stats as stats
 import numexpr as ne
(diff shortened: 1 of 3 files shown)


pgrinaway:
Also, perhaps I'll try to shorten the name of the subclass...

jchodera:
> Also, perhaps I'll try to shorten the name of the subclass...

I actually like the descriptive name.  I think it's fine.

jchodera:
OK, that's a less than useful failure message when creating a PR from a fork.  Not sure how to fix that.

Maybe for future work, use a branch of the `choderalab/perses` repo?  Should be fine to test locally for now.

pgrinaway:
Oops, yeah. Ok. I'll test locally then

pgrinaway:
Unable to merge, not sure what happened here. Superseded by #27

Issue #32, opened 2015-11-14
Add contribution where we perturb from unmodified System objects to alchemical versions


jchodera:
For large systems, it will become critical to include the contribution where we perturb between the physical `System` objects and alchemically-modified versions of the `System` object.

This is just a reminder to myself to include this into the NCMC scheme.

jchodera:
It looks like this will require a change in the interface.

We currently have `run_example.py` call code like this (which will presumably be moved into a `Sampler` class):
```python
old_alchemical_system = alchemical_engine.make_alchemical_system(system, top_proposal, direction='delete')
[ncmc_old_positions, ncmc_elimination_logp] = ncmc_engine.integrate(old_alchemical_system, positions, direction='delete')
...
new_alchemical_system = alchemical_engine.make_alchemical_system(top_proposal.new_system, top_proposal, direction='create')
[ncmc_new_positions, ncmc_introduction_logp] = ncmc_engine.integrate(new_alchemical_system, geometry_proposal.new_positions, direction='create')
```
Here are some possibilities:
* **Option 1.** Have `AlchemicalEngine.make_alchemical_system()` also return the `logP` for changing the system from real to alchemical.  This would require additional arguments (`temperature`).
* **Option 2.** Have `NCMCEngine.integrate()` also incorporate the `logP` for changing the system from real to alchemical. This would require additional arguments (`real_system`).
* **Option 3.** Combine `AlchemicalEngine` with `NCMCEngine` and merge these calls into a single call:
```
ncmc_engine = AlchemicalNCMCEngine(temperature=temperature, timestep=switching_timestep, nsteps=switching_nsteps, functions=switching_functions)
...
[ncmc_old_positions, ncmc_elimination_logp] = ncmc_engine.integrate(top_proposal, positions, direction='delete')
...
[ncmc_new_positions, ncmc_introduction_logp] = ncmc_engine.integrate(top_proposal, geometry_proposal.new_positions, direction='create')
```
I actually like Option 3 even if it means a bit more refactoring, since it reduces the number of arguments that must be specified multiple times and eliminates some intermediate variables (`old_alchemical_system`, `new_alchemical_system`) that are never used again.  I'm not sure if there are really any disadvantages to this.

pgrinaway:
Options 2 and 3 are both fine, IMO. 

Re: option 2, I think that now that we have a convenience class for topology proposals that include both new and old systems, requiring `real_system` is not really a burden at all.

[...]

PR #33, opened 2015-11-14
Fix travis nosetests

jchodera:
This PR will address #32.

Fixed some nosetest failures and warnings first.


pgrinaway:
I was actually getting a PR ready to address this (some commits are in my fork), but I think we can merge this when ready and I can add my additional commits later. I changed the `AlchemicalEliminationEngine.make_alchemical_system` to use the new `TopologyProposal` class.

jchodera:
> I was actually getting a PR ready to address this (some commits are in my fork), but I think we can merge this when ready and I can add my additional commits later. 

Sorry about this!  I had somehow missed this.

Usually best to open a PR early to stake out what is being worked on...

Issue #34, opened 2015-11-14
run_example.py dies with recursion error


jchodera:
```
[LSKI1497:perses.choderalab/examples/rjmc] choderaj% python run_example.py 
LOG: 
Warning: run_antechamber has been moved to openmoltools.amber.
LOG: antechamber -i in.mol2 -fi mol2 -o out.mol2 -fo mol2 -s 2 -c bcc
...
  File "/Users/choderaj/anaconda/lib/python2.7/site-packages/perses-0.1-py2.7.egg/perses/rjmc/topology_proposal.py", line 131, in metadata
    return self.metadata
RuntimeError: maximum recursion depth exceeded while calling a Python object
```

pgrinaway:
Hm, oops! Let me fix this...

jchodera:
Now we have
```
Traceback (most recent call last):
  File "run_example.py", line 229, in <module>
    run()
  File "run_example.py", line 173, in run
    log_weight = bias_calculator.g_k(top_proposal.metadata['molecule_smiles'])
TypeError: 'NoneType' object has no attribute '__getitem__'
```
Suggest:
1. Make `run_example.py` into a nosetest `perses/tests/test_sampler.py` so we can make sure it works
2. Use branches of `choderalab/perses` for PRs so nosetests on travis will work
3. Eventually convert the test to use the `Sampler` driver subclasses when they are ready

pgrinaway:
> Make run_example.py into a nosetest perses/tests/test_sampler.py so we can make sure it works

We should do this, but given the API changes and the fact that we'll be creating a new set of `Sampler` classes, I was going to deprecate the current `run_example.py`, as you point out in 3. This is something that I was going to do next.

> Use branches of choderalab/perses for PRs so nosetests on travis will work.

Ok. I can switch to that, but I think this should be a temporary measure. Eventually, I think the code should either no longer require closed-source software (more time-consuming, and a low-priority task), or should use pre-generated files in the tests, which avoids the requirement for closed-source software in tests.

pgrinaway:
This was resolved.

PR #36, opened 2015-11-14
Fix infinite recursion error

pgrinaway:
Looks like I wanted a property to return itself. This PR is a fix to #34


jchodera:
Re-running travis.

Should we move `run_example.py` to a nosetest?

jchodera:
Merging, since travis won't work from your fork PR.

jchodera:
(I've been using forks in the main repo.)

PR #39, opened 2015-11-15
Refactoring AlchemicalEliminationEngine and NCMCEngine

jchodera:
Starting work on refactoring proposed in #32.


jchodera:
Whoops, I think I made a mistake in merging the API, since we would need to create a new `NCMCEngine` for each `topology_proposal` with this new API.

jchodera:
OK, as I have it now, for each topology proposal, we would need
```python
ncmc_engine = ncmc_switching.NCMCEngine(top_proposal, temperature=temperature, timestep=switching_timestep, nsteps=switching_nsteps, functions=switching_functions)
...
[ncmc_old_positions, ncmc_elimination_logp] = ncmc_engine.integrate(positions, direction='delete')
...
[ncmc_new_positions, ncmc_introduction_logp] = ncmc_engine.integrate(geometry_proposal.new_positions, direction='insert')
```
Presumably, we instead want to have
```python
# do this only once during initialization
ncmc_engine = ncmc_switching.NCMCEngine(temperature=temperature, timestep=switching_timestep, nsteps=switching_nsteps, functions=switching_functions)
...
# inside each inner iteration loop
[ncmc_old_positions, ncmc_elimination_logp] = ncmc_engine.integrate(topology_proposal, positions, direction='delete')
...
# inside each inner iteration loop
[ncmc_new_positions, ncmc_introduction_logp] = ncmc_engine.integrate(topology_proposal, geometry_proposal.new_positions, direction='insert')
```
Does that sound more reasonable?  Or are there cases where we would want to change the temperature, switching timestep, switching steps, or switching functions on the fly?

jchodera:
OK, I've refactored the API to be like the second example above.  I can always change it back or add accessors to modulate these parameters on the fly if we want more flexibility.

I've confirmed the `run_example.py` code runs up to some error I don't understand:
```
  File "run_example.py", line 222, in <module>
    run()
  File "run_example.py", line 168, in run
    log_weight = bias_calculator.g_k(top_proposal.metadata['molecule_smiles'])
TypeError: 'NoneType' object has no attribute '__getitem__'
```

jchodera:
Should be good to merge if the tests pass.

jchodera:
I'll also add the contribution from fully interacting to alchemical systems mentioned in #32.

jchodera:
The contribution of transforming a real system to/from alchemical system has been included.

This can probably be later optimized for speed/efficiency, but this should work for now.

PR #47, opened 2015-11-18
Added demo notebook for more complex forms of SAMS recursion strategy

jchodera:
I've created an IPython notebook illustrating the modified SAMS recursion scheme we discussed, in which we also adapt the target densities on the fly based on a more sophisticated criterion.  Here, I use a system of harmonic oscillator in a two-chain model to mimic the complex and solvent phases of @pgrinaway's ligand binding design problem, and adapt the target density to be proportional to the ratio of normalizing constants (binding affinity), and show that it seems to work, at least empirically.

modified-sams-target-densities
modified-sams-free-energies


jchodera:
Presuming we can figure out the theory as well with Z. Tan, we may be able to use this strategy for @juliebehr's mutation scheme instead of using Thompson sampling.

pgrinaway:
Cool! Can we set a time to discuss this a bit more?

jchodera:
Here's the [IPython notebook link](https://github.com/jchodera/perses/blob/notebooks/notebooks/Harmonic%20oscillators%20example.ipynb).

jchodera:
Mind if I merge this, or did you want to chat about this a bit more first?

Issue #49, opened 2015-11-30
Remaining issues in the GeometryEngine stuff


pgrinaway:
Currently, there are a couple issues with the GeometryEngine, in particular relating to the implementation that uses all angle and torsion potentials when choosing a torsion.

* Sometimes, the proposal potential results in some `nan`s somehow making their way to the rejection sampler. When a uniform random number is drawn with `nan` as the upper limit, `numpy` throws an exception. Looking into why this is still happening despite replacing `nan`s with `inf`.

* The speed is a bit of an issue. I've discovered through profiling that this is a result of the conversion between internal and cartesian coordinates. This can probably be accelerated a lot via vectorization or something of the sort.

jchodera:
Why use a rejection sampler here?  If we have computed the probability vector `p`, why not just use `numpy.random.choice(numpy.arange(phis, p=p)`?

jchodera:
`p` must be normalized for `numpy.random.choice`, but that is easy to do with `p / p.sum()` if that hasn't been done already.

jchodera:
For speeding things up, we have a few possibilities:

#### Compile a list of all torsions and angles and vectorize computations with numpy

This is a bit tricky, but should be straightforward given that all the other positions are fixed.

#### Code the energy terms in Cython

This might be the most straightforward way to recode the existing code.

#### Construct a minimum OpenMM system containing only the relevant atoms, bonds, and torsions

The trick here is to keep track of the atom mapping and compute the Cartesian positions for rotation around the single torsion, but it could also allow us to do more sophisticated things in the future.

jchodera:
You might want to return `phis` in [this function](https://github.com/choderalab/perses/blob/master/perses/rjmc/geometry.py#L593) to make it easier to do `numpy.random.choice` (and to make sure you don't get out of sync between the probability vector and torsions).

jchodera:
We can also refine your `log_p` computations a bit, and vectorize them.

For example, the [`_bond_logp`](https://github.com/choderalab/perses/blob/master/perses/rjmc/geometry.py#L475-L479) currently looks like this:
```python
        k_eq = bond.type.k*units.kilocalories_per_mole/(units.angstrom**2)
        r0 = bond.type.req*units.nanometers
        sigma = beta*2.0/np.sqrt(2.0*k_eq/k_eq.unit)
        logp = stats.distributions.norm.logpdf(r/r.unit, r0/r0.unit, sigma)
        return logp
```
[...]

PR #51, opened 2015-12-02
Fix remaining GeometryEngine issues

pgrinaway:
As discussed in #49, there are some outstanding issues in the `GeometryEngine`. Currently:

* Bond and angle `logp` have been converted to `logq`, so that the exponentiated version of it is the unnormalized probability density. 

* Deprecated the use of `scipy.stats`

* Corrected the form of sigma in the proposals and `logq`

* Switched to choosing from a discrete set of possible torsions instead of rejection sampling.

What remains:

* Calculate reverse torsion probability (next, in this PR)

* Cythonize the coordinate conversion (potentially another PR)


pgrinaway:
Ok, I have now implemented the reverse torsion logp.

pgrinaway:
Oops! stray imports from experimentation last night. Removed.

pgrinaway:
I think there are also some unit errors... please stand by...

pgrinaway:
Looks like I am calculating an excessive number of jacobians...

pgrinaway:
Ok. It runs now without error. However, the final energy is pretty bad. I'm going to head home soon and take a look at a pdb of the output to see what is up with that.

jchodera:
Do you mean this is slow in profiling, or you are just computing more than
you need?

pgrinaway:
I meant that I had accidentally deleted the `calculate_jacobian=False` keyword argument, so I was calculating Jacobians where it was unnecessary, and this caused the test to take a ridiculous amount of time.

pgrinaway:
So, computing more than I needed.

pgrinaway:
Hm, also it looks like there is an issue with integer division [here](https://github.com/choderalab/perses/pull/51/files#diff-a5d8d93e4a23e84d32747673e1244e75R498). 1/17 is unfortunately 0, so I've changed 1 to 1.0 there.

jchodera:
Reviewing PR now!

perses/rjmc/geometry.py
```diff
@@ -90,6 +89,7 @@ def propose(self, top_proposal):
         logp_proposal : float
             The log probability of the forward-only proposal
         """
+        beta = top_proposal.beta.in_units_of(units.kilocalorie_per_mole**-1)
```
jchodera: Isn't it dangerous to have a unitless `beta` hanging around alongside unit-bearing quantities?

jchodera: Whoops, nevermind---this just converts beta to units to 1/(kcal/mol), it doesn't strip the units.

perses/rjmc/geometry.py
```diff
@@ -117,16 +117,22 @@ def propose(self, top_proposal):
 
                 #propose a bond and calculate its probability
                 bond = self._get_relevant_bond(atom, bond_atom)
-                r_proposed = self._propose_bond(bond, top_proposal.beta)
-                logp_r = self._bond_logp(r_proposed, bond, top_proposal.beta)
+                r_proposed = self._propose_bond(bond, beta)
+                bond_k = bond.type.k*units.kilocalorie_per_mole/units.angstrom**2
+                sigma_r = units.sqrt(1/beta*bond_k)
```
jchodera: This should be `1/(beta*bond_k)`, which is different from what you have which maps to `(1/beta)*bond_k`

perses/rjmc/geometry.py
```diff
@@ -117,16 +117,22 @@ def propose(self, top_proposal):
...
+                logZ_r = np.log((np.sqrt(2*np.pi)*sigma_r/sigma_r.unit))
```
jchodera: Suggest you use a standard fixed unit (e.g. unit.angstrom) here just in case we end up changing the units of `sigma_r` or `beta` above.

perses/rjmc/geometry.py
```diff
@@ -117,16 +117,22 @@ def propose(self, top_proposal):
...
+                logp_r = self._bond_logq(r_proposed, bond, beta) - logZ_r
```
jchodera: Not something to worry about optimizing here, but I think lines 120-124 end up retrieving `bond_k` and recomputing `sigma_r` a total of three independent times.  There may be some refactoring we could think of in future that cut down on code duplication.

perses/rjmc/geometry.py
```diff
@@ -117,16 +117,22 @@ def propose(self, top_proposal):
...
 
                 #propose an angle and calculate its probability
                 angle = self._get_relevant_angle(atom, bond_atom, angle_atom)
-                theta_proposed = self._propose_angle(angle, top_proposal.beta)
-                logp_theta = self._angle_logp(theta_proposed, angle, top_proposal.beta)
+                theta_proposed = self._propose_angle(angle, beta)
+                angle_k = angle.type.k*units.kilocalorie_per_mole/units.angstrom**2
+                sigma_theta = units.sqrt(1/beta*angle_k)
```
jchodera: Again should be `1.0/(beta*angle_k)` and not `1/beta*angle_k` which groups as `(1/beta)*angle_k`

perses/rjmc/geometry.py
```diff
@@ -117,16 +117,22 @@ def propose(self, top_proposal):
...
+                sigma_r = units.sqrt(1/beta*bond_k)
+                logZ_r = np.log((np.sqrt(2*np.pi)*sigma_r/sigma_r.unit))
+                logp_r = self._bond_logq(r_proposed, bond, beta) - logZ_r
 
                 #propose an angle and calculate its probability
                 angle = self._get_relevant_angle(atom, bond_atom, angle_atom)
-                theta_proposed = self._propose_angle(angle, top_proposal.beta)
-                logp_theta = self._angle_logp(theta_proposed, angle, top_proposal.beta)
+                theta_proposed = self._propose_angle(angle, beta)
+                angle_k = angle.type.k*units.kilocalorie_per_mole/units.angstrom**2
```
jchodera: Force constant should be `units.kilocalories_per_mole/units.radian**2`, you have `units.kilocalorie_per_mole/units.angstrom**2`.

perses/rjmc/geometry.py
```diff
@@ -117,16 +117,22 @@ def propose(self, top_proposal):
...
+                sigma_theta = units.sqrt(1/beta*angle_k)
+                logZ_theta = np.log((np.sqrt(2*np.pi)*sigma_theta/sigma_theta.unit))
```
jchodera: Suggest again using a fixed unit (e.g. `units.radian`) instead of `sigma_theta.unit`.

perses/rjmc/geometry.py
```diff
@@ -193,11 +199,17 @@ def logp_reverse(self, top_proposal, new_coordinates, old_coordinates):
 
                 #propose a bond and calculate its probability
                 bond = self._get_relevant_bond(atom, bond_atom)
-                logp_r = self._bond_logp(internal_coordinates[0], bond, top_proposal.beta)
+                bond_k = bond.type.k*units.kilocalorie_per_mole/units.angstrom**2
+                sigma_r = units.sqrt(1/beta*bond_k)
```
jchodera: This should be `1/(beta*bond_k)`, not `1/beta*bond_k` which is grouped as `(1/beta)*bond_k`

perses/rjmc/geometry.py
```diff
@@ -193,11 +199,17 @@ def logp_reverse(self, top_proposal, new_coordinates, old_coordinates):
...
+                logZ_r = np.log((np.sqrt(2*np.pi)*sigma_r/sigma_r.unit)) #need to eliminate unit to allow numpy log
```
jchodera: Suggest using fixed units, like `unit.angstrom`, instead of `sigma_r.unit`.

perses/rjmc/geometry.py
```diff
@@ -193,11 +199,17 @@ def logp_reverse(self, top_proposal, new_coordinates, old_coordinates):
...
+                logp_r = self._bond_logq(internal_coordinates[0], bond, beta) - logZ_r
 
                 #propose an angle and calculate its probability
                 angle = self._get_relevant_angle(atom, bond_atom, angle_atom)
-                logp_theta = self._angle_logp(internal_coordinates[1], angle, top_proposal.beta)
+                angle_k = angle.type.k*units.kilocalorie_per_mole/(units.radians**2)
+                sigma_theta = units.sqrt(1/beta*angle_k)
```
jchodera: This should be `1/(beta*angle_k)`

perses/rjmc/geometry.py
```diff
@@ -193,11 +199,17 @@ def logp_reverse(self, top_proposal, new_coordinates, old_coordinates):
...
+                logZ_theta = np.log((np.sqrt(2*np.pi)*sigma_theta/sigma_theta.unit)) #need to eliminate unit to allow numpy log
```
jchodera: Suggest using fixed unit here, like `units.radian` instead of `sigma_theta.unit`.

perses/rjmc/geometry.py
```diff
@@ -474,30 +485,8 @@ def _angle_logp(self, theta, angle, beta):
         """
         k_eq = angle.type.k*units.kilocalories_per_mole/(units.radians**2)
         theta0 = angle.type.theteq*units.radians
```
jchodera: I think `angle.type.theteq` is actually in degrees.

@swails uses `DEG_TO_RAD` to convert these to radians when he uses these values:
https://github.com/ParmEd/ParmEd/blob/master/parmed/structure.py#L1905

perses/rjmc/geometry.py
```diff
@@ -474,30 +485,8 @@ def _angle_logp(self, theta, angle, beta):
         """
         k_eq = angle.type.k*units.kilocalories_per_mole/(units.radians**2)
```
jchodera: The AMBER force constants may also be off by a factor of two.  Trace the call @swails uses here carefully to double-check:
https://github.com/ParmEd/ParmEd/blob/master/parmed/structure.py#L2053-L2054

perses/rjmc/geometry.py
```diff
@@ -536,9 +525,9 @@ def _propose_angle(self, angle, beta):
         Bond angle proposal
         """
         theta0 = angle.type.theteq*units.radians
```
jchodera: I think `angle.type.theteq` is actually in degrees.

@swails uses `DEG_TO_RAD` to convert these to radians when he uses these values:
https://github.com/ParmEd/ParmEd/blob/master/parmed/structure.py#L1905

perses/rjmc/geometry.py
```diff
@@ -536,9 +525,9 @@ def _propose_angle(self, angle, beta):
...
-        k_eq = angle.type.k*units.kilocalories_per_mole/(units.radians**2)
-        sigma = beta*2.0/np.sqrt(2.0*k_eq/k_eq.unit)
-        theta = sigma/sigma.unit*np.random.random()*theta0.unit + theta0
+        k = angle.type.k*units.kilocalorie_per_mole/units.radian**2
```
jchodera: The AMBER force constants may also be off by a factor of two.  Trace the call @swails uses here carefully to double-check:
https://github.com/ParmEd/ParmEd/blob/master/parmed/structure.py#L2053-L2054

perses/rjmc/geometry.py
jchodera: The AMBER force constants may also be off by a factor of two.  Trace the call @swails uses here carefully to double-check:
https://github.com/ParmEd/ParmEd/blob/master/parmed/structure.py#L2053-L2054

perses/rjmc/geometry.py
```diff
@@ -193,11 +199,17 @@ def logp_reverse(self, top_proposal, new_coordinates, old_coordinates):
 
                 #propose a bond and calculate its probability
                 bond = self._get_relevant_bond(atom, bond_atom)
-                logp_r = self._bond_logp(internal_coordinates[0], bond, top_proposal.beta)
+                bond_k = bond.type.k*units.kilocalorie_per_mole/units.angstrom**2
```
jchodera: AMBER bond force constants may be off by a factor of two. Check what @swails does here carefully?
https://github.com/ParmEd/ParmEd/blob/master/parmed/structure.py#L2002-L2003

perses/rjmc/geometry.py
```diff
@@ -455,11 +467,10 @@ def _bond_logp(self, r, bond, beta):
         """
         k_eq = bond.type.k*units.kilocalories_per_mole/(units.angstrom**2)
```
jchodera: AMBER bond force constants may be off by a factor of two. Check what @swails does here carefully?
https://github.com/ParmEd/ParmEd/blob/master/parmed/structure.py#L2002-L2003

perses/rjmc/geometry.py
```diff
@@ -525,10 +514,10 @@ def _propose_bond(self, bond, beta):
         """
         Bond length proposal
         """
-        k_eq = bond.type.k*units.kilocalories_per_mole/(units.angstrom**2)
         r0 = bond.type.req*units.angstrom
-        sigma = beta*2.0/np.sqrt(2.0*k_eq/k_eq.unit)
-        r = sigma/sigma.unit*np.random.random()*r0.unit + r0
+        k = bond.type.k*units.kilocalorie_per_mole/units.angstrom**2
```
jchodera: AMBER bond force constants may be off by a factor of two. Check what @swails does here carefully?
https://github.com/ParmEd/ParmEd/blob/master/parmed/structure.py#L2002-L2003

perses/rjmc/geometry.py
```diff
@@ -669,667 +661,20 @@ def _torsion_logp(self, atom, xyz, torsion, atoms_with_positions, positions, bet
...
-            if p_phi_samp > runif:
-                phi = phi_samp
-                logp = p_phi_samp
-                accepted = True
-            else:
-                continue
+        p, Z, q, phis = self._normalize_torsion_proposal(atom, r, theta, bond_atom, angle_atom, torsion_atom, atoms_with_positions, positions, beta, n_divisions=5000)
+        #choose from the set of possible torsion angles
+        phi_idx = np.random.choice(range(len(phis)), p=p)
+        logp = np.log(p[phi_idx])
```
jchodera: Since `p` might vary by orders of magnitude, would be better if you had `self._normalize_torsion_proposal()` return `log_p` directly as well?

jchodera:
Code review complete!  Probably best to look at the [files changed diff view](https://github.com/choderalab/perses/pull/51/files).

We might want to think about how to reduce code duplication in a future PR (e.g. between forward and reverse proposals, and in other places) to keep the potential for bugs to a minimum.

pgrinaway:
Thanks! With regard to `sigma` and units, wouldn't it be better to use `sigma.units` so that the units of `r_proposed` or `theta_proposed` always match the units of `sigma`?

jchodera:
> Thanks! With regard to sigma and units, wouldn't it be better to use sigma.units so that the units of r_proposed or theta_proposed always match the units of sigma?

Hm. My worry is this: If, somehow, some of the force constants were in different units from each other, dividing by `sigma.unit` would mask a problem in which we *thought* everything was having the same additive constant subtracted but it wasn't.  I agree that it seems unlikely that some force constants will be in different units, however, so it is irrelevant at the moment.

perses/rjmc/geometry.py
```diff
@@ -525,10 +514,10 @@ def _propose_bond(self, bond, beta):
...
         Bond length proposal
         """
-        k_eq = bond.type.k*units.kilocalories_per_mole/(units.angstrom**2)
         r0 = bond.type.req*units.angstrom
-        sigma = beta*2.0/np.sqrt(2.0*k_eq/k_eq.unit)
-        r = sigma/sigma.unit*np.random.random()*r0.unit + r0
+        k = bond.type.k*units.kilocalorie_per_mole/units.angstrom**2
+        sigma_r = units.sqrt(1.0/(beta*k))
+        r = sigma_r*np.random.random() + r0
         return r
```
jchodera: We want to return just `r0` if this is a constrained bond.

perses/rjmc/geometry.py
```diff
@@ -632,7 +624,7 @@ def _normalize_torsion_proposal(self, atom, r, theta, bond_atom, angle_atom, tor
             ub[i]+=ub_i
 
         #exponentiate to get the unnormalized probability
-        q = np.exp(-ub)
+        q = np.exp(ub)
```
jchodera: I'm confused by `exp(ub)`.  Is `ub` supposed to be an energy?  If not, maybe you could just call this `logq` instead of `ub`?

perses/rjmc/geometry.py
```diff
@@ -455,11 +467,10 @@ def _bond_logp(self, r, bond, beta):
         """
         k_eq = bond.type.k*units.kilocalories_per_mole/(units.angstrom**2)
         r0 = bond.type.req*units.nanometers
-        sigma = beta*2.0/np.sqrt(2.0*k_eq/k_eq.unit)
-        logp = stats.distributions.norm.logpdf(r/r.unit, r0/r0.unit, sigma)
-        return logp
+        logq = -beta*0.5*k_eq*(r-r0)**2
+        return logq
```
jchodera: If this bond is constrained, we should return a `logq` of 0.  We might also add a check whether `np.abs((r-r0)/r0)` is within some specified relative tolerance in that case, or throw an exception.

jchodera:
Also, sorry for the force constant factor of two craziness.

I believe OpenMM uses the proper harmonic form `(K/2)*(x-x0)**2`, but the Amber force constants may assume `K*(x-x0)**2`---you'll need to double check that.  Also, having some quantities in radians and others in degrees is pure insanity.

pgrinaway:
Ok, locally it seems subtracting `max(logq)` from `logq` got rid of the `nan`s. Thanks1

pgrinaway:
[Not sure ](http://parmed.github.io/ParmEd/html/openmm.html?highlight=constraints) if `ParmEd` supports OpenMM systems with constraints. Looking in to the harmonic force constant issue now. Once that is fixed, I think it might be good to merge and put in the Cython-based coordinate conversion so that I can debug the other issues here.

jchodera:
Here's how I would handle bond constraints:
* Have your constructor take an argument about `constrants` that defaults to `None` but could be one of the `simtk.openmm.app` constraint types.
* For `constrants=app.Hbonds`, you would check if any of the atoms in the bond are hydrogen, following [this code](https://github.com/ParmEd/ParmEd/blob/master/parmed/amber/_amberparm.py#L1150-L1151) in ParmEd:
```python
if (bond.atom1.atomic_number == 1 or bond.atom2.atomic_number == 1):
   # bond is constrained
```
You might separate out this functionality into an `_bondIsConstrained(bond, constraints)` method or something to isolate the checking behavior.

perses/rjmc/geometry.py
```diff
@@ -193,11 +199,17 @@ def logp_reverse(self, top_proposal, new_coordinates, old_coordinates):
...
-                logp_r = self._bond_logp(internal_coordinates[0], bond, top_proposal.beta)
+                bond_k = bond.type.k*units.kilocalorie_per_mole/units.angstrom**2
+                sigma_r = units.sqrt(1/(beta*bond_k))
+                logZ_r = np.log((np.sqrt(2*np.pi)*sigma_r/sigma_r.unit)) #need to eliminate unit to allow numpy log
+                logp_r = self._bond_logq(internal_coordinates[0], bond, beta) - logZ_r
 
                 #propose an angle and calculate its probability
                 angle = self._get_relevant_angle(atom, bond_atom, angle_atom)
-                logp_theta = self._angle_logp(internal_coordinates[1], angle, top_proposal.beta)
+                angle_k = angle.type.k*units.kilocalorie_per_mole/(units.degrees**2)
```
jchodera: I'm pretty sure the conversion is `units.kilocalorie_per_mole/(units.radians**2)`:
https://github.com/ParmEd/ParmEd/blob/master/parmed/structure.py#L2054

perses/rjmc/geometry.py
```diff
@@ -472,32 +483,10 @@ def _angle_logp(self, theta, angle, beta):
...
-            torsion_atom = torsion.atom1
-
-        internal_coordinates = self._autograd_ctoi(xyz, positions[bond_atom.idx], positions[angle_atom.idx], positions[torsion_atom.idx])
-        beta = beta/beta.unit
-        gamma = torsion.type.phase
-        V = torsion.type.phi_k
-        n = torsion.type.per
-        q = np.exp(-beta*(V/2.0)*(1+np.cos(n*internal_coordinates[2]-gamma)))
-        return q
+        k_eq = angle.type.k*units.kilocalories_per_mole/(units.degrees**2)
```
jchodera: I'm pretty sure the conversion is `units.kilocalories_per_mole/(units.radians**2)`
https://github.com/ParmEd/ParmEd/blob/master/parmed/structure.py#L2054

pgrinaway:
Yeah, that is how I was expecting to handle it. I just need to make sure that the indexing of the `ParmEd` particles correspond to the OpenMM `System` particles. Looking into that now.

perses/rjmc/geometry.py
```diff
@@ -525,23 +514,23 @@ def _propose_bond(self, bond, beta):
         """
         Bond length proposal
         """
-        k_eq = bond.type.k*units.kilocalories_per_mole/(units.angstrom**2)
         r0 = bond.type.req*units.angstrom
-        sigma = beta*2.0/np.sqrt(2.0*k_eq/k_eq.unit)
-        r = sigma/sigma.unit*np.random.random()*r0.unit + r0
+        k = bond.type.k*units.kilocalorie_per_mole/units.angstrom**2
+        sigma_r = units.sqrt(1.0/(beta*k))
+        r = sigma_r/sigma_r.unit*np.random.random()*units.angstrom + r0
```
jchodera: This could probably be simplified to
```python
r = sigma_r*np.random.random() + r0
```

perses/rjmc/geometry.py
```diff
@@ -525,23 +514,23 @@ def _propose_bond(self, bond, beta):
...
     def _propose_angle(self, angle, beta):
         """
         Bond angle proposal
         """
-        theta0 = angle.type.theteq*units.radians
-        k_eq = angle.type.k*units.kilocalories_per_mole/(units.radians**2)
-        sigma = beta*2.0/np.sqrt(2.0*k_eq/k_eq.unit)
-        theta = sigma/sigma.unit*np.random.random()*theta0.unit + theta0
+        theta0 = angle.type.theteq*units.degrees
+        k = angle.type.k*units.kilocalorie_per_mole/units.degree**2
```
jchodera: I'm pretty sure this should be `units.kilocalorie_per_mole/(units.radians**2)`

perses/rjmc/geometry.py
```diff
@@ -525,23 +514,23 @@ def _propose_bond(self, bond, beta):
...
+        sigma_theta = units.sqrt(1.0/(beta*k))
+        theta = sigma_theta/sigma_theta.unit*np.random.random()*units.radian + theta0
```
jchodera: This could be simplified to:
```python
theta = sigma_theta*np.random.random() + theta0
```

perses/rjmc/geometry.py
```diff
@@ -551,8 +540,11 @@ def _torsion_potential(self, torsion, phi, beta):
         gamma = torsion.type.phase
```
jchodera: Radians or degrees?

perses/rjmc/geometry.py
pgrinaway: Yeah. The `ParmEd` value does not have units, but is expressed in `kcal/mol/degrees**2`. So, when I pull that number out of the attribute, I immediately assign it the correct units (I'm not converting anything there). Otherwise, if I had `radians**2` there, then I'd be using the wrong unit.

It looks like in the code you quoted, the `ParmEd` value is in degrees, and has to be converted to radians to work in OpenMM. However, the explicit conversion just has to happen because `ParmEd` is not using unit-bearing quantities, no?

perses/rjmc/geometry.py
pgrinaway: As below, this is not a conversion. It's just assigning the correct unit given the `ParmEd` defaults...unless I'm mistaken about those.

perses/rjmc/geometry.py
jchodera: The `phase` should be converted to degrees:
https://github.com/ParmEd/ParmEd/blob/master/parmed/structure.py#L2081

Maybe also call it `phase` instead of `gamma`?

jchodera: Also, you have `phi = phi/phi.unit` above---probably not a good idea!  Can you attach proper units to `phi`, `beta`, `phase`, and `V` and then make sure you multiply them together into a non-unit-bearing quantity?

swails:
All of the raw angle values are specified in degrees in ParmEd.  Whenever an angle appears in a force constant, that force constant is expressed in radians.  This convention came from Amber (for which ParmEd was originally written).  So for angles:

```Python
angle.type.k      # kcal/mol/radians**2
angle.type.theteq # degrees
```

swails:
And as for loading an OpenMM System with constraints into a ParmEd `Structure` -- ParmEd does nothing with constraints.  It only reads the `HarmonicBondForce` and applies the parameters it finds from there.  If you add constraints, you should do so by setting `flexibleConstraints=True` (or something like that).  This will make OpenMM (or ParmEd, whichever you use to create the `System`) add bonds with force constants even to those terms that are constrained.

This is the only way you can get the bonds correct *by default* in ParmEd when constraints are present.  I'm not sure if `AmberPrmtopFile` from OpenMM exposes `flexibleConstraints` in the top-level `createSystem` call, but I know `AmberParm` does.

jchodera:
> All of the raw angle values are specified in degrees in ParmEd. Whenever an angle appears in a force constant, that force constant is expressed in radians. This convention came from Amber (for which ParmEd was originally written).

I think it's telling that we've managed to get this wrong like five times already because of the confusion that expressing some units in radians and others in degrees brings.  :)

I'd love to see a way to retrieve unit-bearing versions of the various quantities, though I realize that this would detract from the very convenient pythonic nature of working directly with the `TrackedList` items unless you made the decision to *always* have unit-bearing quantities in anything accessible to users.  I think that would be awesome, but a big change.

For now, @pgrinaway is trying to isolate the code that deals with non-unit-bearing quantities to just a few locations in `perses` so that the unit system will work out most of the issues for us.

pgrinaway:
Thanks for the comments, @swails! 

@jchodera : I'm thinking it's easy to just have methods that add units to a torsion, angle, or bond as a convenience function. This reduces unit-adding code to just three (very simple) methods.

jchodera:
Sounds good!

jchodera:
Looks like there is still a units error:
```
  File "/home/travis/miniconda/envs/_test/lib/python2.7/site-packages/perses-0.1-py2.7.egg/perses/rjmc/geometry.py", line 120, in propose
    r_proposed = self._propose_bond(bond, beta)
  File "/home/travis/miniconda/envs/_test/lib/python2.7/site-packages/perses-0.1-py2.7.egg/perses/rjmc/geometry.py", line 558, in _propose_bond
    sigma_r = units.sqrt(1.0/(beta*k))
  File "/home/travis/miniconda/envs/_test/lib/python2.7/site-packages/simtk/unit/unit_math.py", line 141, in sqrt
    return val.sqrt()
  File "/home/travis/miniconda/envs/_test/lib/python2.7/site-packages/simtk/unit/quantity.py", line 449, in sqrt
    new_unit = self.unit.sqrt()
  File "/home/travis/miniconda/envs/_test/lib/python2.7/site-packages/simtk/unit/unit.py", line 279, in sqrt
    raise ArithmeticError('Exponents in Unit.sqrt() must be even.')
ArithmeticError: Exponents in Unit.sqrt() must be even.
```

perses/rjmc/geometry.py
```diff
@@ -455,11 +506,10 @@ def _bond_logp(self, r, bond, beta):
         """
         k_eq = bond.type.k*units.kilocalories_per_mole/(units.angstrom**2)
         r0 = bond.type.req*units.nanometers
```
jchodera: Don't need to attach units here anymore.

jchodera:
Closer!
```
ERROR: Run the geometry engine a few times to make sure that it actually runs
  File "/home/travis/miniconda/envs/_test/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/miniconda/envs/_test/lib/python2.7/site-packages/perses-0.1-py2.7.egg/perses/tests/test_geometry_engine.py", line 92, in test_run_geometry_engine
    new_positions, logp_proposal = geometry_engine.propose(sm_top_proposal)
  File "/home/travis/miniconda/envs/_test/lib/python2.7/site-packages/perses-0.1-py2.7.egg/perses/rjmc/geometry.py", line 124, in propose
    logp_r = self._bond_logq(r_proposed, bond, beta) - logZ_r
  File "/home/travis/miniconda/envs/_test/lib/python2.7/site-packages/perses-0.1-py2.7.egg/perses/rjmc/geometry.py", line 509, in _bond_logq
    logq = -beta*0.5*k_eq*(r-r0)**2
  File "/home/travis/miniconda/envs/_test/lib/python2.7/site-packages/simtk/unit/quantity.py", line 253, in __sub__
    raise TypeError('Cannot subtract two quantities with incompatible units "%s" and "%s".' % (self.unit, other.unit))
TypeError: Cannot subtract two quantities with incompatible units "angstrom" and "angstrom**2"
```

perses/rjmc/geometry.py
pgrinaway: Yeah, going through and finding all of these...

pgrinaway:
Ok I think I see the issue. Will address after lunch.

pgrinaway:
fixing the torsion units..

jchodera:
So close!

pgrinaway:
Ok, there is some intermittent failure here (and even when it doesn't fail, the energies are pretty crazy). This is going to require more debugging. I'm going to replace the coordinate conversions with Cython equivalents to facilitate that.

perses/rjmc/geometry.py
jchodera: You should no longer need to recast to a specific set of units.

perses/rjmc/geometry.py
```diff
@@ -157,6 +163,7 @@ def logp_reverse(self, top_proposal, new_coordinates, old_coordinates):
         logp : float
             The log probability of the proposal for the given transformation
         """
+        beta = top_proposal.beta.in_units_of(units.kilocalorie_per_mole**-1)
```
jchodera: Should no longer need to recast units of beta.

perses/rjmc/geometry.py
```diff
@@ -235,8 +252,48 @@ def _get_relevant_angle(self, atom1, atom2, atom3):
...
+        angle : parmed angle object
+             the angle to get unit-ed
+
+        Returns
+        -------
+        angle_with_units : parmed angle
+            The angle, but with units on its parameters
+        """
+        angle.type.theteq = units.Quantity(angle.type.theteq, unit=units.degree)
+        angle.type.k = units.Quantity(angle.type.k, unit=units.kilocalorie_per_mole/units.radian**2)
```
jchodera: Remember that the spring constants have to be multiplied by a factor of two to convert from ParmEd to OpenMM convention!

perses/rjmc/geometry.py
```diff
@@ -235,8 +252,48 @@ def _get_relevant_angle(self, atom1, atom2, atom3):
...
+        ---------
+        bond : parmed bond object
+            The bond to get units
+
+        Returns
+        -------
+
+        """
+        bond.type.req = units.Quantity(bond.type.req, unit=units.angstrom)
+        bond.type.k = units.Quantity(bond.type.k, unit=units.kilocalorie_per_mole/units.angstrom**2)
```
jchodera: Remember that the spring constants have to be multiplied by a factor of two to convert from ParmEd to OpenMM convention!

perses/rjmc/geometry.py
```diff
@@ -525,34 +560,37 @@ def _propose_bond(self, bond, beta):
...
         Bond length proposal
         """
-        k_eq = bond.type.k*units.kilocalories_per_mole/(units.angstrom**2)
-        r0 = bond.type.req*units.angstrom
-        sigma = beta*2.0/np.sqrt(2.0*k_eq/k_eq.unit)
-        r = sigma/sigma.unit*np.random.random()*r0.unit + r0
+        r0 = bond.type.req
+        k = bond.type.k
+        sigma_r = units.sqrt(1.0/(beta*k))
+        r = sigma_r/sigma_r.unit*np.random.random()*units.angstrom + r0
```
jchodera: Safer to write `r = sigma_r * np.random.random() + r0`.

perses/rjmc/geometry.py
```diff
@@ -525,34 +560,37 @@ def _propose_bond(self, bond, beta):
...
         Bond angle proposal
         """
-        theta0 = angle.type.theteq*units.radians
-        k_eq = angle.type.k*units.kilocalories_per_mole/(units.radians**2)
-        sigma = beta*2.0/np.sqrt(2.0*k_eq/k_eq.unit)
-        theta = sigma/sigma.unit*np.random.random()*theta0.unit + theta0
+        theta0 = angle.type.theteq
+        k = angle.type.k
+        sigma_theta = units.sqrt(1.0/(beta*k))
+        theta = sigma_theta/sigma_theta.unit*np.random.random()*units.radian + theta0
```
jchodera: Safer to write `theta = sigma_theta*np.random.random() + theta0`

jchodera: Otherwise, this is a bug, since `sigma_theta` is in degrees and you are assuming it is in units of radians.

perses/rjmc/geometry.py
```diff
@@ -525,34 +560,37 @@ def _propose_bond(self, bond, beta):
...
         of the torsion
         """
-        phi = phi/phi.unit
-        beta = beta/beta.unit
-        gamma = torsion.type.phase
-        V = torsion.type.phi_k
+        phi = phi
+        beta = beta
+        gamma = torsion.type.phase*units.degree
+        V = torsion.type.phi_k*units.kilocalorie_per_mole
```
jchodera: You shouldn't be assigning units here.  Can you convert these to appropriate units in `_add_torsion_units()`?

perses/rjmc/geometry.py
```diff
@@ -624,23 +662,24 @@ def _normalize_torsion_proposal(self, atom, r, theta, bond_atom, angle_atom, tor
...
         #set up arrays for energies from angles and torsions
-        ub= np.zeros(n_divisions)
+        logq = np.zeros(n_divisions)
         for i, xyz in enumerate(xyzs):
-            ub_i = self._torsion_and_angle_potential(xyz, atom, positions, involved_angles, involved_torsions, beta)
-            if np.isnan(ub_i):
-                ub_i = np.inf
-            ub[i]+=ub_i
+            logq_i = self._torsion_and_angle_logq(xyz, atom, positions, involved_angles, involved_torsions, beta)
+            if np.isnan(logq_i):
```
jchodera: `logq` should never be `NaN`.  Let's eliminate this, since we want to flag this as a bug if we run into `NaN`.

perses/rjmc/geometry.py
```diff
@@ -624,23 +662,24 @@ def _normalize_torsion_proposal(self, atom, r, theta, bond_atom, angle_atom, tor
...
+                logq_i = - np.inf
+            logq[i]+=logq_i
+        logq - max(logq)
```
jchodera: Do you mean to say
```
logq -= max(logq)
```
?  Otherwise, this statement does nothing.

perses/rjmc/geometry.py
```diff
@@ -666,670 +705,21 @@ def _torsion_logp(self, atom, xyz, torsion, atoms_with_positions, positions, bet
...
-        if not Z:
-            p, Z = self._normalize_torsion_proposal(atom, internal_coordinates[0], internal_coordinates[1], bond_atom, angle_atom, torsion_atom, atoms_with_positions, positions, beta, n_divisions=5000)
-        ub_torsion = self._torsion_and_angle_potential(xyz, atom, positions, involved_angles, involved_torsions, beta)
-        p_torsion = np.exp(-ub_torsion) / Z
-        return p_torsion
+        p, Z, q, phis, logq = self._normalize_torsion_proposal(atom, internal_coordinates[0], internal_coordinates[1], bond_atom, angle_atom, torsion_atom, atoms_with_positions, positions, beta, n_divisions=5000)
+        #find the phi that's closest to the internal_coordinate phi:
+        phi_idx, phi = min(enumerate(phis), key=lambda x: abs(x[1]-internal_coordinates[2]))
+        p_phi = p[phi_idx]
+        return np.log(p_phi)
```
jchodera: `log(p_phi)` is not numerically robust.  We want to return 
```python
logq[phi_idx] - logSumExp(logq)
```
instead, which is numerically reobust.

In fact, why does `_normalize_torsion_proposal` return `logq` instead of `logp`?

perses/rjmc/geometry.py
```diff
@@ -666,670 +705,21 @@ def _torsion_logp(self, atom, xyz, torsion, atoms_with_positions, positions, bet
...
-            if p_phi_samp > runif:
-                phi = phi_samp
-                logp = p_phi_samp
-                accepted = True
-            else:
-                continue
+        p, Z, q, phis, logq = self._normalize_torsion_proposal(atom, r, theta, bond_atom, angle_atom, torsion_atom, atoms_with_positions, positions, beta, n_divisions=5000)
+        #choose from the set of possible torsion angles
+        phi_idx = np.random.choice(range(len(phis)), p=p)
+        logp = np.log(p[phi_idx])
```
jchodera: This is not a numerically robust way to compute `logp`.  Instead, use `logq[phi_idx] - logSumExp(logq)`, or just have _normalize_torsion_proposal()` return `logp` instead.

perses/rjmc/geometry.py
pgrinaway: Oops, this is what I was doing before. I think I misunderstood a previous comment, then.

pgrinaway:
Refactoring the torsion unit thing now--found another bug.

perses/rjmc/geometry.py
```diff
@@ -268,7 +343,8 @@ def _get_torsions(self, atoms_with_positions, new_atom):
                 continue
             if set(atoms_with_positions).issuperset(set(torsion_partners)):
                 eligible_torsions.append(torsion)
-        return torsions
+        eligible_torsions_with_units = [self._add_torsion_units(torsion) if type(torsion.type.phi_k)!=units.Quantity else torsion for torsion in eligible_torsions]
```
jchodera: This might be simplified if you have `self._add_XXXX_units()` handle the checking about whether units need to be added or not.

pgrinaway: good point, done

pgrinaway:
Looks like the problem is actually happening in the calculation of the internal coordinates--phi is turning out to be `NaN`

jchodera:
> Looks like the problem is actually happening in the calculation of the internal coordinates--phi is turning out to be NaN

Can you check to make sure we don't have a degenerate case for `r <= 0` or `theta = 0 or 180*degrees`?

perses/rjmc/geometry.py
```diff
@@ -586,8 +644,8 @@ def _torsion_and_angle_potential(self, xyz, atom, positions, involved_angles, in
             torsion_atom_position = xyz if torsion.atom4 == atom else positions[torsion.atom4.idx]
             internal_coordinates, _ = self._autograd_ctoi(atom_position, bond_atom_position, angle_atom_position, torsion_atom_position, calculate_jacobian=False)
             phi = internal_coordinates[2]*units.radians
```
jchodera: So you're thinking `internal_coordinates[2]` is `NaN` here?

pgrinaway:
Hm. Yeah, I can add those tests. I think here the problem is some numerical instability. For this case, 
`np.dot(plane1, plane2)/(np.linalg.norm(plane1)*np.linalg.norm(plane2))` comes out to -1.0--or so it appears. Comparing that to -1.0, -1.0 is greater. So, presumably this is outside the domain of `arccos`, giving us the `nan` problem.

perses/rjmc/geometry.py
pgrinaway: It is, according to the debugger.

pgrinaway:
Ok, I'm going to add some tests for that tonight as well as rewrite the coordinate conversions.

jchodera:
> Hm. Yeah, I can add those tests. I think here the problem is some numerical instability. For this case, 
np.dot(plane1, plane2)/(np.linalg.norm(plane1)*np.linalg.norm(plane2)) comes out to -1.0--or so it appears. Comparing that to -1.0, -1.0 is greater. So, presumably this is outside the domain of arccos, giving us the nan problem.

Ah!  For now, it makes sense to restrict this to the domain of `arccos` if it is outside the limits.

There is a much more stable torsion angle computation scheme from Bill Swope.  Let me dig up.

jchodera:
http://onlinelibrary.wiley.com/doi/10.1002/jcc.540130508/abstract

pgrinaway:
Awesome, thanks!

jchodera:
The PDF is on slack in `eternal-journal-club` now.

perses/rjmc/geometry.py
jchodera: Can you tell which torsion explodes?  In particular, is it a regular torsion, or an improper?

jchodera:
BTW, if debugging is painfully slow, you can totally reduce the number of points you evaluate the dihedral at for testing purposes.

perses/rjmc/geometry.py
pgrinaway: Yeah. I filter out impropers, and I have the problematic coordinates.

pgrinaway:
>The PDF is on slack in eternal-journal-club now.

Thanks!

perses/rjmc/geometry.py
jchodera: Hm.  Maybe we should *include* the impropers in the energy computation here in order to correctly capture ring closing and planarity.  We would still want to exclude them from picking which driver torsions should be used to place new atoms, however.

pgrinaway: That's a good point, thanks!

jchodera:
Very minor point, but we don't want to use `linspace` here to pick the `phis`:
https://github.com/choderalab/perses/pull/51/files#diff-a5d8d93e4a23e84d32747673e1244e75R677

[`linspace`](http://docs.scipy.org/doc/numpy/reference/generated/numpy.linspace.html) gives the closed interval [0,2*pi].  We want [`arange`](http://docs.scipy.org/doc/numpy/reference/generated/numpy.arange.html#numpy.arange), which gives the open interval `[0,2*pi)`.

pgrinaway:
Thanks, fixed!

perses/rjmc/geometry.py
```diff
@@ -616,7 +674,7 @@ def _normalize_torsion_proposal(self, atom, r, theta, bond_atom, angle_atom, tor
         involved_torsions = self._get_torsions(atoms_with_positions, atom)
 
         #get an array of [0,2pi)
-        phis = units.Quantity(np.linspace(0, 2.0*np.pi, n_divisions), unit=units.radians)
+        phis = units.Quantity(np.arange(0, 2.0*np.pi, n_divisions), unit=units.radians)
```
jchodera: Looks like there's a discrepancy in vector sizes now because the syntax for [`arange`] is different:
```
ERROR: Run the geometry engine a few times to make sure that it actually runs
  File "/home/travis/miniconda/envs/_test/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/miniconda/envs/_test/lib/python2.7/site-packages/perses-0.1-py2.7.egg/perses/tests/test_geometry_engine.py", line 92, in test_run_geometry_engine
    new_positions, logp_proposal = geometry_engine.propose(sm_top_proposal)
  File "/home/travis/miniconda/envs/_test/lib/python2.7/site-packages/perses-0.1-py2.7.egg/perses/rjmc/geometry.py", line 135, in propose
    phi_proposed, logp_phi = self._propose_torsion(atom, r_proposed, theta_proposed, bond_atom, angle_atom, torsion_atom, torsion, atoms_with_positions, new_positions, beta)
  File "/home/travis/miniconda/envs/_test/lib/python2.7/site-packages/perses-0.1-py2.7.egg/perses/rjmc/geometry.py", line 741, in _propose_torsion
    phi_idx = np.random.choice(range(len(phis)), p=np.exp(logp))
  File "mtrand.pyx", line 1099, in mtrand.RandomState.choice (numpy/random/mtrand/mtrand.c:10676)
ValueError: a and p must have same size
```
This should be
```python
phis = units.Quantity(np.arange(0, 2.0*np.pi, (2.0*np.pi)/n_divisions), unit=units.radians)
```

pgrinaway:
I think a numerically-stable version exists using `atan2` as described in this (seems to work fine on problem cases that I'm having): http://math.stackexchange.com/questions/47059/how-do-i-calculate-a-dihedral-angle-given-cartesian-coordinates

Or, as you suggested, we could use the alternate form given in the reference.

jchodera:
I wish stackexchange answered every science question!

pgrinaway:
There's a bond length issue. Working on this now...

pgrinaway:
It's a units issue.

pgrinaway:
units issue resolved, but the geometry is still wonky.

pgrinaway:
Yeah, there's something wrong with the coordinate conversion.

jchodera:
> Yeah, there's something wrong with the coordinate conversion.

It's not invertible?

pgrinaway:
No, it isn't, as you suggested. I'm debugging it now.

pgrinaway:
That would be a good test to add, I'd imagine.

jchodera:
Woohoo!  Tests pass, but any idea what these warnings mean?
```
/home/travis/miniconda/envs/_test/lib/python2.7/site-packages/parmed/openmm/topsystem.py:429: OpenMMWarning: Detected incomplete exceptions. Not supported.
```

pgrinaway:
I'm not sure what that warning is. I'll wrap this PR up tomorrow.

pgrinaway:
Working on this now.

jchodera:
There might be something to that warning.  It seems to indicate a discrepancy in exclusions between the `bond_graph_exceptions` (where 1-2 and 1-3 bonded atoms are excluded from LJ and charge interactions) and the list of explicit exceptions (where `epsilon` and `chargeProd` are 0) in the resulting OpenMM `System` object.

This seems to occur when calling `parmed.openmm.topsystem.load_topology()` with both a `topology` and `system` object where the two differ in how exceptions are constructed.

You call this line [here](https://github.com/choderalab/perses/blob/master/perses/rjmc/geometry.py#L94) in `geometry.py`.  Can you double-check that you created the `top_proposal.new_system` object with ParmEd's `createSystem()` for the *same* topology object as `top_proposal.new_topology`?  It seems like they are inconsistent, which will lead to unexpected errors.

pgrinaway:
Hm, I wasn't using parmed to create the system + topology, but rather `simtk.openmm.AmberPrmtopFile`.

pgrinaway:
See https://github.com/choderalab/perses/blob/master/perses/tests/test_geometry_engine.py#L40-L42 for where the `topology` and `system` come from in the test.

jchodera:
OK, I think we need some help from @swails to figure this one out.

Presumably, this piece of code 
```python
from simtk.openmm import app
import parmed
prmtop = app.AmberPrmtopFile(prmtop_filename)
system = prmtop.createSystem(implicitSolvent=app.OBC1)
structure = parmed.openmm.load_topology(prmtop.topology, system)
```
should *not* cause the ParmEd warning
```
/home/travis/miniconda/envs/_test/lib/python2.7/site-packages/parmed/openmm/topsystem.py:429: OpenMMWarning: Detected incomplete exceptions. Not supported.
```
if both the OpenMM `app` and `parmed` are functioning correctly.

jchodera:
I can reproduce this with the minimal test script, and opened an issue with the ParmEd issue tracker: https://github.com/ParmEd/ParmEd/issues/490

jchodera:
In the meantime I wonder if we should instead use an all-ParmEd route for loading Amber systems.

We could replace [these lines](https://github.com/choderalab/perses/blob/master/perses/tests/test_geometry_engine.py#L39-L41)
```python
    prmtop = app.AmberPrmtopFile(prmtop_file)
    system = prmtop.createSystem(implicitSolvent=app.OBC1)
    crd = app.AmberInpcrdFile(inpcrd_file)
```
with something like
```python
    from parmed import AmberParm
    prmtop = AmberParm(prmtop_file)
    system = prmtop.createSystem(implicitSolvent=app.OBC1)
    crd = app.AmberInpcrdFile(inpcrd_file)
```

pgrinaway:
We could. I'm still trying to figure out what is going on with the coordinate conversion. This version:

```python
def _rotation_matrix(axis, angle):
    """
    This method produces a rotation matrix given an axis and an angle.
    """
    axis = axis/np.linalg.norm(axis)
    axis_squared = np.square(axis)
    cos_angle = np.cos(angle)
    sin_angle = np.sin(angle)
    rot_matrix_row_one = np.array([cos_angle+axis_squared[0]*(1-cos_angle), 
                                   axis[0]*axis[1]*(1-cos_angle) - axis[2]*sin_angle,
                                   axis[0]*axis[2]*(1-cos_angle)+axis[1]*sin_angle])
    
    rot_matrix_row_two = np.array([axis[1]*axis[0]*(1-cos_angle)+axis[2]*sin_angle,
                                  cos_angle+axis_squared[1]*(1-cos_angle),
                                  axis[1]*axis[2]*(1-cos_angle) - axis[0]*sin_angle])
    
    rot_matrix_row_three = np.array([axis[2]*axis[0]*(1-cos_angle)-axis[1]*sin_angle,
                                    axis[2]*axis[1]*(1-cos_angle)+axis[0]*sin_angle,
                                    cos_angle+axis_squared[2]*(1-cos_angle)])
    
    rotation_matrix = np.array([rot_matrix_row_one, rot_matrix_row_two, rot_matrix_row_three])
    return rotation_matrix

def _cartesian_to_internal(atom_position, bond_position, angle_position, torsion_position):
            """
            Cartesian to internal function (hah)
            """
            a = bond_position - atom_position
            b = angle_position - bond_position
            #3-4 bond
            c = angle_position - torsion_position
            a_u = a / np.linalg.norm(a)
            b_u = b / np.linalg.norm(b)
            c_u = c / np.linalg.norm(c)

            #bond length
            r = np.linalg.norm(a)

            #bond angle
            theta = np.arccos(np.dot(-a_u, b_u))

            #torsion angle
            plane1 = np.cross(a, b)
            plane2 = np.cross(b, c)

            cos_phi = np.dot(plane1, plane2) / (np.linalg.norm(plane1)*np.linalg.norm(plane2))
            if cos_phi < -1.0:
                cos_phi = -1.0
            elif cos_phi > 1.0:
                cos_phi = 1.0
            
            phi = np.arccos(cos_phi)
            
            if np.dot(np.cross(plane1, plane2), b_u) > 0:
                phi = -phi
            
            if np.isnan(phi):
                raise Exception("phi is nan")

            return np.array([r, theta, phi])

def _internal_to_cartesian(bond_position, angle_position, torsion_position, r, theta, phi):
            a = angle_position - bond_position
            b = angle_position - torsion_position

            a_u = a / np.linalg.norm(a)
            b_u = b / np.linalg.norm(b)


            d_r = r*a_u

            normal = np.cross(a, b)

            #construct the angle rotation matrix
            angle_axis = normal / np.linalg.norm(normal)
            angle_rotation_matrix = _rotation_matrix(angle_axis, theta)
            
            #apply it
            d_ang = np.dot(angle_rotation_matrix, d_r)

            #construct the torsion rotation matrix and apply it
            torsion_axis = a_u
            torsion_rotation_matrix = _rotation_matrix(torsion_axis, phi)
            #apply it
            d_torsion = np.dot(torsion_rotation_matrix, d_ang)
            
            #add the positions of the bond atom
            xyz = bond_position + d_torsion
            return xyz
```

when run with this code:

```python
    example_coordinates = np.random.normal(size=[100,3])
    #try to transform random coordinates to and from
    for i in range(20):
        indices = np.random.randint(100, size=4)
        atom_position = example_coordinates[indices[0]]
        bond_position = example_coordinates[indices[1]]
        angle_position = example_coordinates[indices[2]]
        torsion_position = example_coordinates[indices[3]]
        rtp = _cartesian_to_internal(atom_position, bond_position, angle_position, torsion_position)
        r = rtp[0]
        theta = rtp[1]
        phi = rtp[2]
        xyz = _internal_to_cartesian(bond_position, angle_position, torsion_position, r, theta, phi)
        #print(np.linalg.norm(atom_position))
        print(np.linalg.norm(xyz-atom_position))
```
returns really small deviations between the original and the recalculated cartesian coordinates...around 1e-16 or less (the norms of the xyz vectors are around 1). However, by replacing all of those with unit-bearing quantities, I get a difference in several nanometers, on the same order as the xyz vectors themselves.

pgrinaway:
I just pushed a file `coordinate_tools.py` which contains the same functions, but with units. Any idea what I have missed here?

jchodera:
Looking!

jchodera:
Maybe it would be useful to start writing some unit tests for the individual components?

I'm still reading through, and `_rotation_matrix()` looks good to me, for example but the way to catch these bugs is to narrow them down with unit tests for individual components in parallel with trying to find them.

perses/rjmc/coordinate_tools.py
```diff
@@ -0,0 +1,108 @@
+import numpy as np
+import simtk.unit as units
+
+def _rotation_matrix(axis, angle):
```
jchodera: Later, we might think of moving to a quaternion representation to considerably simplify things.

pgrinaway:
I agree completely, many more unit tests are needed for this component. 

It looks like for this particular bug, the non-unit code is working, but the unit-bearing code is not, so I also think the rotation matrix code is good. I'm not sure what's going on with the unit version, though...

perses/rjmc/coordinate_tools.py
```diff
@@ -0,0 +1,108 @@
...
+def _internal_to_cartesian(bond_position, angle_position, torsion_position, r, theta, phi):
+            a = angle_position - bond_position
+            b = angle_position - torsion_position
+
+            a_u = a / units.norm(a)
+
+
+            d_r = r*a_u
+
+            normal = np.cross(a, b)
```
jchodera: `unit.cross`

perses/rjmc/coordinate_tools.py
```diff
@@ -0,0 +1,108 @@
...
+            c_u = c / units.norm(c)
+
+            #bond length
+            r = units.norm(a)
+
+            #bond angle
+            theta = units.acos(units.dot(-a_u, b_u))
+
+            #torsion angle
+            plane1 = np.cross(a, b)
```
jchodera: You either want
```python
plane1 = units.cross(a, b)
plane2 = units.cross(b,c)
```
or
```python
plane1 = np.cross(a_u, b_u)
plane2 = np.cross(b_u, c_u)
```

perses/rjmc/coordinate_tools.py
```diff
@@ -0,0 +1,108 @@
...
+
+            a_u = a / units.norm(a)
+
+
+            d_r = r*a_u
+
+            normal = np.cross(a, b)
+
+            #construct the angle rotation matrix
+            angle_axis = normal / units.norm(normal)
```
jchodera: `normal` will not have units, but I don't think it hurts to use `units.norm` here.

perses/rjmc/coordinate_tools.py
```diff
@@ -0,0 +1,108 @@
...
+            angle_rotation_matrix = _rotation_matrix(angle_axis, theta)
+
+            #apply it
+            d_ang = units.dot(angle_rotation_matrix, d_r)
```
jchodera: Can you even dot a unitless matrix (`angle_rotation_matrix`) with a unit-bearing quantity (`d_r`)?
I get an error when I try to do this manually.

jchodera:
I'm pretty confused here by lines like ` d_torsion = units.dot(torsion_rotation_matrix, d_ang)`, which dot a unitless rotation matrix with a unit-bearing quantity.

perses/rjmc/coordinate_tools.py
pgrinaway: Ok, giving that a try. Interesting that it does let me `np.cross` two unit-bearing quantities, but that does seem weird...

jchodera: That probably just drops the units silently, which is almost certainly not what you want.

perses/rjmc/coordinate_tools.py
pgrinaway: `d_ang` is supposed to be unit-bearing, yes.

jchodera: How does that work then?  Here's what I get:
```python
>>> angle_rotation_matrix = np.array([[1,0,0],[0,1,0],[0,0,1]], np.float32)
>>> d_r = 3 * unit.angstrom
>>> unit.dot(angle_rotation_matrix, d_r)
  File "<stdin>", line 1, in <module>
  File "/Users/choderaj/anaconda/lib/python2.7/site-packages/simtk/unit/unit_math.py", line 178, in dot
    sum = x[0]*y[0]
  File "/Users/choderaj/anaconda/lib/python2.7/site-packages/simtk/unit/quantity.py", line 733, in __getitem__
    assert not is_quantity(self._value[key])
TypeError: 'int' object has no attribute '__getitem__'
```

pgrinaway:
I agree that that's weird, but I'm not sure what units the rotation matrix should have.

perses/rjmc/coordinate_tools.py
pgrinaway: It does let me dot the two, and gives me a result in `nm`.

jchodera: But how?  Can you see if your rotation matrix ends up having units?

pgrinaway: In fact, this works for me:
```python
import simtk.unit as units
import numpy as np

random_mat = np.random.normal(size=[3,3])
random_vec = units.Quantity(np.random.normal(size=3), unit=units.nanometer)
print(units.dot(random_mat, random_vec))
```
which yields:
`[ 0.23290082 -2.70019196 -1.7732533 ] nm`

jchodera: Huh!  That works for me too.

pgrinaway: Sorry for the spam, but I think I may have found the problem. Continuing from the above example (with the same variable values)
```python
print(units.dot(random_mat, random_vec/random_vec.unit))
```
yields the same as above, but
```python
print(np.dot(random_mat, random_vec/random_vec.unit))
```
yields
`[ 1.45751018 -0.444677    0.51476441]`

jchodera: Oh, my.

pgrinaway:
Removing the `unit` operations did seem to fix the `coordinate_tools.py` functions.

pgrinaway:
Ok, update:

* I have corrected the `cartesian_to_internal` code.

* It looks like the `internal_to_cartesian` code is where the problem is.

jchodera:
Progress!

jchodera:
Could you check in your fixes when you get a chance?

pgrinaway:
Yeah, sorry--forgot to do that. I still have to clean up the tests, but it looks like the coordinate conversion actually works now. I constructed a minimal OpenMM system of four particles, and added custom bond, angle, and torsion forces to them, where the potential was just the bond, angle, and torsion angle, respectively. 

* The `_cartesian_to_internal` code matches what OpenMM calculates, and the `_internal_to_cartesian` results in OpenMM seeing the same `(r, theta, phi)` as originally given. 

* Additionally, the code can reconstruct existing coordinates by going from cartesian to internal and back.

jchodera:
Hooray!  So things might work now for actual molecules?

pgrinaway:
Unfortunately, that doesn't seem to be the case. I'm not sure what the deal is here--the coordinates are clearly converted correctly during the proposal, and the potentials seem to be implemented correctly. Is it possible that we can't capture the geometry of a methyl group by proposing only one atom at a time?

jchodera:
Have you tried some totally planar molecules?  Like toluene to phenol,
where an -H is replaced with an -OH?

pgrinaway:
I did try it, yeah. That is also problematic, though possibly due to ignoring improper torsions in the potential. We should probably, as you said earlier, add these in, but I think maybe a refactor should precede that.

jchodera:
Ethene to ethene-1,2-diol would  be free of the improper issue.

jchodera:
We've still got a bug somewhere.

I am running `perses/tests/test_geometry_engine.py` and printing out the angles from `_angle_logq()` to other atoms as the torsion is cranked around `phi` (see below).

It looks like
* The constant angle---the one we should have drawn, is 67 degrees, but should be within a few degrees of `theta0 = 110` degrees.
* Looking through the whole list, there should be some case where all the angles are ~108 degrees, but no such `phi` is present.
```
phi = 320.4 deg
     theta = 67.1671869485 deg, theta0 = 110.050047451 deg, u_theta = 43.570751
     theta = 163.287622503 deg, theta0 = 108.350046723 deg, u_theta = 60.807370
     theta = 91.6814945347 deg, theta0 = 108.350046723 deg, u_theta = 5.597743
phi = 324.0 deg
     theta = 67.1671869485 deg, theta0 = 110.050047451 deg, u_theta = 43.570751
     theta = 159.969868976 deg, theta0 = 108.350046723 deg, u_theta = 53.684664
     theta = 88.6167925071 deg, theta0 = 108.350046723 deg, u_theta = 7.845392
phi = 327.6 deg
     theta = 67.1671869485 deg, theta0 = 110.050047451 deg, u_theta = 43.570751
     theta = 156.652525546 deg, theta0 = 108.350046723 deg, u_theta = 47.006300
     theta = 85.584753581 deg, theta0 = 108.350046723 deg, u_theta = 10.441520
phi = 331.2 deg
     theta = 67.1671869485 deg, theta0 = 110.050047451 deg, u_theta = 43.570751
     theta = 153.337070837 deg, theta0 = 108.350046723 deg, u_theta = 40.774793
     theta = 82.5891419396 deg, theta0 = 108.350046723 deg, u_theta = 13.370248
phi = 334.8 deg
     theta = 67.1671869485 deg, theta0 = 110.050047451 deg, u_theta = 43.570751
     theta = 150.024590927 deg, theta0 = 108.350046723 deg, u_theta = 34.991209
     theta = 79.6342293237 deg, theta0 = 108.350046723 deg, u_theta = 16.613441
phi = 338.4 deg
     theta = 67.1671869485 deg, theta0 = 110.050047451 deg, u_theta = 43.570751
     theta = 146.715991429 deg, theta0 = 108.350046723 deg, u_theta = 29.655758
     theta = 76.7248732788 deg, theta0 = 108.350046723 deg, u_theta = 20.150372
phi = 342.0 deg
     theta = 67.1671869485 deg, theta0 = 110.050047451 deg, u_theta = 43.570751
     theta = 143.412097073 deg, theta0 = 108.350046723 deg, u_theta = 24.768052
     theta = 73.8666064805 deg, theta0 = 108.350046723 deg, u_theta = 23.957330
phi = 345.6 deg
     theta = 67.1671869485 deg, theta0 = 110.050047451 deg, u_theta = 43.570751
     theta = 140.113704504 deg, theta0 = 108.350046723 deg, u_theta = 20.327232
     theta = 71.0657376998 deg, theta0 = 108.350046723 deg, u_theta = 28.007183
phi = 349.2 deg
     theta = 67.1671869485 deg, theta0 = 110.050047451 deg, u_theta = 43.570751
     theta = 136.821613765 deg, theta0 = 108.350046723 deg, u_theta = 16.332023
     theta = 68.3294642881 deg, theta0 = 108.350046723 deg, u_theta = 32.268891
phi = 352.8 deg
     theta = 67.1671869485 deg, theta0 = 110.050047451 deg, u_theta = 43.570751
     theta = 133.536649488 deg, theta0 = 108.350046723 deg, u_theta = 12.780752
     theta = 65.6659948992 deg, theta0 = 108.350046723 deg, u_theta = 36.706968
phi = 356.4 deg
     theta = 67.1671869485 deg, theta0 = 110.050047451 deg, u_theta = 43.570751
     theta = 130.259677014 deg, theta0 = 108.350046723 deg, u_theta = 9.671355
     theta = 63.0846792812 deg, theta0 = 108.350046723 deg, u_theta = 41.280917
```

jchodera:
Here's the first proposal, which shows that angle (1,3,2) is proposed as 114 degrees but then measured to be 65 degrees in the subsequent energy calculations. This is our primary problem.
```
PROPOSED ANGLE
    1     3     2 : theta = 114.442511463 deg, theta0 = 110.630047471 deg
sigma_theta = 3.93456639833 deg, theta_proposed = 114.442511463 deg
     theta = 114.442511463 deg, theta0 = 110.630047471 deg, u_theta = 0.469448
phi = 0.0 deg
    1     3     2 :       theta = 65.5574885365 deg, theta0 = 110.630047471 deg, u_theta = 65.614629
    1     3    12 :       theta = 122.094925262 deg, theta0 = 110.050047451 deg, u_theta = 3.437418
    1     3    13 :       theta = 128.08734777 deg, theta0 = 110.050047451 deg, u_theta = 7.708513
phi = 3.6 deg
    1     3     2 :       theta = 65.5574885365 deg, theta0 = 110.630047471 deg, u_theta = 65.614629
    1     3    12 :       theta = 125.33750595 deg, theta0 = 110.050047451 deg, u_theta = 5.537301
    1     3    13 :       theta = 124.846376294 deg, theta0 = 110.050047451 deg, u_theta = 5.187230
phi = 7.2 deg
    1     3     2 :       theta = 65.5574885365 deg, theta0 = 110.630047471 deg, u_theta = 65.614629
    1     3    12 :       theta = 128.589171328 deg, theta0 = 110.050047451 deg, u_theta = 8.143404
    1     3    13 :       theta = 121.615624212 deg, theta0 = 110.050047451 deg, u_theta = 3.169291
```

perses/rjmc/geometry.py
```diff
@@ -650,10 +646,15 @@ def _calculate_angle(self, atom_position, bond_atom_position, angle_atom_positio
         b = angle_atom_position - bond_atom_position
         a_u = a / np.linalg.norm(a)
         b_u = b / np.linalg.norm(b)
-        theta = np.arccos(np.dot(a_u, b_u))
+        cos_theta = np.dot(a_u, b_u)
```
jchodera: This is the bug.  Because of the way you define `a` and `b` vectors, this should be
```
cos_theta = np.dot(-a_u, b_u)
```

pgrinaway:
Looks like some unit thing is a problem in one of the tests. Working on that.

perses/tests/test_geometry_engine.py
```diff
@@ -90,11 +108,220 @@ def test_run_geometry_engine():
...
+    geometry_engine = geometry.FFAllAngleGeometryEngine({'test': 'true'})
+    example_coordinates = units.Quantity(np.random.normal(size=[100,3]), unit=units.nanometers)
+    #try to transform random coordinates to and from
+    for i in range(200):
+        indices = np.random.randint(100, size=4)
+        atom_position = units.Quantity(np.array([ 0.80557722 ,-1.10424644 ,-1.08578826]), unit=units.nanometers)
+        bond_position = units.Quantity(np.array([ 0.0765,  0.1  ,  -0.4005]), unit=units.nanometers)
+        angle_position = units.Quantity(np.array([ 0.0829 , 0.0952 ,-0.2479]) ,unit=units.nanometers)
+        torsion_position = units.Quantity(np.array([-0.057 ,  0.0951 ,-0.1863] ) ,unit=units.nanometers)
+        rtp, detJ = geometry_engine._cartesian_to_internal(atom_position, bond_position, angle_position, torsion_position)
```
jchodera: Your documentation for `_cartesian_to_internal` says:
```
 Returns without units!
```
but then you feed the unitless `r` into `_internal_to_cartesian` which immediately removes units from `r`, assuming it is unit-bearing.

We will need some way to be more consistent with units in a refactoring PR.

pgrinaway: >We will need some way to be more consistent with units in a refactoring PR.

Agreed. The error in the current build is due to a different issue with units.

pgrinaway:
Ok, right now it looks like the only problem in the tests is the `test_rotation_matrix`. I'll rewrite that test then commit, and merge this PR. Next PR will clean things up and add the rest of the unit tests.

pgrinaway:
Also need to restore the Jacobian calculation--leaving this here as a note to myself.

pgrinaway:
Passes! Going to merge.

PR #56, opened 2015-12-10
[WIP] Unit tests, impropers, python 3, etc.

pgrinaway:
This PR will contain a collection of new things:

* Python 3 support (mostly done)

* Full set of unit tests (partly done, high priority)

* Support for calculation of improper torsions

* Type hint stubs (`*.pyi`) for use in static type analysis

This is very much a WIP, so if there are objections to this list, or things that should be added in this PR, definitely post about it.

### DIFF
--- a/.idea/dictionaries/grinawap.xml
+++ b/.idea/dictionaries/grinawap.xml
@@ -1,0 +1,3 @@
+<component name="ProjectDictionaryState">
+  <dictionary name="grinawap" />
+</component>
--- a/perses/rjmc/geometry.py
+++ b/perses/rjmc/geometry.py
@@ -734,7 +734,7 @@ class SystemFactory(object):
     _PeriodicTorsionForceEnergy = "step(growth_idx - {})*k*(1+cos(periodicity*theta-phase))"
 
 
-    def create_modified_system(self, reference_system, growth_index, parameter_name, force_names, force_parameters):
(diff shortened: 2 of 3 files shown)


jchodera:
Sounds like a good list!

pgrinaway:
Added the test for calculating torsion potentials--turned out the `GeometryEngine` had been using a `logq` that was twice OpenMMs (the factor of two thing that you had mentioned), so I fixed the computation of `_torsion_logq`. I realize the algorithm is technically correct (since the same equation was being used to calculate `logq`), but having the code do what we want is critical, and matching the energies could help with better proposals.

pgrinaway:
Also, I was using an existing molecule (butane) to test the torsion potential. I'm going to change that test to just rotate an arbitrary particle and calculate the torsion energy.

jchodera:
Great!

pgrinaway:
Ok, so given the recent discussion, the focus of this PR will shift to:

* Unit tests

* Python 3 support (done)

* Calculating torsion scan potentials with OpenMM
    * Constructing the appropriate OpenMM system
    * Pre-enumeration of atoms to have positions proposed+the order of proposal.

jchodera:
Great!

jchodera:
Is this PR still in process?

pgrinaway:
I think given the changes to its original purpose, as well as ongoing discussion, I'm going to close it and restart this process.

PR #64, opened 2016-01-14
Initial PolymerProposalEngine

juliebehr:
Does not select mutants; just constructs new topologies assuming mutations specified in metadata are exactly what need to be made


jchodera:
Yay!

jchodera:
Where did the PDB templates come from?  Do we need those for the atom names/order, or the geometry?

perses/rjmc/topology_proposal.py
```diff
@@ -1,9 +1,17 @@
 """
 This file contains the base classes for topology proposals
 """
+### implement updates to TopologyProposal
+### calculate logp in different ways in different subclasses
 
 import simtk.openmm.app as app
 from collections import namedtuple
+<<<<<<< HEAD
```
pgrinaway: I think this is causing the build failure on travis

pgrinaway: Oh, there's also the end part of that a few lines down...

jchodera:
Perhaps we could use information from the ForceField itself, since we don't need the geometry here? That way, we could make sure this works for any polymer rather than just the residues for which we have PDBs generated here.

pgrinaway:
>Perhaps we could use information from the ForceField itself, since we don't need the geometry here? That way, we could make sure this works for any polymer rather than just the residues for which we have PDBs generated here.

+1

jchodera:
We may have to dig into the private ForceField API a bit for that, but we can expose things via the public API as we figure out what the best way to do that is.

Note that we want to be using openmm-dev if we aren't already since this will have the necessary ForceField changes.

perses/rjmc/topology_proposal.py
juliebehr: I thought I had found all of these and removed them but clearly missed a couple, thanks for the catch!

juliebehr:
Yes, I had intended to do that as well!  This was just simpler initially, and is directly taken from pdbfixer.

pgrinaway:
Do you think we can enable the Python 3 builds and tests now? It's a change to `.travis.yml` so that [this part](https://github.com/choderalab/perses/blob/master/.travis.yml#L37-L38) is:

```
    - python=3.4  CONDA_PY=34
    - python=3.5  CONDA_PY=35
```
(note that I've changed the 3.3 to 3.4 and 3.4 to 3.5.)

Thanks!

pgrinaway:
This failure is a result of the OpenEye licensing model, I think. We need the key to be present during the build, but it's only accessible from branches in the master repository.

pgrinaway:
What are we doing with this PR now?

jchodera:
> Do you think we can enable the Python 3 builds and tests now?

Go for it, but note the OEChem tools have a bug for python 3.5.

pgrinaway:
Interesting. I am running Python 3.5.1 as my primary interpreter on my laptop and I have no issue with OpenEye. Maybe it's a linux issue?

juliebehr:
Currently converting from template pdb directory to importing templates from ForceField; after that commit I think this PR will be ready

pgrinaway:
sounds good. I'll implement the remainder of the skeleton things after that.

juliebehr:
Not quite ready to merge: I'm going through and trying to make sure that all changes have been self-consistent in terms of how the "new" and "old" residues are referenced

jchodera:
I think we decided that you could add some nosetests and doctests of the `PointMutationEngine` and merge once those pass!

juliebehr:
Which histidine protonation state should be selected as the default?

jchodera:
Yikes---good question.  I guess we eventually want to include pKa capabilities as well.

For now, is there a way we can pick either of the singly-protonated HIS protomers for now?  I think these are `HID` and `HIE`?  That would solve 90% of the problem of not knowing which tautomer is the optimal one.

pgrinaway:
>For now, is there a way we can pick either of the singly-protonated HIS protomers for now? I think these are HID and HIE? That would solve 90% of the problem of not knowing which tautomer is the optimal one.

So we could just (for now) uniformly draw one of them?

jchodera:
> So we could just (for now) uniformly draw one of them?

Yes!  We just want to choose the residue first (HIS) and then choose one of the tautomers uniformly (HID/HIE).

jchodera:
We do want to make sure both tautomers (HID/HIE) count in the same `HIS` bin for free energy estimates, though---they shouldn't be counted separately as different species in the target weight \pi_j or the free energy estimates \zeta_j.

perses/tests/test_topology_proposal.py
```diff
@@ -0,0 +1,440 @@
...
+                # This is a possible match, so trying matching the rest of the residue.
+
+                matches[position] = i
+                hasMatch[i] = True
+                if _findAtomMatches(atoms, template, bondedTo, externalBonds, matches, hasMatch, position+1):
+                    return True
+                hasMatch[i] = False
+    return False
+
+def _match_template_and_res_name(topology, residue_map, forcefield):
```
juliebehr: This test is still currently broken -- I was trying to confirm that the new residue has been built correctly by using forcefield to find a matching template and then asserting that they have the same name (i.e. the residue has a match and also it's the correct template) by copying the first lines of ```createSystem()``` in forcefield.py where the template matches are found.  Something about this method isn't working, as it will find the ```signature``` in ```_templateSignatures``` [(Line 147)] (https://github.com/choderalab/perses/pull/64/files#diff-2551428e6650157ac591946d623ca04dR147) but then does not match the atoms ([Line 149](https://github.com/choderalab/perses/pull/64/files#diff-2551428e6650157ac591946d623ca04dR149): ```matches = None```).  There are then no ```generators``` in ```forcefield._templateGenerators``` which is why the loop at [Line 156] (https://github.com/choderalab/perses/pull/64/files#diff-2551428e6650157ac591946d623ca04dR156) does not lead to a ```template``` or ```Exception```.

I don't think this has to do with the topologies being built incorrectly such that a template match cannot be found, because I can successfully run ```ff.createSystem(new_topology)```.

(Note that the excessive assert and prints are just from me trying to pinpoint where the method was going wrong).

jchodera: Note that the template signatures matching onlymeans that you have the same composition of atoms; you can still subsequently fail matching in the `_findAtomMatches()` stage because the bonds don't match up.

perses/tests/test_topology_proposal.py
```diff
@@ -0,0 +1,440 @@
...
+    file.close()
+    file = StringIO(contents)
+
+    if _guessFileFormat(file, url) == 'pdbx':
+        pdbx = app.PDBxFile(contents)
+        topology = pdbx.topology
+        positions = pdbx.positions
+    else:    
+        structure = PdbStructure(file)
+        pdb = app.PDBFile(structure)
```
jchodera: You should just be able to do
```
pdb = app.PDBFile(file)
```
and get rid of the `structure` intermediate.

perses/tests/test_topology_proposal.py
```diff
@@ -0,0 +1,440 @@
...
+            file.seek(0)
+            return 'pdbx'
+        if line.startswith('HEADER') or line.startswith('REMARK') or line.startswith('TITLE '):
+            file.seek(0)
+            return 'pdb'
+    # It's certainly not a valid PDBx/mmCIF.  Guess that it's a PDB.
+    file.seek(0)
+    return 'pdb'
+
+def _matchResidue(res, template, bondedToAtom):
```
jchodera: I think you can simply import this from `simtk.openmm.app.forcefield`:
```python
from simtk.openmm.app.forcefield import _matchResidue
```
No need to duplicate it here unless you have changed the behavior.

jchodera:
Is the goal of the tests just to use the `ForceField` template matching machinery to read off the list of residues you have after the topology proposal?  If so, that's (1) really clever, and (2) a feature we can probably even incorporate into the `ForceField` class itself so that you can feed in a `Topology` object and get back a list of matching forcefield residue templates or template names.  I could see that being widely useful.

perses/tests/test_topology_proposal.py
juliebehr: The failure to find a match is happening inside here somewhere, which is why I copied it (originally was using app.forcefield._matchResdiue, as seen commented out in Line 154)

juliebehr:
Yes; unfortunately not as successfully as I'd hoped.

pgrinaway:
I'm getting a weird error when I try to run this:
```pytb
  File "/Users/grinawap/PycharmProjects/perses_juile/perses/tests/test_topology_proposal.py", line 437, in <module>
    test_mutate_from_every_amino_to_every_other()
  File "/Users/grinawap/PycharmProjects/perses_juile/perses/tests/test_topology_proposal.py", line 339, in test_mutate_from_every_amino_to_every_other
    modeller = pm_top_engine._add_new_atoms(modeller, missing_atoms, residue_map)
  File "/Users/grinawap/PycharmProjects/perses_juile/perses/rjmc/topology_proposal.py", line 675, in _add_new_atoms
    modeller.topology.addBond(new_bonded_0, new_bonded_1)
  File "/Users/grinawap/anaconda/lib/python3.5/site-packages/simtk/openmm/app/topology.py", line 168, in addBond
    self._bonds.append((atom1, atom2))
AttributeError: 'filter' object has no attribute 'append'
```

It's weird, because it's happening inside `app.Topology.addBond`, where the `_bonds` attribute seems to be a `filter` instead of a `list`.

pgrinaway:
For the record, the iteration index `k` is at 0, so this is the first iteration of the `_add_new_atoms`

perses/rjmc/topology_proposal.py
```diff
@@ -238,10 +299,394 @@ def propose(self, current_system, current_topology, current_positions, current_m
...
+        modeller : simtk.openmm.app.Modeller 
+            extra atoms from old residue have been deleted, missing atoms in new residue not yet added
+        """
+        # delete_atoms : list(simtk.openmm.app.topology.Atom) atoms from existing residue not in new residue
+        # atom : simtk.openmm.app.topology.Atom
+        for atom in delete_atoms:
+            atom.residue._atoms.remove(atom)
+            for bond in modeller.topology._bonds:
+                if atom in bond:
+                    modeller.topology._bonds = filter(lambda a: a != bond, modeller.topology._bonds)
```
pgrinaway: what about here? the list becomes a filter?

perses/rjmc/topology_proposal.py
```diff
@@ -238,10 +299,394 @@ def propose(self, current_system, current_topology, current_positions, current_m
...
+            for atom1, atom2 in modeller.topology._bonds:
+                if atom1.residue == residue or atom2.residue == residue:
+                    old_res_bonds.append((atom1.name, atom2.name))
+            # make a list of bonds that should exist in new residue
+            # template_bonds : list(tuple(str (atom name), str (atom name))) bonds in template
+            template_bonds = [(template.atoms[bond[0]].name, template.atoms[bond[1]].name) for bond in template.bonds]
+            # add any bonds that exist in template but not in new residue
+            for bond in old_res_bonds:
+                if bond not in template_bonds and (bond[1],bond[0]) not in template_bonds:
+                    modeller.topology._bonds = filter(lambda a: a != bond, modeller.topology._bonds)
```
pgrinaway: and here. I think this is the problem?

juliebehr: Ughhhhhh python 2 vs 3 problem, good catch -- this also won't solve the test thing, since it was still ```_bonds``` was still a list for me

jchodera:
See https://github.com/pandegroup/openmm/pull/1359 for new function added to help support your tests.

This would be used as:
```python
templates = forcefield.getMatchingTemplates(new_topology)
assert(template[index].name == expected_template_name)
```

juliebehr:
Awesome!

jchodera:
https://github.com/pandegroup/openmm/pull/1359 was merged, so you should be able to use it for your test @juliebehr

juliebehr:
@jchodera great, updating openmm and adding it to the test now!

jchodera:
You'll need to switch to `openmm-dev` if you haven't already.

jchodera:
The nightly openmm-dev build broke
Trying to figure out what went wrong.

jchodera:
Fixed!

juliebehr:
As far as I can tell (unless there are additional things that should be tested) this PR is now complete.

pgrinaway:
Runs fine on my local computer.

PR #68, opened 2016-01-22
Basic run test, etc.

pgrinaway:
Apparently, I forgot there was a `run_example.py` in the `examples/` folder. I'm going to restart building the skeleton thing based on that. I've put it here as a test, and will refactor it to also use empty classes (nonetheless, it will be nice to have an example that actually does something). In the process of that, I noticed a couple things:

* `NCMCEngine` is directly derived from `object`. There is nothing wrong with that, other than that it can't be used in the skeleton. So, I've added a base class--I will flesh it out as needed to make the skeleton work.

* The tests are now failing--looks like there's something wrong with the `GeometryEngine` test itself. I will remedy that in this PR.

* I'm going to try to get the openeye tests to be skipped for PRs from forks.

* Since this is the "skeleton" PR, it would be good if the API were finalized here. 

Anything else?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ script:
   - source activate _test
   - conda install --yes --quiet pip nose nose-timer
   - pip install $OPENEYE_CHANNEL openeye-toolkits
-  - cd devtools && nosetests $PACKAGENAME --nocapture --verbosity=3 --with-doctest --with-timer && cd ..
+  #- cd devtools && nosetests $PACKAGENAME --nocapture --verbosity=3 --with-doctest --with-timer && cd ..
 env:
   matrix:
     - python=2.7  CONDA_PY=27 OPENEYE_CHANNEL="-i https://pypi.anaconda.org/openeye/channel/main/simple"
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ env:
   matrix:
     - python=2.7  CONDA_PY=27 OPENEYE_CHANNEL="-i https://pypi.anaconda.org/openeye/channel/main/simple"
     - python=3.4  CONDA_PY=34 OPENEYE_CHANNEL="-i https://pypi.anaconda.org/openeye/channel/main/simple"
-    - python=3.5  CONDA_PY=35 OPENEYE_CHANNEL="-i https://pypi.anaconda.org/openeye/channel/main/simple"
+    #- python=3.5  CONDA_PY=35 OPENEYE_CHANNEL="-i https://pypi.anaconda.org/openeye/channel/main/simple"
 
   global:
     - ORGNAME="omnia"
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,12 +30,12 @@ script:
   - source activate _test
   - conda install --yes --quiet pip nose nose-timer
   - pip install $OPENEYE_CHANNEL openeye-toolkits
-  - cd devtools && nosetests $PACKAGENAME --nocapture --verbosity=2 --with-doctest --with-timer && cd ..
+  - cd devtools && nosetests $PACKAGENAME --nocapture --verbosity=3 --with-doctest --with-timer && cd ..
 env:
   matrix:
     - python=2.7  CONDA_PY=27 OPENEYE_CHANNEL="-i https://pypi.anaconda.org/openeye/channel/main/simple"
-    - python=3.4  CONDA_PY=34
-    - python=3.5  CONDA_PY=35
+    - python=3.4  CONDA_PY=34 OPENEYE_CHANNEL="-i https://pypi.anaconda.org/openeye/channel/main/simple"
+    - python=3.5  CONDA_PY=35 OPENEYE_CHANNEL="-i https://pypi.anaconda.org/openeye/channel/main/simple"
 
   global:
     - ORGNAME="omnia"
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ script:
   - source activate _test
   - conda install --yes --quiet pip nose nose-timer
   - pip install $OPENEYE_CHANNEL openeye-toolkits
-  #- cd devtools && nosetests $PACKAGENAME --nocapture --verbosity=3 --with-doctest --with-timer && cd ..
+  - cd devtools && nosetests $PACKAGENAME --nocapture --verbosity=3 --with-doctest --with-timer && cd ..
 env:
   matrix:
     - python=2.7  CONDA_PY=27 OPENEYE_CHANNEL="-i https://pypi.anaconda.org/openeye/channel/main/simple"
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -44,6 +44,9 @@ test:
     - pymbar
   imports:
     - perses
+  
+  commands:
+    - nosetests perses -vvv
 
 about:
   home: https://github.com/choderalab/perses
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -45,8 +45,6 @@ test:
   imports:
     - perses
   
-  commands:
-    - nosetests perses -vvv
 
 about:
   home: https://github.com/choderalab/perses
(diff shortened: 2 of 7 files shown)


pgrinaway:
Hm, seems to be in `prmtop.createSystem`

jchodera:
For the "empty skeleton", we can have parts of the engine that actually do things (because we implemented them!) actually do them.  The rest of the "empty" bits can just pass around their input objects without modifying them.  So it doesn't have to be totally empty---you could still have NCMC do its thing, for example, but in changing one molecule into the same molecule.

pgrinaway:
Ok, will change that back.

pgrinaway:
Update: I am getting `nan`s from the NCMC bits.. (before the `GeometryEngine`). Investigating..

jchodera:
I don't see the `NaN`s in travis.  What test are you observing `NaN`s in?

pgrinaway:
The one that I'm adding (`test_run_example`). I'm seeing them locally. I think there's a different error in travis (easier to solve). Looks like the `nan`s are not the `NCMCEngine`'s fault, though. 100 steps of verlet causes the positions to be `nan`.

jchodera:
Can you check in the test? Would be useful to see it fail on Travis too.

I can take a look in a bit.

pgrinaway:
it should be checked in...

pgrinaway:
@jchodera any idea what [this](https://travis-ci.org/choderalab/perses/jobs/104953007#L885) might be?

`ImportError: cannot import name convertParameters`.

jchodera:
I bet that is related to this: https://github.com/pandegroup/openmm/pull/1311

jchodera:
Oh!  That's a problem with importing `parmed`.  I wonder if @swails has insight here.

jchodera:
I'm not sure what is bringing in `parmed` anyway.  It's not in the `meta.yaml` conda requirements.

jchodera:
It looks like `perses/tests/test_geometry_engine.py` uses `parmed`, but neither [`devtools/conda-recipe/meta.yaml`](https://github.com/choderalab/perses/blob/master/devtools/conda-recipe/meta.yaml) nor [`setup.py`](https://github.com/choderalab/perses/blob/master/setup.py#L81-L89) specify `parmed` as a dependency.  Can you add this @pgrinaway?

pgrinaway:
Looking into that.. thanks! Additionally, can anyone spot what could be wrong with this minimal example?
```python
import numpy as np
from simtk import unit, openmm
from simtk.openmm import app
import openeye.oechem as oechem
import openeye.oeomega as oeomega
import openmoltools
import logging
import copy
import logging



def generate_initial_molecule(mol_smiles):
    """
    Generate an oemol with a geometry
    """
    mol = oechem.OEMol()
    oechem.OESmilesToMol(mol, mol_smiles)
    oechem.OEAddExplicitHydrogens(mol)
    omega = oeomega.OEOmega()
    omega.SetMaxConfs(1)
    omega(mol)
    return mol

def oemol_to_openmm_system(oemol, molecule_name):
    """
    Create an openmm system out of an oemol

    Returns
    -------
    system : openmm.System object
        the system from the molecule
    positions : [n,3] np.array of floats
    """
    _ , tripos_mol2_filename = openmoltools.openeye.molecule_to_mol2(oemol, tripos_mol2_filename=molecule_name + '.tripos.mol2', conformer=0, residue_name='MOL')
    gaff_mol2, frcmod = openmoltools.amber.run_antechamber(molecule_name, tripos_mol2_filename)
    prmtop_file, inpcrd_file = openmoltools.utils.run_tleap(molecule_name, gaff_mol2, frcmod)
    prmtop = app.AmberPrmtopFile(prmtop_file)
    system = prmtop.createSystem(implicitSolvent=app.OBC1)
    crd = app.AmberInpcrdFile(inpcrd_file)
    return system, crd.getPositions(asNumpy=True)

logger = logging.getLogger('antechamber')
logger.setLevel(logging.INFO)

mol = generate_initial_molecule("CCCCC")
sys, pos = oemol_to_openmm_system(mol, "pentane")

integrator = openmm.VerletIntegrator(300*unit.kelvin)
context = openmm.Context(sys, integrator)
context.setPositions(pos)
context.setVelocitiesToTemperature(300*unit.kelvin)
initial_state = context.getState(getEnergy=True, getPositions=True, getVelocities=True, getForces=True, getParameters=True) #everything is fine

#take one step:
integrator.step(1)
step_state = context.getState(getEnergy=True, getPositions=True, getVelocities=True, getForces=True, getParameters=True)

print(step_state.getPotentialEnergy()) #NaN

**Comment by pgrinaway** [comment]
I've tried the three platforms on my laptop (Reference, CPU, and OpenCL), and the only difference is the rate at which they `nan`.

**Comment by jchodera** [comment]
You don't actually generate a conformation in there anywhere.

**Comment by pgrinaway** [comment]
Ok, the mistake was stupidly setting the timestep to `300*unit.kelvin`. That is solved...

**Comment by swails** [comment]
That's a hot timestep.

**Comment by pgrinaway** [comment]
:fire:

**Comment by jchodera** [comment]
It looks like the GBSA model changes in OpenMM from https://github.com/pandegroup/openmm/pull/1311 broke [`alchemy`](https://github.com/choderalab/alchemy) involving GBSA.  I disabled the GBSA model in the test for now---vacuum should be good enough for testing.

I'm working on fixing `alchemy` now.

**Comment by jchodera** [comment]
There's still a test failure, but it seems to be related to a separate issue with the test code.

At some point it might be useful to change your `oemol_to_openmm_system` to use the new GAFF template generators instead of going through LEaP.  For example, you can use something like:
```python
def oemol_to_openmm_system(oemol, molecule_name):
    """
    Create an openmm system out of an oemol

    Returns
    -------
    system : openmm.System object
        the system from the molecule
    positions : [n,3] np.array of floats
    topology : openmm.app.Topology object
    """
    # Generate a topology.
    from openmoltools.forcefield_generators import generateTopologyFromOEMol
    topology = generateTopologyFromOEMol(molecule)

    # Initialize a forcefield with GAFF.
    from simtk.openmm.app import ForceField
    forcefield = ForceField('gaff.xml')

    # Generate template and parameters.
    from openmoltools.forcefield_generators import generateResidueTemplate
    [template, ffxml] = generateResidueTemplate(molecule)

    # Register the template.
    forcefield.registerResidueTemplate(template)

    # Add the parameters.
    forcefield.loadFile(StringIO(ffxml))

    # Create the system.
    system = forcefield.createSystem(topology)
    
    return (system, positions, topology)
    """

**Comment by pgrinaway** [comment]
Ok, will do. Thanks for pointing this out!

**Comment by jchodera** [comment]
Let me know if you get stuck again!  Would be great to get this running ASAP.

**Comment by pgrinaway** [comment]
Yep. Just sorting through dependency hell at the moment, will try to get this fixed today.

**Comment by pgrinaway** [comment]
Hm, still getting `nan` even without implicit solvent...

**Comment by pgrinaway** [comment]
Oh, I think I know the problem.

**Comment by pgrinaway** [comment]
Hm, nevermind... I thought perhaps I was accidentally adding in implicit solvent somewhere, but there is no GB force in the system...

**Comment by jchodera** [comment]
This error?
```
  File "perses/tests/test_sampler.py", line 155, in <module>
    test_run_example()
  File "perses/tests/test_sampler.py", line 110, in test_run_example
    assert np.sum(np.isnan(ncmc_old_positions)) == 0
AssertionError
```
or something else?

pgrinaway:
Yep, that error. The positions and energy are `nan` after NCMC...

jchodera:
What if you just leave `nswitching_steps = 0` for now (which you said works?) and I'll debug the NCMC switch further over the weekend?

pgrinaway:
Ok I'll do that for now and take care of the other errors first.

pgrinaway:
Ok, now the Python 2.7 build is passing, but not the others...

jchodera:
Close enough!  We can figure out the py3x issues separately!

jchodera:
Merge and move on?

pgrinaway:
Oops, just saw this. Pushed an attempt to change this--just want to see if that works. If it doesn't, I'll just merge and worry about this later.

pgrinaway:
Ok, that didn't work. Merging.

PR #70, opened 2016-01-30
Fix py34 issues

jchodera:
I noticed both `openmm` and `openmm-dev` were dependencies, as well as `parmed` and `parmed-dev`.  Trying to eliminate this and debug the py34 issues.


jchodera:
The py34 issue involves relative versus absolute imports:
http://blog.tankywoo.com/python/2013/10/07/python-relative-and-absolute-import.html

Fixing this now.

jchodera:
Looks like that did the trick!

Issue #77, opened 2016-01-31
Missing a contribution to sampler acceptance probability?


jchodera:
I think we're missing a critical contribution in the [acceptance criteria for `test_sampler`](https://github.com/choderalab/perses/blob/master/perses/tests/test_sampler.py#L157).  While we include the ratio of proposal probabilities in the geometry proposal, I don't believe we include the difference in potential energies between the old and new conformations.

This results in some pretty weird geometries when the trajectory is visualized.

jchodera:
Notably, this is the difference in potential energies for the alchemically eliminated forms of these molecules, so the majority of the difference in potential arises from the new valence terms (which should cancel to some degree with the ratio of proposal probabilities).

juliebehr:
I believe this was addressed in https://github.com/choderalab/perses/pull/223 here: https://github.com/choderalab/perses/commit/4bf4e7ef52694e560bf6b15f5bb96f276a453f24

jchodera:
I've actually completely lost track of what all the contributions are called in the code and how they map to actual contributions to the mathematically-derived acceptance probability. Perhaps we need to document that?

pgrinaway:
>Perhaps we need to document that?

Yep. I think we need to take a detailed pass through the code, ensuring everything is properly documented, once we clear out some of the smaller issues. This will help identify:

* Test omissions
* API flaws
* Undocumented code
* Bugs in implementations
* Problems with code readability

I'm happy to start doing this as soon as the smaller issues are closed out.

jchodera:
I think @pgrinaway and @juliebehr said this was fixed in our last meeting. Reopen if needed.

PR #78, opened 2016-01-31
Implement additional contribution to acceptance probability

jchodera:
Attempting to address #77.  NCMC switching now returns a potential energy that can be used in computing this additional contribution.

Currently, this contribution is essentially uniformly unfavorable, leading to universal rejection.  This might be a real problem, or might be a dumb mistake.

Very experimental.

### DIFF
--- a/perses/annihilation/ncmc_switching.py
+++ b/perses/annihilation/ncmc_switching.py
@@ -269,11 +269,15 @@ class NCMCEngine(object):
         context.applyVelocityConstraints(integrator.getConstraintTolerance())
         # Store initial potential if 'insert'
         if direction == 'insert':
+            for parameter_name in available_parameters:
+                context.setParameter(parameter_name, 0)
             potential = context.getState(getEnergy=True).getPotentialEnergy()
         # Only take a single integrator step since all switching steps are unrolled in NCMCAlchemicalIntegrator.
         integrator.step(1)
         # Store final potential if 'insert'
         if direction == 'delete':
+            for parameter_name in available_parameters:
+                context.setParameter(parameter_name, 0)
             potential = context.getState(getEnergy=True).getPotentialEnergy()
         # Store final positions and log acceptance probability.
         final_positions = context.getState(getPositions=True).getPositions(asNumpy=True)
--- a/perses/annihilation/ncmc_switching.py
+++ b/perses/annihilation/ncmc_switching.py
@@ -30,8 +30,8 @@ class NCMCEngine(object):
     >>> topology_proposal = TopologyProposal(old_system=testsystem.system, old_topology=testsystem.topology, old_positions=testsystem.positions, new_system=testsystem.system, new_topology=testsystem.topology, logp_proposal=0.0, new_to_old_atom_map=new_to_old_atom_map, metadata=dict())
     >>> ncmc_engine = NCMCEngine(temperature=300.0*unit.kelvin, functions=default_functions, nsteps=50, timestep=1.0*unit.femtoseconds)
     >>> positions = testsystem.positions
-    >>> [positions, logP_delete] = ncmc_engine.integrate(topology_proposal, positions, direction='delete')
-    >>> [positions, logP_insert] = ncmc_engine.integrate(topology_proposal, positions, direction='insert')
+    >>> [positions, logP_delete, potential_delete] = ncmc_engine.integrate(topology_proposal, positions, direction='delete')
+    >>> [positions, logP_insert, potential_insert] = ncmc_engine.integrate(topology_proposal, positions, direction='insert')
 
     """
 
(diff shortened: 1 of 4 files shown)


jchodera:
On further reflection, I think this contribution is correct, but I don't think it should be biasing the simulation toward rejection quite as much as it seems to.

The overall contribution comes from
```
P(old | new) P(new)   
--------------------
P(new | old) P(old)
```
If we assume there is only one particle being deleted and another being inserted, we have
```
P(r_old, theta_old, phi_old | x_core) J(r_old, theta_old, phi_old) exp[-u_core(x_core)] exp[-u_new(r_new, theta_new, phi_new | x_core)]
---------------------------------------------------------------------------------------------------------------------------------------
P(r_new, theta_new, phi_new | x_core) J(r_new, theta_new, phi_new) exp[-u_core(x_core)] exp[-u_old(r_old, theta_old, phi_old | x_core)]
```
The `exp[-u_core(x_core)]` contributions obviously cancel.

The contributions involving old atoms are
```
P(r_old, theta_old, phi_old | x_core)
-----------------------------------------------    J(r_old, theta_old, phi_old)
exp[-u_old(r_old, theta_old, phi_old | x_core)]
```
Now we can see why we use the valence energy of these atoms to draw the positions.  This should be almost exactly
```
exp[-u_old(r_old, theta_old, phi_old | x_core)] / Z_old
-------------------------------------------------------    J(r_old, theta_old, phi_old)
exp[-u_old(r_old, theta_old, phi_old | x_core)]
```
where `Z_old` is the normalizing constant for that old atom, which should be approximately proportional to the number of atoms being added or deleted for larger transformations.  This simplifies to
```
J(r_old, theta_old, phi_old)
---------------------
Z_old
```
Putting it all together, we should end up with this contribution being something like
```
J(r_old, theta_old, phi_old)   Z_new
------------------------------- * ------
J(r_new, theta_new, phi_new)   Z_old
```
The left term should be of order unity, and the right term has something to do with the net change in the number of atoms.

We should be able to compute this factor in the geometry engine and see how close the `geometry_logp + switch_logp` comes to this.  They should be of the same order of magnitude and sign.  (`switch_logP` is given by `-(u_new - u_old)`.)

If not, there might be a bug either in NCMC---where the nonbonded interactions aren't being correctly switched off for the energy evaluations---or in the geometry engine where either the normalizing constants `Z` aren't correctly computed or an insufficient number of valence terms are being accounted for in drawing the new atom positions.

jchodera:
Looking through `geometry.py`, one issue might be that the normalized torsion probability is actually a probability mass function and not a probability density function on the space of `[0,2*pi)`, so when we compute 
```
P(r_old, theta_old, phi_old | x_core) J(r_old, theta_old, phi_old)
------------------------------------------------------------------
P(r_new, theta_new, phi_new | x_core) J(r_new, theta_new, phi_new)
```
the left-most ratio ends up depending on `n_divisions` if the net number of atoms in the transformation is changing.

I see now why you wanted to focus on validating/testing the geometry engine separately---there are a lot of subtle details in there that we will need to get right!

jchodera:
Thought experiment: If we had a set of linear alkanes (methane, ethane, propane, n-butane) where there were only valence forcefield terms (bond, angle, torsion), what do we expect the sampler to do?  Should it sample each one of these equally, or should it spend most of its time in the longest or shortest chain?

I'm not quite sure if it should naturally equally sample all of these molecules (the dimensionality is different for each one, but the partition functions are factorizable).

These per-atom issues may not actually matter if we use the extra recursion step with SAMS to do optimization---any intrinsic biases would get evened out by adjusting the weights in the unbound states.

jchodera:
Let's consider a simpler RJMC experiment: Sampling X = {x_1, x_2, ..., x_N} where the number of 1D components N is variable from 1 to `N_max`.  If our likelihood function is factorizable:
```
p(X) = \prod_{n=1}^N (2*pi)^{-1/2} \exp[-x_n^2 / 2]
```
then I believe we should see uniform sampling of `N` over `1..N_max` because the likelihood is factorizable.

I think our simple alkane example above *should* be factorizable in this way too as long as there are no rings, where the probability density function of the new atom being added can capture *all* contributions from valence energy components.  In that case, we should have uniform sampling over chemical species if we get the RJMC right.

If we have nonbonded interactions or rings being created/destroyed, I think that's no longer the case and we need to apply some iteratively-determined weights to make sampling uniform over chemical species (if so desired).  But we can probably use this as a validation test (cc: #79).

jchodera:
I think this should be good to merge once tests pass.

I'll look into whether we can move the energy evaluation to the initial and final fully-interacting molecules as we discussed yesterday, but we can do that in a separate PR.

jchodera:
This is ready to merge unless you have incorporated the changes into the other PR already.

jchodera:
Can we merge this?

pgrinaway:
Hm, the `test_sampler.py` changes will cause a conflict now. I guess we can merge, if you're finished here.

jchodera:
Shoot.  How do you want to do this---I can wait until after #83 and then pull in your changes to my PR, or vice-versa....

pgrinaway:
Might as well just merge and then I'll deal with it.

PR #81, opened 2016-02-01
Allow specification of desired mutations in metadata

juliebehr:
Calling the list of mutations ```metadata['mutation_library']``` but this is distinct from ```metadata['sequence_library']``` which will be used to define things like peptides, probably in a separate class.


jchodera:
Whoa!  You beat me to it.

See #82 as well.  This should probably be an optional argument to the `PointMutationEngine` constructor rather than to `metadata` (which is meant to keep track of information about the current sampler state, not tell the engine how it should create proposals).  Other than that, this might be what we need!

Take a look at #82?

jchodera:
I think the idea with `metadata` is that it would keep track of additional sampler state information not encoded by `Topology` or `positions` that will change during the simulation.  For example, @pgrinaway can use it to keep track of the `smiles` string during the simulation, which is hard to figure out from the `Topology` object. You could use it to keep track of the list of current point mutants that deviate from WT.

Arguments that provide information about how the `ProposalEngine` is supposed to make proposals (like `mutation_library` or `sequence_library`) can be optional arguments to the `PointMutationEngine` constructor.

jchodera:
Looks good!  The only issue I spotted is that unless we also want to have to specify the wild-type residue for each resid we want to allow mutations for, we probably want to have the constructor add the current residue name at that resid to the list of `allowed_mutations` at that site. 

For example, currently, if we have `ALA-VAL-LEU` and specify `allowed_mutations = [ ('1', ['GLY']) ]`, it won't let us go back to `ALA`---it will just change it to `GLY` and be done with it.  Not sure if this is a feature or a bug.

Otherwise, good to merge!

juliebehr:
In general for my intended uses, this is always the case (I'll always want to go back to WT before making another change) -- which is why ```_save_mutations()``` is keeping track of ```['ALA-1-GLY']``` and putting it in ```metadata```.  Should this be moved?

jchodera:
I think it might be easier to simply add wild-type residues to the list of `allowed_mutations` by default.

jchodera:
Actually, it will probably be easier to test if we can use this as a feature to test forced mutations with NCMC. If you haven't made changes already, go ahead and merge this as is!

PR #83, opened 2016-02-02
Refactor small molecule ProposalEngines

pgrinaway:
This is the refactor of the small molecule proposal engine stuff that we discussed earlier. I've created a new `SystemGenerator` that uses the new `ForceField` stuff. Will now refactor the `ProposalEngine`s to get rid of tleap and simplify (as well as test). Also on the agenda for this PR is randomly selecting an alignment.

### DIFF
--- a/perses/rjmc/topology_proposal.py
+++ b/perses/rjmc/topology_proposal.py
@@ -812,7 +812,7 @@ class SmallMoleculeSetProposalEngine(ProposalEngine):
         self._generated_topologies = dict()
         super(SmallMoleculeSetProposalEngine, self).__init__(system_generator, proposal_metadata)
 
-    def propose(self, current_system, current_topology, current_positions, beta, current_metadata):
+    def propose(self, current_system, current_topology, current_positions, beta, current_metadata=None):
         """
         Propose the next state, given the current state
 
@@ -1027,7 +1027,7 @@ class SmallMoleculeSetProposalEngine(ProposalEngine):
             The current topology
         positions : [n, 3] np.ndarray of floats (Quantity nm)
             The current positions of the system
-        molecule_smiles : dict
+        molecule_smiles : string
             The current molecule smiles
 
         Returns
--- a/perses/rjmc/topology_proposal.py
+++ b/perses/rjmc/topology_proposal.py
@@ -857,12 +857,33 @@ class SmallMoleculeSetProposalEngine(ProposalEngine):
         Construct a new topology
         Parameters
         ----------
-        oemol_proposed
+        oemol_proposed : oechem.OEMol object
+            the proposed OEMol object
 
         Returns
         -------
-
+        new_topology : app.Topology object
+            A topology with the receptor and the proposed oemol
+        mol_start_index : int
+            The first index of the small molecule
         """
+        mol_topology = forcefield_generators.generateTopologyFromOEMol(oemol_proposed)
+        new_topology = copy.deepcopy(self._receptor_topology)
+        mol_start_index = 0
+        newAtoms = {}
+        for chain in mol_topology.chains():
+            newChain = new_topology.addChain(chain.id)
+            for residue in chain.residues():
+                newResidue = new_topology.addResidue(residue.name, newChain, residue.id)
+                for atom in residue.atoms():
+                    newAtom = new_topology.addAtom(atom.name, atom.element, newResidue, atom.id)
+                    if atom.index == 0:
+                        mol_start_index = newAtom.index
+                    newAtoms[atom] = newAtom
+        for bond in mol_topology.bonds():
+            new_topology.addBond(newAtoms[bond[0]], newAtoms[bond[1]])
+        
+        return new_topology, mol_start_index
 
 
     def _smiles_to_oemol(self):
--- a/perses/rjmc/topology_proposal.py
+++ b/perses/rjmc/topology_proposal.py
@@ -763,7 +763,7 @@ class SystemGenerator(object):
 
     def __init__(self, forcefields_to_use, forcefield_kwargs=None, metadata=None):
         self._forcefields = forcefields_to_use
-        self._forcefield_kwargs = forcefield_kwargs
+        self._forcefield_kwargs = forcefield_kwargs if forcefield_kwargs is not None else {}
         self._forcefield = app.ForceField(*self._forcefields)
         self._forcefield.registerTemplateGenerator(forcefield_generators.gaffTemplateGenerator)
 
--- a/perses/rjmc/topology_proposal.py
+++ b/perses/rjmc/topology_proposal.py
@@ -282,7 +282,7 @@ class ProposalEngine(object):
         Contains information necessary to initialize proposal engine
     """
 
-    def __init__(self, proposal_metadata, system_generator):
+    def __init__(self, system_generator, proposal_metadata):
         self._system_generator = system_generator
 
     def propose(self, current_system, current_topology, current_positions, beta, current_metadata):
@@ -787,14 +787,138 @@ class SystemGenerator(object):
 class SmallMoleculeSetProposalEngine(ProposalEngine):
     """
     This class proposes new small molecules from a prespecified set. It uses
-    exponentiated tanimoto for proposal probabilities.
+    uniform proposal probabilities, but can be extended.
 
+    Parameters
+    ----------
+    list_of_smiles : list of string
+        list of smiles that will be sampled
+    receptor_topology : app.Topology object
+        topology of the receptor
+    system_generator : SystemGenerator object
+        SystemGenerator initialized with the appropriate forcefields
+    proposal_metadata : dict
+        metadata for the proposal engine
     """
 
     def __init__(self, list_of_smiles, receptor_topology, system_generator, proposal_metadata=None):
         self._list_of_smiles = list_of_smiles
-        self._receptor_pdb = receptor_topology
+        self._receptor_topology = receptor_topology
+        super(SmallMoleculeSetProposalEngine, self).__init__(system_generator, proposal_metadata)
 
+    def propose(self, current_system, current_topology, current_positions, beta, current_metadata):
+        """
+        Propose the next state, given the current state
+
+        Parameters
+        ----------
+        current_system : openmm.System object
+            the system of the current state
+        current_topology : app.Topology object
+            the topology of the current state
+        current_positions : [n, 3] np.ndarray of float
+            current positions
+        beta : float
+            inverse temperature
+        current_metadata : dict
+            dict containing current smiles as a key
+
+        Returns
+        -------
+        proposal : SmallMoleculeTopologyProposal object
+           topology proposal object
+        """
+        pass
+
+    def _smiles_to_oemol(self):
+        """
+        Convert the list of smiles into a list of oemol objects. Explicit hydrogens
+        are added, but no geometry is created.
+
+        Returns
+        -------
+        oemols : np.array of type object
+            array of oemols
+        """
+        list_of_smiles = self._smiles_list
+        oemols = np.zeros(self._n_molecules, dtype=object)
+        oemol_smile_dict = dict()
+        for i, smile in enumerate(list_of_smiles):
+            mol = oechem.OEMol()
+            oechem.OESmilesToMol(mol, smile)
+            oechem.OEAddExplicitHydrogens(mol)
+            omega = oeomega.OEOmega()
+            omega.SetMaxConfs(1)
+            omega(mol)
+            oemols[i] = mol
+            oemol_smile_dict[smile] = i
+        return oemols, oemol_smile_dict
+
+    def _get_mol_atom_map(self, current_molecule, proposed_molecule):
+        """
+        Given two molecules, returns the mapping of atoms between them.
+
+        Arguments
+        ---------
+        current_molecule : openeye.oechem.oemol object
+             The current molecule in the sampler
+        proposed_molecule : openeye.oechem.oemol object
+             The proposed new molecule
+
+        Returns
+        -------
+        new_to_old_atom_map : dict
+            Dictionary of {new_idx : old_idx} format.
+        logp_alignment : float
+            logp of the molecule alignment if there is more than one
+        """
+        oegraphmol_current = oechem.OEGraphMol(current_molecule)
+        oegraphmol_proposed = oechem.OEGraphMol(proposed_molecule)
+        mcs = oechem.OEMCSSearch(oechem.OEMCSType_Exhaustive)
+        atomexpr = oechem.OEExprOpts_AtomicNumber
+        bondexpr = 0
+        mcs.Init(oegraphmol_current, atomexpr, bondexpr)
+        mcs.SetMCSFunc(oechem.OEMCSMaxBondsCompleteCycles())
+        unique = True
+        matches = [m for m in mcs.Match(oegraphmol_proposed, unique)]
+        match = np.random.choice(matches)
+        new_to_old_atom_map = {}
+        for matchpair in match.GetAtoms():
+            old_index = matchpair.pattern.GetIdx()
+            new_index = matchpair.target.GetIdx()
+            new_to_old_atom_map[new_index] = old_index
+        return new_to_old_atom_map, 1.0/len(matches)
+
+    def _propose_molecule(self, system, topology, positions, molecule_smiles):
+        """
+        Simple method that randomly chooses a molecule unformly.
+        Symmetric proposal, so logp is 0. Override with a mixin.
+
+        Arguments
+        ---------
+        system : simtk.openmm.System object
+            The current system
+        topology : simtk.openmm.app.Topology object
+            The current topology
+        positions : [n, 3] np.ndarray of floats (Quantity nm)
+            The current positions of the system
+        molecule_smiles : dict
+            The current molecule smiles
+
+        Returns
+        -------
+        proposed_idx : int
+             The index of the proposed oemol
+        mol : oechem.OEMol
+            The next molecule to simulate
+        logp : float
+            The log probability of the choice
+        """
+        current_idx = self._smiles_list.index(molecule_smiles)
+        prob = np.array([1.0/(self._n_molecules-1) for i in range(self._n_molecules)])
+        prob[current_idx] = 0.0
+        proposed_idx = np.random.choice(range(self._n_molecules), p=prob)
+        return proposed_idx, self._oemol_list[proposed_idx], 0.0
 
 class SmallMoleculeProposalEngine(ProposalEngine):
     """
--- a/perses/rjmc/topology_proposal.py
+++ b/perses/rjmc/topology_proposal.py
@@ -802,8 +802,12 @@ class SmallMoleculeSetProposalEngine(ProposalEngine):
     """
 
     def __init__(self, list_of_smiles, receptor_topology, system_generator, proposal_metadata=None):
-        self._list_of_smiles = list_of_smiles
         self._receptor_topology = receptor_topology
+        self._smiles_list = list_of_smiles
+        self._n_molecules = len(list_of_smiles)
+        self._oemol_list, self._oemol_smile_dict = self._smiles_to_oemol()
+        self._generated_systems = dict()
+        self._generated_topologies = dict()
         super(SmallMoleculeSetProposalEngine, self).__init__(system_generator, proposal_metadata)
 
     def propose(self, current_system, current_topology, current_positions, beta, current_metadata):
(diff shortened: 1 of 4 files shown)


jchodera:
Great!  Let me know if you run into trouble.

perses/rjmc/topology_proposal.py
```diff
@@ -742,6 +746,51 @@ def propose(self):
...
+    forcefields_to_use : list of string
+        List of the names of ffxml files that will be used in system creation.
+    metadata : dict, optional
+        Metadata associated with the SystemGenerator.
+    """
+
+    def __init__(self, forcefields_to_use, metadata=None):
+        self._forcefields = forcefields_to_use
+
+    def build_system(self, new_topology, oemol_list, forcefield_kwargs=None):
```
jchodera: `forcefield_kwargs` should probably go in `__init__`, since you want the generator to cache the forcefields and extra arguments that go to `createSystem`.  You can keep an option for additional `kwargs` here, but we want to make sure that the `SystemGenerator` generates systems in exactly the same way each time.

pgrinaway:
Cool. Quick question--should the `SystemGenerator` be called inside the `ProposalEngine` or outside it?

jchodera:
The API for `ProposalEngine` has it return a `TopologyProposal` that includes `new_system`, so we have a few choices:
* refactor things so `TopologyProposal` does not have contain `new_system` (sounds like a bad idea)
* have the base `ProposalEngine` class fill in the `new_system` using a cached `SystemGenerator` (probably the best idea); maybe we would need to pass in a `SystemGenerator` to the constructor for `ProposalEngine` 
* have `TopologyProposal` generate `new_system` automatically if it isn't passed to the constructor; this might be convenient, but I'm not sure where how would pass around the `SystemGenerator` that is to be used for this purpose

pgrinaway:
We could just give a `SystemGenerator` object as an argument to `ProposalEngine`. It wouldn't need to be passed around beyond that, and it would be easy to switch `SystemGenerator` with something else should the need arise. That might be the easiest approach.

My concern here is primarily for host-guest systems. We don't want to run the charge calculations on the host every time we make a proposal, since it takes an extremely long time. Now, it wouldn't be difficult to memoize the template and ffxml* in the `SystemGenerator`, but ideally we'd be able to pre-parameterize the host as I was doing before (to avoid a 2 hour startup time).

Any thoughts on the easiest way to do this?

pgrinaway:
Actually, should I try a quick PR to openmoltools to have the ffxml and template generated by separate functions? At least to make it optional. That would solve the problem--I could generate the ffxml ahead of time and only generate the template at the beginning of the run, which would not be so terrible.

jchodera:
> We could just give a SystemGenerator object as an argument to ProposalEngine. It wouldn't need to be passed around beyond that, and it would be easy to switch SystemGenerator with something else should the need arise. That might be the easiest approach.

Yep, that's what I suggested is the simplest approach.  It should be passed in as part of the `ProposalEngine` constructor.

> My concern here is primarily for host-guest systems. We don't want to run the charge calculations on the host every time we make a proposal, since it takes an extremely long time. Now, it wouldn't be difficult to memoize the template and ffxml* in the SystemGenerator, but ideally we'd be able to pre-parameterize the host as I was doing before (to avoid a 2 hour startup time).

Good point, though I think the problem was with using `sqm` to assign charges rather than `oequacpac`.  

The simpler approach here is to just use the `gaffTemplateGenerator` for now:
```python
    # Add the residue template generator.
    from openmoltools.forcefield_generators import gaffTemplateGenerator
    forcefield.registerTemplateGenerator(gaffTemplateGenerator)
```
Parameters and residue template are automatically memoized then.  We lose out on getting the stereochemistry information, but that's not relevant for now.

jchodera:
Full example is [here](https://github.com/choderalab/openmoltools/blob/master/openmoltools/tests/test_forcefield_generators.py#L220-L247)

jchodera:
> Actually, should I try a quick PR to openmoltools to have the ffxml and template generated by separate functions? At least to make it optional. That would solve the problem--I could generate the ffxml ahead of time and only generate the template at the beginning of the run, which would not be so terrible.

The `ffxml` only contains the parameters and not the residue template.  What you want is a function to generate an `ffxml` containing both parameter and residue template definitions that need to be added to `gaff.xml`.  I'll add a function to do this for you.  (`gaff2xml` used to do this, but that functionality went away, I think.)

pgrinaway:
>The ffxml only contains the parameters and not the residue template. What you want is a function to generate an ffxml containing both parameter and residue template definitions that need to be added to gaff.xml

Actually, what I was suggesting was that I could generate an ffxml ahead of time, then generate and memoize the template at runtime.

pgrinaway:
I think my suggestion is a bit simpler, since it just means creating a new function in openmoltools with the template generation stuff...

jchodera:
Will definitely add that function to openmoltools tomorrow. Can you get by
with using the gaffTemplateGenerator for now?

pgrinaway:
I wouldn't mind doing this myself right now... all I'd like to do is put [these lines](https://github.com/choderalab/openmoltools/blob/master/openmoltools/forcefield_generators.py#L234-L249) into a separate function so that I can call residue template generation separately from parameter generation.

pgrinaway:
Or are those things actually inseparable due to the line about charges?

bas-rustenburg:
Interesting, there is a good chance I'll need that suggested functionality to set up constant-pH ligands without rerunning antechamber with nonphysical atoms. +1 on implementing a separating the template generation in separate from parameter generation.

jchodera:
You will still eventually need to generate parameters on the fly, so for now, I'd change `SystemGenerator` to use `gaffTemplateGenerator`.  That will at least get you up and running, and will still support the use of pregenerated small molecules `ffxml` files when we get those changes made.

The changes in `openmoltools` are nontrivial.  You will need to actually generate the `<ResidueTemplate>` XML block and add that to the `ffxml` file, which might require adopting old `gaff2xml` code and updating it to store the charges in the residue.  We also need to write a number of tests.  I can do this effectively in a day or so, but it shouldn't slow you down now because you can use the `gaffTemplateGenerator`.

jchodera:
> Interesting, there is a good chance I'll need that suggested functionality to set up constant-pH ligands without rerunning antechamber with nonphysical atoms. +1 on implementing a separate template generation from parameter generation.

Can you open an issue in `openmoltools` with more details about what you need to do for that, @bas-rustenburg?

pgrinaway:
Already did! thanks!

pgrinaway:
Ok, as I'm refactoring, another question came to my mind. It seems that the `TopologyProposal` class is really useful in general. I'm wondering if it is what should be passed between all the `Engines`. The base class contains all the necessary attributes for geometry generation, etc., while the child classes contain information that is useful for the `ProposalEngines`. What do others think about making `propose` just take the `TopologyProposal` (or any of its children) as a single argument. This avoids the catch-all `metadata` argument, and might be a bit cleaner. 

Thoughts? Going to tag @juliebehr since this would affect her code (although I don't think it would be a dramatic change, maybe I'm wrong).

juliebehr:
I don't think it would be a dramatic change either; certainly not prohibitive if this makes the most sense.  Seems fine to me.

jchodera:
Can you give an example of what the new API use would look like?  Do you envision that `ProposalEngine.propose()` would take, as input, a half-filled-in `TopologyProposal`, and it would just fill in the rest of the stuff before returning it?  Is it weird to have `TopologyProposal` constructed with only half the data?

pgrinaway:
Hm, you know, it is kind of weird to construct it with half the data. I think I'm actually going to leave this alone right now in the interest of time and raise an issue where we can discuss an API. I think ultimately this would be promoting `TopologyProposal` to some sort of `SamplerState` object. But it definitely requires more thought.

perses/rjmc/topology_proposal.py
```diff
@@ -742,6 +745,181 @@ def propose(self):
...
+        forcefield_kwargs : dict of arguments to createSystem, optional
+            Allows specification of various aspects of system creation.
+
+        Returns
+        -------
+        new_system : openmm.System
+            A system object generated from the topology
+        """
+        from openmoltools import forcefield_generators
+        forcefield = app.ForceField(*self._forcefields)
```
jchodera: You will want to move the creation of `forcefield` and the addition of the GAFF template generator into the constructor if you want `SystemGenerator` to cache new templates that are generated.

perses/rjmc/topology_proposal.py
```diff
@@ -742,6 +745,181 @@ def propose(self):
...
+        forcefield.registerTemplateGenerator(forcefield_generators.gaffTemplateGenerator)
+        system = forcefield.createSystem(new_topology, **forcefield_kwargs)
```
jchodera: You will want to move `forcefield_kwargs` into the constructor so you can cache these when you create a `SystemGenerator`.  Otherwise, you will have to keep passing around the options in every class, which would fuck up all your APIs.

perses/rjmc/topology_proposal.py
pgrinaway: Oh, missed that. Thanks!

perses/rjmc/topology_proposal.py
pgrinaway: True!

pgrinaway:
Hm, looks like it can't find `gaff.xml`... I have `openmm-dev` and `openmoltools-dev` as dependencies, so I'd imagine it should be there...

jchodera:
It's in [`perses/data/gaff.xml`](https://github.com/choderalab/perses/tree/master/perses/data) and [`openmoltools/parameters/gaff.xml`](https://github.com/choderalab/openmoltools/tree/master/openmoltools/parameters) until we get it checked into `openmm`.

I used this scheme [in the perses tests](https://github.com/choderalab/perses/blob/master/perses/tests/test_elimination.py#L145).

We might want to put [this function](https://github.com/choderalab/perses/blob/master/perses/tests/test_elimination.py#L93-L112) somewhere to look up data files.

pgrinaway:
Hm, I wonder what [this error](https://travis-ci.org/choderalab/perses/jobs/106802574#L813)is...

jchodera:
I've seen that before.  It usually means the total charge or the molecule is nonsense.  I saw this before when I was trying to get it to parameterize small molecules from PDB files without `CONECT` records.  It thought imatinib was just a bunch of disconnected atoms.

Did you change the molecules you were feeding into it?  Is there any possibility your molecules are missing important bonds?

pgrinaway:
I'll take a closer look, thanks!

jchodera:
I'll see if I can trap this case and have it spit out a mol2 file of what it thinks you are asking it to parameterize.  That should help us catch issues like this.

pgrinaway:
Ok, well, first let me replace the initial system generation stuff with the new `ForceField` based approach.

jchodera:
This error sounds *exactly* like what I had when I fed in a molecule without any bonds. It was just trying to do a QM calculation on the single carbon:
```
Warning:  SemiEmpirical calculation for carbon was unsuccessful; stopping the charging process for this molecule
```

jchodera:
Dude, where do you create bonds?
https://github.com/choderalab/perses/blob/fix-topology-proposals/perses/rjmc/topology_proposal.py#L390-L393
```python
        # modeller : simtk.openmm.app.Modeller extra atoms from old residue have been deleted, missing atoms in new residue not yet added
        # missing_atoms : dict, key : simtk.openmm.app.topology.Residue, value : list(simtk.openmm.app.topology._TemplateAtomData)
        modeller, missing_atoms = self._delete_excess_atoms(modeller, residue_map)
        # modeller : simtk.openmm.app.Modeller new residue has all correct atoms for desired mutation
        modeller = self._add_new_atoms(modeller, missing_atoms, residue_map)
```

pgrinaway:
I think that is in @juliebehr 's part, not mine.

pgrinaway:
I'm not running that code here (or I shouldn't be)

perses/rjmc/topology_proposal.py
```diff
@@ -742,83 +746,184 @@ def propose(self):
...
+        for chain in mol_topology.chains():
+            newChain = new_topology.addChain(chain.id)
+            for residue in chain.residues():
+                newResidue = new_topology.addResidue(residue.name, newChain, residue.id)
+                for atom in residue.atoms():
+                    newAtom = new_topology.addAtom(atom.name, atom.element, newResidue, atom.id)
+                    if atom.index == 0:
+                        mol_start_index = newAtom.index
+                    newAtoms[atom] = newAtom
+        for bond in mol_topology.bonds():
```
pgrinaway: Adding bonds here

pgrinaway:
When I run this locally, I get `SystemError: <built-in function delete_OEAtomIter> returned a result with an error set` on [this line](https://github.com/choderalab/openmoltools/blob/master/openmoltools/forcefield_generators.py#L46), even though when I run the very same command in the debugger, I get a list of atoms as I'd expect.

pgrinaway:
It seems that this fails on its own:

```python
import openeye.oechem as oechem
import openeye.oeomega as oeomega
from openmoltools import forcefield_generators

def generate_initial_molecule(mol_smiles):
    """
    Generate an oemol with a geometry
    """
    mol = oechem.OEMol()
    oechem.OESmilesToMol(mol, mol_smiles)
    oechem.OEAddExplicitHydrogens(mol)
    omega = oeomega.OEOmega()
    omega.SetMaxConfs(1)
    omega(mol)
    return mol

mol = generate_initial_molecule("CC")
top = forcefield_generators.generateTopologyFromOEMol(mol)
```

yields this:
```pytb
/Users/grinawap/anaconda/lib/python3.5/site-packages/openmoltools-0.7.0.dev0-py3.5.egg/openmoltools/forcefield_generators.py in generateTopologyFromOEMol(molecule)
     44 
     45     # Create atoms in the residue.
---> 46     for atom in molecule.GetAtoms():
     47         name = atom.GetName()
     48         element = Element.getByAtomicNumber(atom.GetAtomicNum())

SystemError: <built-in function delete_OEAtomIter> returned a result with an error set
```

pgrinaway:
Looks like this is the OpenEye failure we've already identified.

jchodera:
Is that the py35 error?  Can you use py27 or py34?

pgrinaway:
Yeah, I'll use py34. Give me a bit to switch my environment on my laptop (I was using the current version of python, which I assumed would be supported, but...). Hopefully this gets fixed soon--it's a bit ridiculous.

pgrinaway:
Ok, up and running again. Looks like this is happening with the very first system generation (before the ProposalEngine even gets a chance to see it). The OEMol has correct bonds, but the topology that comes out of `generateTopologyFromOEMol` is a bit...odd.

Your intuition is correct--the molecule here (ethane) seems to have 7 bonds, all of which are from Atom 7 to Atom 7. Looking into this further...

pgrinaway:
Looks like the problem might be coming from the fact that none of the atoms in the OEMol actually have names. In [openmoltools](https://github.com/choderalab/openmoltools/blob/master/openmoltools/forcefield_generators.py#L52-L54) it's adding bonds from atoms[""] to atoms[""], which is always the same atom.

```python
    # Create bonds.
    atoms = { atom.name : atom for atom in topology.atoms() }
    for bond in molecule.GetBonds():
        topology.addBond(atoms[bond.GetBgn().GetName()], atoms[bond.GetEnd().GetName()])
```

jchodera:
> Ok, up and running again. Looks like this is happening with the very first system generation (before the ProposalEngine even gets a chance to see it). The OEMol has correct bonds, but the topology that comes out of generateTopologyFromOEMol is a bit...odd.

Oh!  Are the atom names not unique?

The code relies onunique atom names
https://github.com/choderalab/openmoltools/blob/master/openmoltools/forcefield_generators.py#L23-L56

pgrinaway:
They are empty, actually.

jchodera:
OK, so I should throw an exception in that case.  

You can just add this to your own code after you generate the OEMol:
```python
    oechem.OETriposAtomTypeNames(molecule)
    oechem.OETriposBondTypeNames(molecule)
```
Let me fix `openmoltools` and add that warning to the docs.

pgrinaway:
Ok, sounds good. Thanks.

pgrinaway:
I think it actually needs to be `oechem.OETriposAtomNames(mol)`

jchodera:
Ooooh, good point.  Sorry!

pgrinaway:
Ok, we're back to the NCMC elimination failing.

pgrinaway:
so maybe it's best if we merge the other PR and get that stuff in here. I still need to implement a couple tests.

jchodera:
Weird. I couldn't reproduce those failures. Can I come see what happens?

pgrinaway:
Sure. I still need to resolve the conflict, though.

pgrinaway:
Ok I got this to work like this:

```python
for x in test_alchemical_elimination_molecule():
    x()
```

pgrinaway:
It seems to run just fine...

pgrinaway:
Oops, spoke too soon. There is a NaN...

pgrinaway:
It's actually ethane with one step.

jchodera:
OK, I just saw `test_elimination` fail on my machine too!  Will debug.

jchodera:
Can you try this change and see if it still fails for you?
```
    platform = openmm.Platform.getPlatformByName('Reference')
    ncmc_engine = NCMCEngine(temperature=temperature, nsteps=ncmc_nsteps, platform=platform)
```

jchodera:
It hasn't failed for me yet on `Reference`

pgrinaway:
Running now...

pgrinaway:
Still hasn't failed on `Reference`

jchodera:
I wonder if it's numerical stability issues in running with single-precision on the `CPU` platform, or an actual bug in `CustomIntegrator` or `CompoundIntegrator`.

pgrinaway:
Yeah, I was wondering the same thing. I'm actually curious if it was using OpenCL on my laptop by default, which might be weird if it's on the integrated GPU.

pgrinaway:
Finished without errors.

pgrinaway:
It definitely dies early when I set the platform to `OpenCL`. I wonder if it's running on the (single-precision) integrated GPU...

jchodera:
I'm not seeing failures on the `CPU` platform either.

jchodera:
OK, no failures on `Reference` or `CPU`.  Only `OpenCL`.

pgrinaway:
Ok, so I've run the NCMC test on the cluster using both `CUDA` and `OpenCL`, in default and double precision modes. The tests fail/NaN on `CUDA` and `OpenCL` even under double precision. I'm running `CPU` on the cluster now just to make sure I get the same results, but it still hasn't failed.

jchodera:
Weeeeird.

I will try to file a bug report.

pgrinaway:
The `CPU` test failed with this:

```
Exception: Delta F (50 steps switching) = 2.229345 +- 0.353267 kT; should be within 6.000000 sigma of 0
```
but I'm not sure if that's just a stochastic test failure.

jchodera:
That's imatinib, right?  I think this is because I am not simulating enough to generate uncorrelated samples.  I'm working on this fix.

pgrinaway:
I think so, yeah.

jchodera:
OK, I'd just continue to use `Reference` or `CPU` for now.  The `OpenCL` platform on my iMac is just not obeying the laws of statistical mechanics.

pgrinaway:
Ok. I'll do that. I need to get one more test in tomorrow morning then this can be merged.

pgrinaway:
Ah, finally merged the branches. Going to add a couple tests then merging.

pgrinaway:
Running the test for the new `ProposalEngine` then merging. I think there is a test failure related to the `GeometryEngine`, but I will open a PR right after this one to fix that + add the final tests there.

jchodera:
Great. I found a bug in CustomIntegrator that was causing our NCMC failures
on cuda and OpenCL. We can switch back once that is fixed in OpenMM.

What else do @juliebehr and I need to do right now?

pgrinaway:
Ok. I think @juliebehr just needs to make a minor modification or two as a result of the new `SystemGenerator` class. I don't think you need to do anything here at the moment--thanks so much for the help, though! (Just noticed your comments in the other PR--sorry!). I'll comment over there in a second.

jchodera:
We can add an NCMC based absolute free energy calculation tool to help you
test, but that won't be useful until you have the
MultipleTargetOptimization sampler class working.

There is another thread with my remaining TODO items; feel free to add to
this if you think of anything else.

pgrinaway:
Probably the dual topology stuff then?

pgrinaway:
Just waiting for this test to pass on travis.. then I'll merge.

pgrinaway:
Ok, the test in question passes. I'm going to merge now and begin resolving the other issue.

Issue #61, opened 2015-12-16
Topology proposals and system construction


pgrinaway:
General question--how should topology proposals be organized, and how should they be matched with `System` construction?

Currently, under the small molecule scheme, there is:

* `Transformation`
   * `SmallMoleculeTransformation`
       * `SingleSmallMolecule`
       * `SmallMoleculeProteinComplex`

So, it's grouped by the task of constructing systems. However, @jchodera points out that it's wise to group by the type of proposal (again, using small molecules as an example):

* `Transformation`
   * `SmallMoleculeTransformation`
      * `LibrarySmallMoleculeTransformation`
      * `SyntheticAccessbilityTransformation`
       ...etc.

Thoughts on this?

jchodera:
I like grouping this by type of proposal:
```
* Transformation
   * SmallMoleculeTransformation
      * SmallMoleculeLibrary
      * CombinatorialLibrary
      * ReactionAwareSynthesis
   * PolymerTransformation
      * PointMutation
      * PeptideLibrary
```
These classes should focus on manipulating the `Topology` object and creating the atom mapping between old and new `Topology` objects

We can separate out the system building machinery into a helper `SystemBuilder` tool that uses the `Topology` to create a modified `System`.  We're nearly done making changes to the OpenMM `ffxml` handler and AMBER forcefields that will allow us to resurrect the old YANK `SystemBuilder` for this purpose.

pgrinaway:
Yeah, the more I think about this, the more I like grouping by type of proposal. Scientifically, it seems to make more sense.

jchodera:
The design should maximize code reuse (or minimize code duplication). There are certainly opportunities for experimentation within these classes to improve efficiency. This could either be done with adding more class options or through subclasses, though class options may be easier.  We should think about how we might want to specify these options at runtime, perhaps through a YAML file as with YANK.

We will also need sampler drivers:
* MCMCSampler - samples with fixed log weights g_k
* SAMSSampler - adjusts weights g_k to produce marginal distribution that matches imposed pi_k

Once we write these components, it should be easy to confine most of our scientific tinkering to various design/optimization classes that run multiple SAMSSamplers (via MPI?) and do their thing by using the online g_k estimated to update the target pi_k that is fed back to the samplers.

jchodera:
Would it help to flesh out the APIs for these things further, or is it pretty straightforward?

pgrinaway:
[...]

PR #87, opened 2016-02-05
[WIP] Begin implementing samplers

pgrinaway:
Starting to implement the sampler objects that we've described in #61 and elsewhere.

### DIFF
--- a/perses/rjmc/mcmcsamplers.py
+++ b/perses/rjmc/mcmcsamplers.py
@@ -1,0 +1,55 @@
+"""
+MCMC and SAMS samplers for expanded ensemble simulations
+"""
+
+import simtk.openmm as openmm
+import simtk.unit as unit
+import simtk.openmm.app as app
+
+import numpy as np
+
+
+class MCMCSampler(object):
+    """
+    MCMC sampler object
+
+    Parameters
+    ----------
+    system : simtk.openmm.System object
+        The system to simulate
+    integrator : simtk.openmm.Integrator object
+        OpenMM integrator to use (could be CustomIntegrator)
+    positions : [n,3] np.ndarray
+        Initial positions of the system
+    """
+
+    def __init__(self, system, integrator, positions, platform_name='CPU'):
+        self._system = system
+        self._integrator = integrator
+        self._initial_positions = positions
+        platform = openmm.Platform.getPlatformByName(platform_name)
+        self._context = openmm.Context(system, integrator, platform)
+
+    def integrate(self, nsteps):
+        """
+        Perform n steps of integration
+
+        Parameters
+        ----------
+        nsteps : int
+            Number of steps to integrate
+        """
+        self._integrator.step(nsteps)
+
+    @property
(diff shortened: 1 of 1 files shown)


jchodera:
The `MCMCSampler` probably needs some way to specify the thermodynamic state to be sampled (temperature, pressure, etc).  

Also, it generally helps to monitor the reduced potential in `kT`, since it's easier than using unit-bearing potential energies (which also don't include pressure-volume stuff).

pgrinaway:
>The MCMCSampler probably needs some way to specify the thermodynamic state to be sampled (temperature, pressure, etc).

Doesn't this get specified in a combination of inputs to the `Integrator` as well as `MonteCarloBarostat` or something like that (those objects are passed in instantiated already)? I agree that it still needs to be there though.


>Also, it generally helps to monitor the reduced potential in kT, since it's easier than using unit-bearing potential energies (which also don't include pressure-volume stuff).

Ok.

pgrinaway:
I guess I could just accept a `ThermodynamicState` object for that...

pgrinaway:
@jchodera what is the easiest way to reliably retrieve the pressure?
it looks like the way to do this is
```python
pressure_param_name = barostat.Pressure() #Anisotropic barostats have three obviously
pressure = context.getParameter(pressure_param_name)
```
Since different barostats have different ways of retrieving the pressure, I'm not sure what the most general solution is. 

On the other hand, in the Yank code, it seems like you just use the set value for the pressure (which makes sense, since obviously an NPT ensemble should have a constant pressure...). Is that more correct? Or should I take the current calculated pressure as above?

jchodera:
It's tricky.  

Are you retrieving it from a `Context` or a `System`?

Setting temperature and pressure is also hard.  See what `MCMCSampler` does here to *set* the pressure and temperature:
https://github.com/choderalab/openmmmcmc/blob/master/openmmmcmc/thermodynamics.py#L115-L145

More generally, are you sure you want to pass a `Context` in to `MCMCSampler`, rather than a `System`?

jchodera:
Sorry, I am an idiot. I totally forgot the current MCMCSampler doesn't have
a way to update the System yet.

Perhaps the best way to proceed is to propose an API that will make things
simple for you and I can implement this into MCMCSampler for you? Or we
could decide that it is easier to implement a basic MCMCSampler in perses
that does what you need.

Whatever the case, what we need is a minimal working example for the whole
stack that does what you need (passes information in the order you want)
without the internals filled then. Only then can we make sure to implement
the minimal API bits.

Defining these classes one at a time and implementing them fully before
going to the next layer of the sampler stack is unlikely to get us there
quickly, as we have seen.

jchodera:
I can take a stab at this if you would prefer to just move on to geometry
engine tests.

pgrinaway:
>I can take a stab at this if you would prefer to just move on to geometry
engine tests.


Ok, thanks. That would probably help things along a bit...

pgrinaway:
I'll close this PR then.

Issue #90, opened 2016-02-08
How big are the chemical state spaces we are supporting?


jchodera:
This is relevant for how we should represent the log weights for expanded ensemble and SAMS samplers.  

* How big are the chemical state spaces we are supporting?  
* Can we assume we will have some way to enumerate all possible chemical state `key`s at the beginning?  If so, how would we retrieve this list of `key`s from the `ProposalEngine`?
* Should we instead be lazy and only keep track of the `key`s for states we have encountered so far?  If so, can all of the recursion strategies described in [the SAMS paper](http://www.stat.rutgers.edu/home/ztan/Publication/SAMS_redo4.pdf) support this strategy, where only the visited state (and maybe others in its neighborhood) are updated each iteration?

pgrinaway:
> How big are the chemical state spaces we are supporting?

I think it's finite, in the case of a bag of molecules or mutations, and finite of unknown cardinality for reaction-aware proposals.

>Can we assume we will have some way to enumerate all possible chemical state keys at the beginning? If so, how would we retrieve this list of keys from the ProposalEngine?

Not in the reaction-aware case (or perhaps we will find a way, but I don't think we should assume that that's true). In the other cases, we could add a property that allows retrieval of such a thing. It is given in the constructor, anyway.

>Should we instead be lazy and only keep track of the keys for states we have encountered so far? If so, can all of the recursion strategies described in the SAMS paper support this strategy, where only the visited state (and maybe others in its neighborhood) are updated each iteration?

I think that being lazy is more scalable, but for smaller sets maybe we'd just want to pre-enumerate the entire thing?

jchodera:
OK, so we will assume 
* Chemical/mutation space is finite but either of known or unknown cardinality
* We won't have a convenient way to pre-enumerate all states
* We wil assume all proposals are "local", in that one or a small number of proposals are considered at a time
* We will use free energy update rules that assume estimates of free energies for unvisited states will remain at zero, and updates for visited states are computed relative to this.  This requires carefully modifying the update rules as we implement them.

Issue #89, opened 2016-02-08
TopologyProposal should contain keys of old and new states


jchodera:
We should have TopologyProposal contain unique keys for old and new states. These keys will need to be matched up across different samplers to do the free energy adaptation. 

Metadata isn't the best place for this since it is integral to the sampling scheme. I envision metadata as being something that ProposalEngine might need or analysis scripts might find useful.

jchodera:
Should I implement this in my PR?  And if so, how should this work?

jchodera:
Actually, could @pgrinaway and @juliebehr implement this, since it will require you add some mechanism to `PointMutationEngine` and `SmallMoleculeSetProposalEngine` to generate a unique key for each proposed state?  You don't have to call it `key`---something like `state` or `chemical_state_key` or something would work.  It just needs to be a hashable object.

pgrinaway:
+1. I can do that. I think that (canonical isomeric) SMILES allows [tautomer specification] (http://www.daylight.com/meetings/summerschool98/course/dave/smiles-convent.html), and a quick check with openeye seems to demonstrate that the openeye tools are happy with it. Protonation state changes can also be reflected in SMILES, right? So I suggest:

1) canonical isomeric SMILES string for chemical identity
2) sequence for protein identity

Is that ok?

jchodera:
Sounds great.

At some point, we might think of allowing a `set` or `tuple` of `ProposalEngine`s to be specified to propose multiple chemical changes in a Gibbs sampling framework, but we can just use a `set` or `tuple` of those keys at that point.

What do you want to call it?  `key`?  `chemical_state`?  `sampler_discrete_state`?  `bob`?

pgrinaway:
`chemical_state_key`?

jchodera:
Sure. Can always change it later.

pgrinaway:
Added `chemical_state_key` in #97

PR #92, opened 2016-02-08
Sampler implementation in process

jchodera:
Starting basic samplers implementation.

Very much in progress and untested.  Just figuring out APIs right now.


jchodera:
I've temporarily ingested `ThermodynamicState` and `SamplerState` and written a minimal functioning placeholder for `MCMCSampler` until we can switch over to the `openmmmcmc` package.

jchodera:
Some decisions on issues #90 and #89 are needed before I can go further.

jchodera:
This is mostly done.  Once some of the other issues I raised are cleaned up, I can finish the testing and we're off to the races with a very basic functional implementation.

This currently only replicates the NCMC-geometry-NCMC scheme we have in `test_sampler.py`, but we can add more schemes (`geometry-ncmc-geometry` and `geometry-ncmc`) later.

jchodera:
Added tests for the whole sampler stack.  Hopefully this will work once #103 is resolved!

jchodera:
Did a bit more cleanup and documentation.  Just waiting on @juliebehr's changes to finish.

jchodera:
Also, the API for this shouldn't be considered final.  I'm more than happy to change this however we want to make things easier.  This is just a first stab at the whole stack.

pgrinaway:
>This is just a first stab at the whole stack.

That's fine. I think the first stab is important because we want to get the paper out as well as get things started for collaborators. Taking a look now while the debugger takes its time...

jchodera:
Probably easier to link to the whole new files:
* [`perses/samplers/samplers.py`](https://github.com/choderalab/perses/blob/samplers/perses/samplers/samplers.py)
* [`perses/tests/testsystems.py`](https://github.com/choderalab/perses/blob/samplers/perses/tests/testsystems.py)
* [`perses/tests/test_samplers.py`](https://github.com/choderalab/perses/blob/samplers/perses/tests/test_samplers.py)

jchodera:
OK, I'm now stuck waiting on #93 too.

pgrinaway:
What do you need from #93? I can just do that now and merge. Is there any reason we can't merge this first?

jchodera:
See test failures for master. There were some thing being cleaned up in
geometry or TopologyProposal, or so I thought?

jchodera:
Or maybe these are things I forgot to change after merging in changes from
master?

pgrinaway:
Looks like there are 3 failures in master:

1) `GeometryEngine` failure because of a change in the `TopologyProposal` API. This is fixed in my PR and shouldn't really affect this PR.

2) `test_run_example` fails but this is basically deprecated by this PR. (also, the failure is the same change as in 1)...perhaps the changes weren't pulled in?

3) The last one looks like an oversight on my part. I need to change the test for the `SmallMoleculeSetProposalEngine`, but again I don't think it should affect this PR.

jchodera:
Should we merge this then?

pgrinaway:
I think that would be best.

jchodera:
Go for it then! Will have time to clean up anything I missed tonight.

pgrinaway:
Ok, I'll do that now. Hopefully I can finish up #93 tonight as well (if there is an API change I can fix that in #93)

PR #93, opened 2016-02-08
Geometry Engine tests and cleanup

pgrinaway:
This PR will 

* Complete the `GeometryEngine` tests
* Fix any test failures
* Return to sampling torsions from a pdf

Right now, the openeye problem seems to have reared its head again.

### DIFF
--- a/perses/rjmc/geometry.py
+++ b/perses/rjmc/geometry.py
@@ -72,7 +72,7 @@ class FFGeometryEngine(GeometryEngine):
     def __init__(self, metadata):
         self._metadata = metadata
 
-    def propose(self, top_proposal):
+    def propose(self, top_proposal, beta):
         """
         Make a geometry proposal for the appropriate atoms.
 
@@ -80,6 +80,8 @@ class FFGeometryEngine(GeometryEngine):
         ----------
         top_proposal : TopologyProposal object
             Object containing the relevant results of a topology proposal
+        beta : float, in kT
+            The inverse temperature
 
         Returns
         -------
@@ -88,7 +90,6 @@ class FFGeometryEngine(GeometryEngine):
         logp_proposal : float
             The log probability of the forward-only proposal
         """
-        beta = top_proposal.beta
         logp_proposal = 0.0
         structure = parmed.openmm.load_topology(top_proposal.new_topology, top_proposal.new_system)
         new_atoms = [structure.atoms[idx] for idx in top_proposal.unique_new_atoms]
@@ -151,7 +152,7 @@ class FFGeometryEngine(GeometryEngine):
         return new_positions, logp_proposal
 
 
-    def logp_reverse(self, top_proposal, new_coordinates, old_coordinates):
+    def logp_reverse(self, top_proposal, new_coordinates, old_coordinates, beta):
         """
         Calculate the logp for the given geometry proposal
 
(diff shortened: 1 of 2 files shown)


pgrinaway:
Ok, going to go through and see if I can fix the openeye thing:

* Confirmed I am running Python 3.4

* Confirmed the `OEMol` has no bonds when it is given to `OEAssignPartialCharges`

* `generateTopologyFromOEMol` seems to produce a topology with all bonds from one atom to itself. Will move discussion to openmoltools.

pgrinaway:
Actually, I think I see the problem.

pgrinaway:
I neglected to 
```python
    oechem.OETriposAtomNames(mol)
    oechem.OETriposBondTypeNames(mol)
```

pgrinaway:
(debugging unit madness at the moment)

pgrinaway:
Found an error... will fix now.

perses/rjmc/geometry.py
```diff
@@ -163,30 +164,31 @@ def logp_reverse(self, top_proposal, new_coordinates, old_coordinates):
...
-            reverse_proposal_coordinates[atom.idx] = new_coordinates[new_index]
+        old_structure = parmed.openmm.load_topology(top_proposal.old_topology, top_proposal.old_system)
+        old_atoms_with_positions = [old_structure.atoms[atom_idx] for atom_idx in top_proposal.old_to_new_atom_map.keys()]
+        old_unique_atoms = [old_structure.atoms[idx] for idx in top_proposal.unique_old_atoms]
+
+        #copy common atomic positions
+        for atom in old_structure.atoms:
+            if atom.idx in old_atoms_with_positions:
+                corresponding_new_index = top_proposal.old_to_new_atom_map[atom.idx]
+                old_coordinates[atom.idx] = new_coordinates[corresponding_new_index]
```
jchodera: This looks fishy to me.  What are you trying to do here---shouldn't you already have the old coordinates?

The goal of computing the reverse proposal probability is just to compute the probability of placing the old atoms that are deleted back where they were originally.

jchodera: Ah---I think I understand.  You're trying to reuse as much code as possible to go through the same sequence of placing the old atoms one at a time?

pgrinaway:
definitely some bond length weirdness in the reverse logp..

perses/rjmc/geometry.py
pgrinaway: Yeah, I was trying to reduce code duplication. I realize I have a bit of a ways to go on that front, but I was hoping it'd help solve some of the weird problems.

perses/rjmc/geometry.py
```diff
@@ -198,31 +200,34 @@ def logp_reverse(self, top_proposal, new_coordinates, old_coordinates):
...
                 angle_k = angle.type.k
                 sigma_theta = units.sqrt(1/(beta*angle_k))
                 logZ_theta = np.log((np.sqrt(2*np.pi)*sigma_theta/sigma_theta.unit)) #need to eliminate unit to allow numpy log
-                logp_theta = self._angle_logq(internal_coordinates[1], angle, beta) - logZ_theta
+                logp_theta = self._angle_logq(theta, angle, beta) - logZ_theta
 
                 #calculate torsion probability
-                logp_phi = self._torsion_logp(atom, atom_coords, torsion, atoms_with_positions, reverse_proposal_coordinates, beta)
+                logp_phi = self._torsion_logp(atom, atom_coords, torsion, old_atoms_with_positions, old_coordinates, beta)
                 logp = logp + logp_choice + logp_r + logp_theta + logp_phi + np.log(detJ)
```
jchodera: Do you know which of these terms is responsible for the huge magnitude of the reverse `logp`?

pgrinaway: yeah, it was the bond. The length was cut by a factor of 10 as a result of a unit error.

pgrinaway:
Ok, now the `logp_bond` is only ~2, but `logp_phi` is ~-75 for the reverse when it's more like -4 for the forward proposal. Investigating...

pgrinaway:
I think it might be another unit error. One sec.

pgrinaway:
I'm going to merge this and then make sure the `GeometryEngine` API is consistent with the rest of the package (fixing the build failures) then add the last tests.

PR #97, opened 2016-02-09
Removed SmallMoleculeTopologyProposal

### LINKED ISSUE #94: Eliminate SmallMoleculeTopologyProposal
The only difference is that `SmallMoleculeTopologyProposal` adds a `molecule_smiles` property but I don't think this is ever used.

My understanding was that there should be an `old_metadata` and `new_metadata` that would store whatever additional information might be needed for convenience in a `dict`, such as the `smiles` strings for old and new molecules.

pgrinaway:
As the title says, and to resolve #94


pgrinaway:
Should I add the `chemical_state_key` in this PR?

pgrinaway:
Also added `chemical_state_key` to the `TopologyProposal` and the `SmallMoleculeSetProposalEngine`

pgrinaway:
Going to merge so that I can continue to make changes to the `GeometryEngine` as well as other minor changes to accommodate this PR.

jchodera:
We need to add both the *old* `chemical_state_key` and the *new* one.

Issue #99, opened 2016-02-10
Remove PolymerTopologyProposal


jchodera:
I think we can remove `PolymerTopologyProposal`, just like we removed `SmallMoleculeTopologyProposal`.

pgrinaway:
Resolved in #105

PR #105, opened 2016-02-10
Fix proposal engine (minor things)

pgrinaway:
I was going to do this in the geometry PR but my git-fu is not as strong as I would like. This should be a quick PR to shore up some issues identified in the course of #92 .

### DIFF
--- a/perses/rjmc/topology_proposal.py
+++ b/perses/rjmc/topology_proposal.py
@@ -158,7 +158,7 @@ class ProposalEngine(object):
         Contains information necessary to initialize proposal engine
     """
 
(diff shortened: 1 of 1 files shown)


pgrinaway:
Deleted `PolymerTopologyProposal` as suggested in #99

pgrinaway:
Added the `new_chemical_state_key` and `old_chemical_state_key` to the `TopologyProposal` and adjusted `SmallMoleculeSetProposalEngine` accordingly.

pgrinaway:
Merging this so that other fixes to the `ProposalEngine`s can be PR'd.

Issue #103, opened 2016-02-10
Store `chain_id` of chain to mutate in PointMutationEngine object during construction


jchodera:
Initialization of `PointMutationEngine` should take the `chain_id` to mutate and store that, and any other information it needs.

`chain_id` currently needs to be passed in via `PointMutationEngine.propose()`, which doesn't seem reasonable since we would need to pass that in every time.  Instead, we should just take that at object construction time and cache it.

jchodera:
Relevant code:
https://github.com/choderalab/perses/blob/master/perses/rjmc/topology_proposal.py#L268-L290

pgrinaway:
Yeah, I think we want to make the `propose` methods as similar as possible--that way the classes can be easily interchanged.

juliebehr:
Resolved https://github.com/choderalab/perses/pull/108

Issue #102, opened 2016-02-10
Does TopologyProposal use positions?


jchodera:
Does `TopologyProposal` actually make use of `old_positions` or `new_positions`?  I don't think we actually use these anywhere since we split out the geometry engine.

Can we eliminate these?

jchodera:
We would still need positions passed to `GeometryEngine.propose()`, but it could be a separate argument:
```python
new_positions, geometry_logp  = self.geometry_engine.propose(topology_proposal, old_positions)
```

pgrinaway:
Again, this was for the possibility of position-dependent proposals.

jchodera:
We can leave it for now, but it leads to really weird constructs that require us to replace the `positions` in `TopologyProposal`:
https://github.com/choderalab/perses/blob/master/perses/tests/test_sampler.py#L151-L166

pgrinaway:
Oh that is weird, you're right. I hadn't thought about that.

jchodera:
We could change the NCMC API, but I'm not sure which is better.

pgrinaway:
Better to change the `ProposalEngine` I think.

pgrinaway:
Addressed in #106

Issue #101, opened 2016-02-10
Make `metadata` optional?


jchodera:
There are lots of places where we have a mandatory `metadata` argument (such as everywhere in the geometry engine) even though it isn't used yet.  Can we make this an optional argument that defaults to `dict()` or `None`?

pgrinaway:
+1. I think defaulting to `None` is a good option here.

pgrinaway:
I've done that for the small molecule engine in #105 , but I think `metadata` is actually used in the polymer proposal engines.

jchodera:
>  I think metadata is actually used in the polymer proposal engines.

@juliebehr can we get that fixed?  In #103 it seems that `chain_id` should really be in the constructor, and not `metadata`.

juliebehr:
I'm storing the information about what mutations have been made in ```metadata```, where should this go instead?  Also, the constructor has its own ```proposal_metadata``` (which has been there the whole time, I was just following the skeleton) which I'm currently only using to pass ```ffxml```s; should I make them their own argument and remove ```proposal_metadata``` too?

juliebehr:
I just saw that the ```ffxml```s are addressed in https://github.com/choderalab/perses/issues/95 and yes I can easily change them to be retrieved from SystemGenerator -- so then the point remains, just to clarify we're eliminating both ```metadata``` and ```proposal_metadata```?

jchodera:
Can we have chain_id in proposal_metadata in the constructor, or as a separate optional argument? This just gets it out of needing to be passed to propose() every time.

Totally OK to keep the current mutation info in the metadata associated with the current state. We also need to provide a way to get the current chemical state key.

juliebehr:
Resolved https://github.com/choderalab/perses/pull/108

Issue #100, opened 2016-02-10
PointMutationEngine should set chemical_state_key in TopologyProposal


jchodera:
#97 has `SmallMoleculeProposalEngine` set `chemical_state_key`, but it doesn't look like the `PolymerProposalEngine`s (like `PointMutationEngine`) does this.

@juliebehr @pgrinaway can you fix?

We should make `chemical_state_key` mandatory in `TopologyProposal`, since none of the sampler swill work without it.

pgrinaway:
That's a good point. I have an open PR so I'll fix the issues related to the chemical_state_key in that tomorrow.

jchodera:
For the purposes of initializing the samplers, I also need to figure out how to set the *initial* chemical state key the sampler is started in.  

This means we either need:
* a well-documented description of how the state key is defined for each kind of `ProposalEngine` so that we can make one by hand when initializing a sampler
* a method in `ProposalEngine` that computes the key corresponding to a given `Topology`, like
```
proposal_engine = PointMutationEngine(system_generators[environment], max_point_mutants=1, proposal_metadata=proposal_metadata)
state_key = proposal_engine.computeStateKey(topology)
# now I can initialize my sampler to start in `state_key`...
```

pgrinaway:
>a method in ProposalEngine that computes the key corresponding to a given Topology, like

I like this one. I already have code that computes the current state key (per a discussion earlier), so it would be easy to make it into its own method.

juliebehr:
Resolved https://github.com/choderalab/perses/pull/108

PR #108, opened 2016-02-10
Creating chemical_state_key for PolymerMutationEngines (and other minor fixes)

juliebehr:
Changes:
* turned ```proposal_metadata``` and ```current_metadata``` into optional arguments
* removed ```current_positions``` from ```PointMutationEngine.propose()``` -- generating ```np.zeros``` for ```Modeller``` input instead, can later eliminate ```Modeller``` altogether but not currently a priority
* ```chain_id``` is input to ```PointMutationEngine``` constructor
* ```PointMutationEngine._ff``` obtained from ```SystemGenerator```
* added method ```compute_state_key(self, topology)```
* returning ```old_chemical_state_key``` and ```new_chemical_state_key```

### DIFF
--- a/perses/annihilation/ncmc_switching.py
+++ b/perses/annihilation/ncmc_switching.py
@@ -27,7 +27,7 @@ class NCMCEngine(object):
     >>> testsystem = testsystems.AlanineDipeptideVacuum()
     >>> from perses.rjmc.topology_proposal import TopologyProposal
     >>> new_to_old_atom_map = { index : index for index in range(testsystem.system.getNumParticles()) if (index > 3) } # all atoms but N-methyl
-    >>> topology_proposal = TopologyProposal(old_system=testsystem.system, old_topology=testsystem.topology, old_positions=testsystem.positions, new_system=testsystem.system, new_topology=testsystem.topology, logp_proposal=0.0, new_to_old_atom_map=new_to_old_atom_map, metadata=dict())
+    >>> topology_proposal = TopologyProposal(old_system=testsystem.system, old_topology=testsystem.topology, old_chemical_state_key='AA', new_chemical_state_key='AA', new_system=testsystem.system, new_topology=testsystem.topology, logp_proposal=0.0, new_to_old_atom_map=new_to_old_atom_map, metadata=dict())
     >>> ncmc_engine = NCMCEngine(temperature=300.0*unit.kelvin, functions=default_functions, nsteps=50, timestep=1.0*unit.femtoseconds)
     >>> positions = testsystem.positions
     >>> [positions, logP_delete, potential_delete] = ncmc_engine.integrate(topology_proposal, positions, direction='delete')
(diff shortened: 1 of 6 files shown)


juliebehr:
Addresses: 
* https://github.com/choderalab/perses/issues/103
* https://github.com/choderalab/perses/issues/101
* https://github.com/choderalab/perses/issues/100
* https://github.com/choderalab/perses/issues/95
* https://github.com/choderalab/perses/issues/102
* https://github.com/choderalab/perses/issues/98
* https://github.com/choderalab/perses/issues/89

jchodera:
Woohoo!  Thanks!

jchodera:
Looks like you may need to first merge in changes from `master` and fix the failing tests?

juliebehr:
```This branch is 3 commits ahead of master.```

There's something in ```master``` causing the tests to fail I think?

juliebehr:
Ah, but there are some non-topology tests failing due to changes in inputs -- fixing now

juliebehr:
Errors due to new inputs to ```TopologyProposal``` and ```propose()``` fixed in
* ```test_elimination.py```
* ```test_sampler.py```
* ```test_topology_proposal.py``` (```test_small_molecule_proposals()```)
* ```ncmc_switching.py``` doctest

Also, ```AlanineDipeptideVacuum``` test system gives a residue named ```ACE``` which then cannot be matched by ```PointMutationEngine.compute_state_key()``` -- for now catching unknown residues and translating them as ```'X'```

jchodera:
Ah, OK!  Those residues are indeed defined in the `ForceField`, but I guess the residue names are hardcoded for now.

jchodera:
I think the desired behavior here would be to not try to mutate residues that don't fall in the list of `aminos`.

juliebehr:
Does ForceField have a way to translate to 1 letter codes? The issue is that Topology does not, hence the hardcoded dict.

juliebehr:
There is currently no possibility of mutating _to_ residues not in ```aminos``` (obviously), but no restrictions on the input residues (and even if an unknown residue isn't allowed to mutate it would still be included in the sequence).  Do we never want to allow sequences with nonstandard residues?

juliebehr:
Never mind, ACE obviously doesn't have a 1-letter code.. I think the issue is not whether these residues could be mutated but how to define them in the ```chemical_state_key``` -- can they be omitted since they will never change?

jchodera:
> Never mind, ACE obviously doesn't have a 1-letter code.. I think the issue is not whether these residues could be mutated but how to define them in the chemical_state_key -- can they be omitted since they will never change?

You can either omit them because they never change, or you can just use the sequence of residue names that `Topology` has for the chain you are mutating.

jchodera:
>  Do we never want to allow sequences with nonstandard residues?

Yeah, I think we will need to allow sequences with nonstandard residues.

juliebehr:
The remaining build failures are addressed by https://github.com/choderalab/perses/pull/93

jchodera:
Awesome, thanks!  I'm going to merge this so I can test and merge my samplers PR.

Issue #50, opened 2015-12-01
Bond constraints


jchodera:
What are we doing about constrained bonds at the moment?

We obviously have to make sure not to try to convert constrained bonds to unconstrained bonds.

If we only deal with hydrogen constraints, does our current MCSS atom matching algorithm ensure we will *never* map a bond that changes constraint status in old/new topologies as atoms that should be retained?

Also, we will screw up if we draw a bond length from a harmonic X-H bond length distribution and then apply constraints that change the bond length.  We need to make sure to *fix* those bond lengths to the constrained bond length in the geometry engine.

pgrinaway:
This is the problem that @jchodera is observing with the lack of bond. Parmed does not perceive constraints, and so there is just no bond. I'm going to add the constraint stuff in the next PR.

pgrinaway:
Discussed in [here](https://github.com/choderalab/perses/pull/51).

pgrinaway:
Constraint tolerance is a context parameter in OpenMM, no? Is there a good way to deal with this in `GeometryEngine`?

jchodera:
Constraint tolerance is an integrator parameter!
http://docs.openmm.org/6.3.0/api-python/classsimtk_1_1openmm_1_1openmm_1_1Integrator.html#a311f37a6d317055063bd8b390ba47737

pgrinaway:
Oops. Thanks! Still not accessible to `GeometryEngine` as written, though. Should `GeometryEngine` accept constraint tolerance as a parameter? Should it be in the constructor?

jchodera:
> Oops. Thanks! Still not accessible to GeometryEngine as written, though. Should GeometryEngine accept constraint tolerance as a parameter? Should it be in the constructor?

Are we talking about constraint tolerance or what kind of bond constraints to use?

We definitely want to specify what kind of bond constraints to use, but that is permitted already.  See [this example](https://github.com/choderalab/perses/blob/master/perses/tests/testsystems.py#L102-L107).  Note that these are all changed to `constraints=None` in my PR #109

pgrinaway:
Constraint tolerance. `GeometryEngine` can figure out whether all bonds or H bonds are constrained (but not whether there are angle constraints) in the current code. I assume that `bond_length-constraint` will not always be exactly 0, and I assumed that the `constraint_tolerance` would be a good tolerance.

```python
if np.abs(bond_length-constraint)<constraint_tolerance:
    logp_bond = 0.0
else:
[...]

PR #113, opened 2016-02-12
Adding constraints, fixing a couple tests.

pgrinaway:
Fixed a test API and added support for bond constraints. As noted in #50 , the only remaining issue is to determine the `constraint_tolerance` for the reverse logp.

### DIFF
--- a/perses/rjmc/geometry.py
+++ b/perses/rjmc/geometry.py
@@ -245,23 +245,6 @@ class FFGeometryEngine(GeometryEngine):
                 old_unique_atoms.remove(atom)
         return logp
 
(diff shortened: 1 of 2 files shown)


pgrinaway:
Ok, assuming that the constraints have been iterated to convergence, if there is a constraint, we will assume it is satisfied and set the corresponding logp to 0.0.

pgrinaway:
Just waiting for the travis build. It's otherwise finished.

jchodera:
Sweet!

jchodera:
I think the remaining failures are mine, so I'll merge this and fix.

Issue #114, opened 2016-02-14
No eligible torsions found during geometry engine proposal


jchodera:
See #109:
```

======================================================================
ERROR: Testing MultiTargetDesign sampler with alanine dipeptide mutation transfer free energy from vacuum -> implicit
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/choderaj/github/choderalab/perses/perses.choderalab/perses/samplers/samplers.py", line 934, in run
    self.update()
  File "/Users/choderaj/github/choderalab/perses/perses.choderalab/perses/samplers/samplers.py", line 919, in update
    self.update_samplers()
  File "/Users/choderaj/github/choderalab/perses/perses.choderalab/perses/samplers/samplers.py", line 889, in update_samplers
    sampler.update()
  File "/Users/choderaj/github/choderalab/perses/perses.choderalab/perses/samplers/samplers.py", line 811, in update
    self.update_sampler()
  File "/Users/choderaj/github/choderalab/perses/perses.choderalab/perses/samplers/samplers.py", line 785, in update_sampler
    self.sampler.update()
  File "/Users/choderaj/github/choderalab/perses/perses.choderalab/perses/samplers/samplers.py", line 673, in update
    self.update_sampler()
  File "/Users/choderaj/github/choderalab/perses/perses.choderalab/perses/samplers/samplers.py", line 635, in update_sampler
    geometry_new_positions, geometry_logp  = self.geometry_engine.propose(topology_proposal, ncmc_old_positions, self.sampler.thermodynamic_state.beta)
  File "/Users/choderaj/github/choderalab/perses/perses.choderalab/perses/rjmc/geometry.py", line 112, in propose
    torsion, logp_choice = self._choose_torsion(atoms_with_positions, atom)
  File "/Users/choderaj/github/choderalab/perses/perses.choderalab/perses/rjmc/geometry.py", line 546, in _choose_torsion
    torsion_idx = np.random.randint(0, len(eligible_torsions))
  File "mtrand.pyx", line 945, in mtrand.RandomState.randint (numpy/random/mtrand/mtrand.c:10732)
ValueError: low >= high
```
Looks like this might be triggered when the new molecule does not have any eligible torsions.  Will try to track down exactly what kinds of transformations cause this to happen.

jchodera:
[...]

PR #121, opened 2016-02-15
Canonicalize input smiles

pgrinaway:
Makes sure that the input smiles strings are canonical isomeric smiles in the openeye flavor by taking the input library and converting each to an `OEMol` then to canonical isomeric SMILES.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ script:
   - conda remove --yes parmed parmed-dev && conda install --yes parmed-dev
   - conda remove --yes openmm openmm-dev && conda install --yes openmm-dev
   # Run tests
-  - cd devtools && nosetests --processes=-1 --process-timeout=1800 $PACKAGENAME --nocapture --verbosity=3 --with-doctest --with-timer && cd ..
+  - cd devtools && nosetests $PACKAGENAME --nocapture --verbosity=3 --with-doctest --with-timer && cd ..
 env:
   matrix:
     - python=2.7  CONDA_PY=27 OPENEYE_CHANNEL="-i https://pypi.anaconda.org/openeye/channel/main/simple"
--- a/perses/rjmc/topology_proposal.py
+++ b/perses/rjmc/topology_proposal.py
@@ -885,7 +885,7 @@ class SmallMoleculeSetProposalEngine(ProposalEngine):
     """
 
     def __init__(self, list_of_smiles, system_generator, residue_name='MOL', proposal_metadata=None):
-        self._smiles_list = list_of_smiles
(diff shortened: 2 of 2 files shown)


jchodera:
Canonicalicious.

jchodera:
We might also want to do some other cleanup here, like then taking the unique set of SMILES, e.g.
```python
molecules = list(set(molecules))
```

pgrinaway:
oh ok i can add that in now.

pgrinaway:
At least one of those builds seemed to have passed everything to a point and [then stalled](https://travis-ci.org/choderalab/perses/builds/109448524#L886).

pgrinaway:
Also timed out...hm...

jchodera:
I'm confused by that too.  It's possible my [parallelization of nosetests](https://github.com/choderalab/perses/blob/master/.travis.yml#L39) is to blame.  Want to disable that?

pgrinaway:
Looks like you're only using 1 process though?

pgrinaway:
Or were you suggesting to increase that number?

pgrinaway:
Oh, nevermind, that says -1. I missed the negative sign.

jchodera:
This is looking much better.

I bet we can parallelize the tests with a travis build matrix, testing each component in a separate branch of the matrix.  I can try that is a separate PR.

pgrinaway:
Yeah, I think we've passed the earlier issue. It seems like some of the tests were just not writing output in the parallel scheme, causing travis to end the build?

pgrinaway:
Ok, I'm going to try to figure out what is going wrong here.

jchodera:
I imagine there are simply issues with the SMILES strings I downloaded from the Shoichet lab website.  The files are here: https://github.com/choderalab/perses/tree/master/perses/data

jchodera:
If you run `nosetests -vv perses/tests/test_samplers.py` locally, you can trigger the failing tests right away.

pgrinaway:
Ok, well, this might be an opportune moment to fix the Python 3 failures as well.

pgrinaway:
Looks like it's running into the issue mentioned in #114 . Also, a minor issue regarding reading the SMILES csvs: I needed to read with `r` instead of `rb`.

jchodera:
Shoot.  This is still timing out:
```
Testing expanded ensemble sampler with AlkanesTestSystem 'explicit' ... 
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
```
It must be that there simply isn't enough output printed when the samplers are not failing.

Let's go ahead and merge and I can add some sampler output and reduce the number of steps per iteration.

pgrinaway:
Ok. Merging this. I will look more into the molecule construction failures.

PR #122, opened 2016-02-15
Speed up samplers tests

jchodera:


jchodera:
On my machine, this seems to hang in the geometry proposal part:
```
Testing expanded ensemble sampler with T4LysozymeInhibitorsTestSystem 'explicit' ... --------------------------------------------------------------------------------
Expanded Ensemble sampler iteration        0
MCMC sampler iteration 5
Taking 5 steps of GHMC...
Updating chemical state with ncmc-geometry-ncmc scheme...
Proposing new topology...
Generating new system...
Performing NCMC annihilation
Geometry engine proposal...
```
It doesn't progress from there after many minutes.

Is there anything in the geometry proposal that could be going into an infinite loop?

pgrinaway:
Do you know which molecule this was? On my computer, this test seems to pass fine. (or fail when there are no eligible torsions). I could potentially imagine that it gets stuck in an infinite loop if it's trying to add positions atoms that it can't, but never raises an exception. In this case I'd need to raise an exception. If you know which molecule pair causes the problem, that'd be great. Otherwise, I'll look for it myself tomorrow.

jchodera:
Here's what I'm seeing:
```
Testing expanded ensemble sampler with T4LysozymeInhibitorsTestSystem 'explicit' ... --------------------------------------------------------------------------------
Expanded Ensemble sampler iteration        0
MCMC sampler iteration 5
Taking 5 steps of GHMC...
Updating chemical state with ncmc-geometry-ncmc scheme...
Proposing new topology...
Proposed transformation from CC(=O)CSc1cccs1 to c1ccc2c(c1)cco2
Generating new system...
Performing NCMC annihilation
Geometry engine proposal...
```

pgrinaway:
Ok, looking into it...

pgrinaway:
Possibly a problem with the atom matching? In the `GeometryEngine` tests, I have some code to generate a test atom mapping. This code fails to obtain any match on that molecule pair. Do you see anything particularly wrong with it?
```python

def _get_mol_atom_map(current_molecule, proposed_molecule):
        """
        Given two molecules, returns the mapping of atoms between them.

        Arguments
        ---------
        current_molecule : openeye.oechem.oemol object
             The current molecule in the sampler
        proposed_molecule : openeye.oechem.oemol object
             The proposed new molecule

        Returns
        -------
        new_to_old_atom_map : dict
            Dictionary of {new_idx : old_idx} format.
        logp_alignment : float
            logp of the molecule alignment if there is more than one
        """
        oegraphmol_current = oechem.OEGraphMol(current_molecule)
        oegraphmol_proposed = oechem.OEGraphMol(proposed_molecule)
        mcs = oechem.OEMCSSearch(oechem.OEMCSType_Exhaustive)
        atomexpr = oechem.OEExprOpts_AtomicNumber
        bondexpr = 0
        mcs.Init(oegraphmol_current, atomexpr, bondexpr)
        mcs.SetMCSFunc(oechem.OEMCSMaxBondsCompleteCycles())
        unique = True
        matches = [m for m in mcs.Match(oegraphmol_proposed, unique)]
        match = np.random.choice(matches)
        new_to_old_atom_map = {}
        for matchpair in match.GetAtoms():
            old_index = matchpair.pattern.GetIdx()
            new_index = matchpair.target.GetIdx()
            new_to_old_atom_map[new_index] = old_index
        return new_to_old_atom_map, 1.0/len(matches)
```

jchodera:
Can you generate images using oedepict to help debug?

pgrinaway:
Oh, I found the problem. Stupid mistake on my part in the test. The `GeometryEngine` does not end up in an infinite loop for me here...

pgrinaway:
I should add that that mistake was just added to my local repo...it shouldn't be in your PR.

jchodera:
Weird.  So there is something else going on?

jchodera:
Should we merge this PR?  It looks like breaking the tests into different build matrix branches helps find errors more quickly.

pgrinaway:
Yeah, I think we should merge. I'm not sure what's going on there.

PR #124, opened 2016-02-16
Add py3.4 testing

jchodera:


jchodera:
Merging since @pgrinaway uses python 3.4.

Issue #125, opened 2016-02-16
Atom mapping in SmallMoleculeProposalEngine


pgrinaway:
There is a problem with atom mapping where it does not take into account the environment, resulting in a very incorrect atom map. Fixing this now, but raising an issue for the record.

pgrinaway:
Resolved in #126

PR #126, opened 2016-02-16
Fix atom map

pgrinaway:
Fixed atom mapping to account for the environment. `GeometryEngine` needs some fixes still regarding eligible torsions. Those are next.


pgrinaway:
Addressing #125

jchodera:
Woohoo!

pgrinaway:
Looks like it's the eligible torsions problem. I think this is good to merge.

PR #127, opened 2016-02-17
Make proposal based on topological torsions

pgrinaway:
Still have a few things to do. I've written code to extract topological torsions from parmed. I need to get these into the proposal/logp bits, as well as allow calculation of improper torsions.

### DIFF
--- a/perses/rjmc/geometry.py
+++ b/perses/rjmc/geometry.py
@@ -111,14 +111,9 @@ class FFGeometryEngine(GeometryEngine):
             for atom in atoms_for_proposal:
                 torsion, logp_choice = self._choose_torsion(atoms_with_positions, atom)
 
-                if torsion.atom1 == atom:
-                    bond_atom = torsion.atom2
-                    angle_atom = torsion.atom3
-                    torsion_atom = torsion.atom4
-                else:
-                    bond_atom = torsion.atom3
-                    angle_atom = torsion.atom2
-                    torsion_atom = torsion.atom1
+                bond_atom = torsion.atom2
+                angle_atom = torsion.atom3
+                torsion_atom = torsion.atom4
 
                 #propose a bond and calculate its probability
                 #if it's not a bond, it's a constraint
@@ -197,14 +192,11 @@ class FFGeometryEngine(GeometryEngine):
             atoms_for_proposal = self._atoms_eligible_for_proposal(old_unique_atoms, old_structure, old_atoms_with_positions)
             for atom in atoms_for_proposal:
                 torsion, logp_choice = self._choose_torsion(old_atoms_with_positions, atom)
-                if torsion.atom1 == atom:
-                    bond_atom = torsion.atom2
-                    angle_atom = torsion.atom3
-                    torsion_atom = torsion.atom4
-                else:
-                    bond_atom = torsion.atom3
-                    angle_atom = torsion.atom2
-                    torsion_atom = torsion.atom1
+
+                bond_atom = torsion.atom2
+                angle_atom = torsion.atom3
+                torsion_atom = torsion.atom4
+
                 #get the internal coordinate representation
                 #we want to see the probability of old_atom where all other coordinates are new_atom
                 atom_coords = old_coordinates[atom.idx]
@@ -584,7 +576,7 @@ class FFGeometryEngine(GeometryEngine):
         """
         Pick an eligible torsion uniformly
         """
-        eligible_torsions = self._get_topological_torsions(atoms_with_positions, new_atom)
+        eligible_torsions = self._get_topological_torsions(atoms_with_positions, atom_for_proposal)
         if len(eligible_torsions) == 0:
             raise Exception("No eligible torsions found for placing atom %s." % str(atom_for_proposal))
         torsion_idx = np.random.randint(0, len(eligible_torsions))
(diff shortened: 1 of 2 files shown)


pgrinaway:
Question: how arbitrary is the ordering of the atoms in a topological torsion (1-2-3-4 vs 4-3-2-1)? I realize this probably shifts the angle, but since the potentials are calculated with their own angle (in the correct order for that torsion), should it make a difference when finding topological torsions?

pgrinaway:
This is probably a good time to move to OpenMM-based potential calculation.

jchodera:
> Question: how arbitrary is the ordering of the atoms in a topological torsion (1-2-3-4 vs 4-3-2-1)? I realize this probably shifts the angle, but since the potentials are calculated with their own angle (in the correct order for that torsion), should it make a difference when finding topological torsions?

The torsion angle computed from 1-2-3-4 should be identical to the one computed from 4-3-2-1.

jchodera:
> This is probably a good time to move to OpenMM-based potential calculation.

+1 for that.  It would allow us to support CHARMM and other potentials.

pgrinaway:
Thanks. I'll implement that stuff tonight then.

pgrinaway:
Going to sleep. Still need to finish the torsion stuff, but most of it is done, and a bunch of duplicate code has been removed.

pgrinaway:
Ok. Looks like the issue that I'm having right now is with the OpenMM context. When I try to instantiate one, it complains:
```pytb
  File "/Users/grinawap/PycharmProjects/perses/perses/rjmc/geometry.py", line 96, in propose
    logp_proposal, new_positions = self._logp_propose(top_proposal, current_positions, beta, direction='forward')
  File "/Users/grinawap/PycharmProjects/perses/perses/rjmc/geometry.py", line 184, in _logp_propose
    context = openmm.Context(growth_system, integrator, platform)
  File "/Users/grinawap/anaconda/envs/openeye-works/lib/python3.4/site-packages/simtk/openmm/openmm.py", line 6250, in __init__
    this = _openmm.new_Context(*args)
Exception: Unknown variable in expression: growth_stage
```

despite the fact that I have added the global parameter to each of the forces. I assume I'm missing something, but I'm not sure what. [This](https://github.com/choderalab/perses/pull/127/files#diff-a5d8d93e4a23e84d32747673e1244e75R659) is the method that creates the system.

pgrinaway:
Nevermind, found my mistake.

pgrinaway:
Debugging some other stuff

perses/rjmc/geometry.py
```diff
@@ -582,162 +504,402 @@ def _propose_angle(self, angle, beta):
...
+        modified_angle_force.addPerAngleParameter("K")
+        modified_angle_force.addPerAngleParameter("growth_idx")
+        modified_angle_force.addGlobalParameter(parameter_name, 0)
+
+        modified_torsion_force = openmm.CustomTorsionForce(self._PeriodicTorsionForceEnergy.format(parameter_name))
+        modified_torsion_force.addPerTorsionParameter("periodicity")
+        modified_torsion_force.addPerTorsionParameter("phase")
+        modified_torsion_force.addPerTorsionParameter("k")
+        modified_torsion_force.addPerTorsionParameter("growth_idx")
+        modified_angle_force.addGlobalParameter(parameter_name, 0)
```
jchodera: This should be `modified_torsion_force`

pgrinaway:
Currently an issue retrieving constraints...

pgrinaway:
I have another angle issue...

pgrinaway:
I wonder if there should be multiple torsion forces on the same atom set. Is this how OpenMM deals with multiplicity, or does this sound like a bug somewhere?

pgrinaway:
Honestly not sure what's going on here. I've attached a plot of p(phi) for adding a methyl to ethane.

Dark blue - first carbon
Dark green - first hydrogen
Red - second hydrogen
teal - third hydrogen.

Kind of looks as I'd expect. However, the angles on the methyl are off (~127deg instead of ~108). The forces seem to be correct, including their growth stage as well.

adding_methyl

pgrinaway:
Found the problem. I was turning forces off instead of on.

jchodera:
Ooh!

pgrinaway:
Have a weird issue retrieving constraints right now. Looking into it. Hopefully then we will be near the end of this saga...

pgrinaway:
On the constraint front, it looks like the remaining weirdness is that there are C-H pairs where there is neither a bond term nor a constraint (there should be a constraint). This is in the system that `GeometryEngine` accepts as input, so I wonder if there is a problem building that system (propane...). I'm playing around with it now.

pgrinaway:
Changed API. Fixed.

pgrinaway:
Hm, the torsion scan potentials can still get `NaN`dy for transformations such as benzene-biphenyl (virtually all the positions have `NaN` potential). If the angles are excluded (by not adding them to the `growth_system`), then the `NaN`s go away, suggesting they come from the angle force. This is a bit disturbing. I need to prepare for tomorrow but I will continue looking into this.

jchodera:
Can we merge this now (to at least fix the issues with `no eligible torsions found`) and continue on from there?

jchodera:
Also the result for `benzene -> biphenyl` sounds totally reasonable to me.  This molecule is essentially completely planar, so most torsion options should be rejected.  If the concern is how we get adequate sampling of the small region of torsion space that has significant probability, we can certainly come up with a more clever way of doing that, but for now, a fine torsion discretization should do the trick.

pgrinaway:
Ok, in that case, we can merge.

pgrinaway:
Actually, I'm gonna commit one more thing.

pgrinaway:
Ok, merging.

Issue #133, opened 2016-02-17
Exception: Called setPositions() on a Context with the wrong number of positions


jchodera:
I've just figured out an error that I introduced in the samplers design:
```
Traceback (most recent call last):
  File "perses/tests/testsystems.py", line 645, in <module>
    testsystem.designer.run(niterations=100)
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/perses-0.1-py2.7.egg/perses/samplers/samplers.py", line 1027, in run
    self.update()
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/perses-0.1-py2.7.egg/perses/samplers/samplers.py", line 1009, in update
    self.update_samplers()
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/perses-0.1-py2.7.egg/perses/samplers/samplers.py", line 973, in update_samplers
    sampler.update()
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/perses-0.1-py2.7.egg/perses/samplers/samplers.py", line 889, in update
    self.update_sampler()
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/perses-0.1-py2.7.egg/perses/samplers/samplers.py", line 860, in update_sampler
    self.sampler.update()
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/perses-0.1-py2.7.egg/perses/samplers/samplers.py", line 737, in update
    self.update_positions()
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/perses-0.1-py2.7.egg/perses/samplers/samplers.py", line 625, in update_positions
    self.sampler.update()
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/perses-0.1-py2.7.egg/perses/samplers/samplers.py", line 441, in update
    context = self.sampler_state.createContext(integrator=integrator)
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/perses-0.1-py2.7.egg/perses/samplers/samplers.py", line 289, in createContext
    raise Exception(msg)
Exception: Called setPositions() on a Context with the wrong number of positions
System has 430 particles
Positions has 427 particles
```
Currently, both `SamplerState` and `ThermodynamicState` have a `System` object, but only `ThermodynamicState.system` was being updated when a proposal was accepted.

I just cobbled together existing code to build this, so I'll have to think a bit about which one of these should have a `System` object.  Probably not both!

[...]

PR #135, opened 2016-02-17
Bugfixes for samplers

jchodera:
This fixes a number of issues with the samplers, including a quick fix for #133.


jchodera:
This also uses the `pdbfixer` to clean up the input NMR files.

jchodera:
I'm stuck on the `no eligible torsions found` bug, so will merge this.

PR #136, opened 2016-02-17
Started work on test of alkanes with only valence energy terms

jchodera:
This test will check that we get appropriate statistics (equal sampling) of molecules with only valence terms.

### DIFF
--- a/perses/data/gaff-valence-only.xml
+++ b/perses/data/gaff-valence-only.xml
@@ -1,0 +1,5819 @@
+<ForceField>
+ <Info>
+  <DateGenerated>2017-02-17</DateGenerated>
+  <Reference>gaff with only valence terms</Reference>
+  <OriginalFile>gaff.dat</OriginalFile>
+ </Info>
+ <AtomTypes>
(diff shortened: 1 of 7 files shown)


jchodera:
Fixed a minor bug in `determine_proposal_order` where, if no atoms needed to be proposed, `logp_choice` would be undefined.

jchodera:
Addressed #137

jchodera:
I've implemented a work-around for https://github.com/choderalab/perses/issues/137#issuecomment-185511489 that randomizes the positions of the atoms that have not yet been placed.

jchodera:
We're no longer in `nan`dyland, but I'm still concerned about the vastly unfavorable geometry engine `logp` values.

jchodera:
Implemented the more robust fix suggested in https://github.com/pandegroup/openmm/issues/1397

jchodera:
Done for now.  There were lots of fixes, but some issues in #137 still need to be addressed (most notably the weird issues with `logp_reverse`).

This can be merged for now!

pgrinaway:
Are we still good to merge this?

jchodera:
Yes!

PR #143, opened 2016-02-18
More updates

jchodera:


jchodera:
OK, this should hopefully be ready to merge if most tests pass.  This seems to work with NCMC steps = 0, but there seem to be issues with the alchemically modified systems that I still have to debug that cause the energy immediately after the geometry engine to be crazy big.

I'm seeing much more reasonable acceptance probabilities right now, but there is still an increase in potential energy after the geometry engine even for the valence-only alkane test:
```
================================================================================
SAMS sampler iteration    37
--------------------------------------------------------------------------------
Expanded Ensemble sampler iteration       37
MCMC sampler iteration 37
Taking 2500 steps of Langevin dynamics...
Final energy is       12.802 kT
Updating chemical state with ncmc-geometry-ncmc scheme...
Proposing new topology...
proposed SMILES string: CCCCC
LOG: MOL
LOG: MOL
Proposed transformation: CC => CCCCC
Generating new system...
Performing NCMC annihilation
Geometry engine proposal...
       2 logp_r        2.596 | logp_theta        1.599 | logp_phi        0.461 | log(detJ)       -4.487
Performing NCMC insertion
potential before geometry  :       12.802 kT
potential after geometry   :       34.555 kT
---------------------------------------------------------
switch_logp                :      -21.753
geometry_logp_propose      :      -13.318
geometry_logp_reverse      :       -0.929
logp_accept = -6.1259e+00 [logp_proposal +8.3333e-02 geometry_logp +1.2389e+01 switch_logp -2.1753e+01 ncmc_elimination_logp +0.0000e+00 ncmc_introduction_logp +0.0000e+00 old_log_weight -3.1543e+00 new_log_weight +0.0000e+00]
    rejected
--------------------------------------------------------------------------------
Writing frame...
================================================================================
```
I *think* this is to be expected, however.  We should see ~kT/2 per degree of freedom, and in this case, adding three methylene groups translates to the addition of 9 atoms, which should be associated with an increase of ~13.5 kT, which is at least the same order of magnitude as the 34-12 = 22 kT of energy increase we see.  We can probably use this as the basis for a geometry engine test: If you try many additions, the average increase in potential energy should be ~(3/2) (N_new - N_old) kT.

pgrinaway:
I think we decided we can merge this.

PR #146, opened 2016-02-20
Fix the geometry logp_reverse by copying coordinates.

pgrinaway:
Potentially WIP, though it does work on my machine.


pgrinaway:
To be clear, the issue here was that the torsion scan modified `old_positions`, leaving them in a messed-up position for the next round.

pgrinaway:
Any idea why travis is not doing anything here?

jchodera:
> To be clear, the issue here was that the torsion scan modified old_positions, leaving them in a messed-up position for the next round.

OOOOOOOH

jchodera:
This looks much better!  Want to merge and we can deal with the other issues separately? 

Looks like there might be some mutation proposal issues and some topology proposal issues (where the wrong number of atoms is present).

pgrinaway:
Sounds good!

PR #150, opened 2016-02-23
Fix atom mapping, proposal

pgrinaway:
Ok, this is marked as WIP but hopefully it will pass sufficient tests that we can merge it. I am now proposing molecules with a probability proportional to the maximum* number of atoms in common, with 0 probability for remaining in the current state and 0 for transitioning to a state with no common atoms.

*MCSS returns multiple matches.


pgrinaway:
Ok, something is still up with the atom map.

pgrinaway:
If tests pass then we can probably merge this.

PR #153, opened 2016-02-25
Drastically curtail number of iterations for samplers tests

jchodera:
This should greatly reduce travis testing time.


jchodera:
Well, I think this is at least quicker, so maybe we can merge?

Some tests on large systems still take up to 2 minutes.

jchodera:
Going to merge this since the next PR is very different.

PR #158, opened 2016-02-29
Deep copy for new topology

juliebehr:


jchodera:
Awesome.  There shouldn't be any problem with merging this, so I'll merge and test locally.

Issue #164, opened 2016-03-01
Custom*Forces are nonzero when lambda=0


pgrinaway:
When the "initial potential components" are printed by the `NCMCEngine`, we get something like this:
```
initial potential before 'insert' : 336004.777863 kT
initial potential components:   [('HarmonicBondForce', Quantity(value=189.8220866815897, unit=kilojoule/mole)), ('CustomBondForce', Quantity(value=21316.236561565125, unit=kilojoule/mole)), ('HarmonicAngleForce', Quantity(value=109.75259884087667, unit=kilojoule/mole)), ('CustomAngleForce', Quantity(value=3027.3701748361514, unit=kilojoule/mole)), ('PeriodicTorsionForce', Quantity(value=176.6403225078547, unit=kilojoule/mole)), ('CustomTorsionForce', Quantity(value=0.0, unit=kilojoule/mole)), ('NonbondedForce', Quantity(value=837634.6902652165, unit=kilojoule/mole)), ('CustomNonbondedForce', Quantity(value=8364162186458531.0, unit=kilojoule/mole)), ('CustomNonbondedForce', Quantity(value=-434.5643798153403, unit=kilojoule/mole)), ('CustomBondForce', Quantity(value=2966343.3743730546, unit=kilojoule/mole)), ('CMMotionRemover', Quantity(value=0.0, unit=kilojoule/mole))]
```

Since `lambda` is 0 before the insertion, shouldn't those forces have no contribution (or near 0)? Or am I missing something?

jchodera:
Yep.  That looks like a bug...

pgrinaway:
I'm happy to dig deeper, but do you think the bug is in `alchemy`? or maybe even `OpenMM`?

jchodera:
I thought the issue is that we don't currently specify `lambda_bonds`, `lambda_angles`, and `lambda_torsions` in [`default_functions` here](https://github.com/choderalab/perses/blob/master/perses/annihilation/ncmc_switching.py#L7-L13), but we do specify these.

Are you specifying your own functions somewhere in `perses`, instead of using the defaults?

pgrinaway:
I don't think so. Let me double check.

jchodera:
You're installing `alchemy` from the repo, or from `conda`?

pgrinaway:
I think I may have found the problem

pgrinaway:
(it was from conda previously)

pgrinaway:
Nevermind...

[...]

PR #166, opened 2016-03-01
Fix reporting of energy components during NCMC

jchodera:
Bugfix for #164


jchodera:
This also turns on alchemical annihilation of bonds and angles during the NCMC phase.

PR #167, opened 2016-03-07
[WIP] Add torsions for ring closure

pgrinaway:
This PR features/will feature a few improvements:

* Addition of a torsion term to non-rotatable bonds to assist in ring closure.

* Computation of internal coordinates via OpenEye (if this is faster, which it probably is)

* Possible other speedups (such as omitting irrelevant atoms from the `System` that `GeometryEngine` uses).

### DIFF
--- a/perses/rjmc/geometry.py
+++ b/perses/rjmc/geometry.py
@@ -826,10 +826,13 @@ class ProposalOrderTools(object):
     ----------
     topology_proposal : perses.rjmc.topology_proposal.TopologyProposal
         The topology proposal containing the relevant move.
+    add_extra_torsions : bool, optional
+        Whether to add additional torsions to keep rings flat. Default true.
     """
 
-    def __init__(self, topology_proposal):
+    def __init__(self, topology_proposal, add_extra_torsions=True):
         self._topology_proposal = topology_proposal
+        self._add_extra_torsions = add_extra_torsions
 
     def determine_proposal_order(self, direction='forward'):
         """
--- a/perses/rjmc/geometry.py
+++ b/perses/rjmc/geometry.py
@@ -818,9 +818,7 @@ class GeometrySystemGenerator(object):
         torsion_force : openmm.CustomTorsionForce
             The torsion force with extra torsions added appropriately.
         """
-        import simtk.openmm.app as app
         import openmoltools.forcefield_generators as forcefield_generators
-        reference_topology = app.Topology()
         atoms = list(reference_topology.atoms())
 
         #get residue from first atom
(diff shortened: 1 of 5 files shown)


perses/rjmc/geometry.py
```diff
@@ -790,6 +794,68 @@ def create_modified_system(self, reference_system, growth_indices, parameter_nam
...
+        omega(oemol)
+
+        #get the list of torsions in the molecule that are relevant
+        torsion_list = list(oechem.OEGetTorsions(oemol))
+        relevant_torsion_list = []
+        for bond in non_rotor_bonds:
+            atom1 = bond.GetBgn()
+            atom2 = bond.GetEnd()
+            for torsion in torsion_list:
+                if torsion.b==atom1 or torsion.b==atom2 and torsion.c==atom1 or torsion.c==atom2:
```
jchodera: Shouldn't this be
```python
if (torsion.b==atom1 and torsion.c==atom2) or (torsion.c==atom1 and torsion.b==atom2):
```
?

jchodera:
I think you can use [`OEGetTorsion`](https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemFunctions.html#OEChem::OEGetTorsion) to measure torsions from the `OEMol` with omega-provided geometries.  

Even easier would be to iterate over all `isRotor=True` torsions by simply using [`OEGetTorsions`](https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemFunctions.html#OEChem::OEGetTorsions) directly:
```python
for torsion in OEGetTorsions(molecule):
    # use torsion.radians directly...
```

pgrinaway:
>I think you can use OEGetTorsion to measure torsions from the OEMol with omega-provided geometries.

I'm aware--I'm also extending this to the other coordinate conversion stuff. I just decided to open the pull request before I finished--it's a WIP. 

>Even easier would be to iterate over all isRotor=True torsions by simply using OEGetTorsions directly:

Ok. Wouldn't I want `isRotor=False` though?

jchodera:
> Ok. Wouldn't I want isRotor=False though?

Whoops!  Yes!

pgrinaway:
Ok, thanks, simplified the torsion collection considerably with that parameter. As I brought up [here](https://github.com/choderalab/openmoltools/issues/203), it might be easier/less error-prone if the atoms in the `OEMol` that is extracted from the `topology` retain an `old_index` or something in their data. If the issue on `openmoltools` looks good, I can add that.

pgrinaway:
Hm, it doesn't look like this is helping much. Both the benzene-->biphenyl and erlotinib-->imatinib examples fail in the usual way. I did check to make sure the torsion was being applied, and that it didn't give ridiculous energy for the omega configuration, so I'm not sure what the issue here is.

jchodera:
Interesting! Let me take a look at the PR and see if I can spot anything.

pgrinaway:
Thanks!

pgrinaway:
Hm, I think I'm going to merge this and work on optimizing the `GeometryEngine` torsion calculation in the next PR.

jchodera:
Please do!

pgrinaway:
Ok.

PR #168, opened 2016-03-09
Write atomwise PDB file for both forward and reverse geometry proposals

jchodera:
This adds writing of atomwise multi-model PDB files for both forward and reverse proposals.


pgrinaway:
Oh, sorry, I must have missed this. Do you want to merge this? Or is there still some work to be done here?

jchodera:
I think it's useful for debugging and illustration purposes, provided the changes I made don't interfere with the actual algorithm.  Perhaps we can add a `writeProposalPDB` flag to the geometry engine and have this off by default?  If that make sense, I'll add that and then you can merge this if everything looks good.

pgrinaway:
I agree that it's useful. Your plan sounds good--just checking to see if anything else needed to be done.

jchodera:
OK, I've made this PDB sequential atom placement writing optional. Should be good to review and merge.

jchodera:
Merging this.

PR #170, opened 2016-03-13
Fix geometry engine torsion biases and MCSS overlap criteria

jchodera:
I've turned on writing of atom-by-atom PDBs for geometry proposals to debug what is going wrong with kinase inhibitors.

It looks like the torsion restraints are not actually biasing the placement of closed rings.  I don't know if there was somehow a reversion somewhere, since this was working before.

### DIFF
--- a/perses/rjmc/geometry.py
+++ b/perses/rjmc/geometry.py
@@ -76,7 +76,7 @@ class FFAllAngleGeometryEngine(GeometryEngine):
     def __init__(self, metadata=None):
         self._metadata = metadata
         self.write_proposal_pdb = False # if True, will write PDB for sequential atom placements
-        self.pdb_filename_prefix = None # PDB file prefix for writing sequential atom placements
+        self.pdb_filename_prefix = 'geometry-proposal' # PDB file prefix for writing sequential atom placements
         self.nproposed = 0 # number of times self.propose() has been called
 
     def propose(self, top_proposal, current_positions, beta):
--- a/perses/rjmc/geometry.py
+++ b/perses/rjmc/geometry.py
@@ -298,6 +298,12 @@ class FFAllAngleGeometryEngine(GeometryEngine):
         if self.write_proposal_pdb:
             pdbfile.close()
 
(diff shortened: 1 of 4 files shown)


jchodera:
I've made two major changes in this PR that fix the geometry engine failures for kinase inhibitors:
* MCSS criteria are modified to
```
        atomexpr = oechem.OEExprOpts_Aromaticity | oechem.OEExprOpts_RingMember | oechem.OEExprOpts_HvyDegree
        bondexpr = oechem.OEExprOpts_Aromaticity | oechem.OEExprOpts_RingMember 
```
* Torsion bias phase is mapped to `[-pi,+pi)` interval, which seems to be the domain OpenMM actually uses for torsion biases.  Previously, the domain was `[0,2*pi)`, which was causing the torsion bias to be totally unreliable.

@pgrinaway : Can you review and merge?

jchodera:
When running `python perses/tests/testsystems.py`, I've noticed that the `geometry-proposal-*-forward-initial.pdb` structures look great, but the `kinase-inhibitors-vacuum-geometry-proposals.pdb` are terrible. I think there must be some topology corruption or mixup between the generation of these two files.

jchodera:
I can't figure out what is going on.  The atom-by-atom PDB files in `_logp_propose` look great now, but the resulting `(topology_proposal.new_topology, geometry_new_positions)` pair seems to be totally messed up:
https://github.com/choderalab/perses/blob/master/perses/samplers/samplers.py#L723-L731

jchodera:
The problem is just with the multi-model PDB files being read incorrectly by PyMOL. Trying to find a workaround.

jchodera:
I'm not sure how to fix the issues with PyMOL reading multi-model connectivity rendering, so for now, I am just concatenating PDB files into a single file. This renders in PyMOL as separate objects---one per frame---but the manual inspection of structures looks good at this point.

Should be ready to review and merge.

pgrinaway:
This looks fine to merge.

PR #173, opened 2016-03-15
New atoms for OEMol purpose get -1 index

pgrinaway:
This PR is done. All it does is assign an "index" of -1 for the atoms that we add to make an `OEMol`. This doesn't impact anything else, but will make it easy for @juliebehr to distinguish these atoms when making an atom map between protein residues.


juliebehr:
Yay!

pgrinaway:
Ok, I'm going to merge this to get out of the way.

jchodera:
Thanks!

PR #174, opened 2016-03-15
Changing atom map to use MCSS for modified residues

juliebehr:


juliebehr:
Updating the geometry engine test, then this will be ready!

juliebehr:
The OpenEye failure for LYS, ASP and GLU has returned.  ```test_geometry_engine``` fails at ```Proposed mutation: ASN 2 ASP```:
```
Proposed mutation: ASN 2 ASP
(<Atom 4 (C) of chain 0 residue 0 (ACE)>, <Atom 6 (N) of chain 0 residue 1 (ASN)>)
(<Atom 18 (C) of chain 0 residue 1 (ASN)>, <Atom 20 (N) of chain 0 residue 2 (NME)>)
internal atom
<Atom 6 (N) of chain 0 residue 1 (ASN)>
Adding H to N
internal atom
<Atom 18 (C) of chain 0 residue 1 (ASN)>
Adding OH to C
(<Atom 4 (C) of chain 0 residue 0 (ACE)>, <Atom 6 (N) of chain 0 residue 1 (ASP)>)
(<Atom 15 (C) of chain 0 residue 1 (ASP)>, <Atom 18 (N) of chain 0 residue 2 (NME)>)
internal atom
<Atom 6 (N) of chain 0 residue 1 (ASP)>
Adding H to N
internal atom
<Atom 15 (C) of chain 0 residue 1 (ASP)>
Adding OH to C
  File "test_geometry_engine.py", line 584, in <module>
    test_mutate_from_all_to_all()
  File "test_geometry_engine.py", line 190, in test_mutate_from_all_to_all
    context.setPositions(new_positions)
  File "/home/behrj/miniconda/lib/python2.7/site-packages/simtk/openmm/openmm.py", line 13580, in setPositions
    return _openmm.Context_setPositions(self, positions)
Exception: Called setPositions() on a Context with the wrong number of positions

```

jchodera:
That's not the same OpenEye error as we had before---it doesn't look like there is any OpenEye error at all.  It's just that the `System` and `positions` differ in number of particles.

juliebehr:
That's in addition to the OpenEye thing.

juliebehr:
```
Warning: Can't assign MMFF type to atom number    10  CG  on residue ASP  1
Warning: Can't assign MMFF type to atom number    11  OD1 on residue ASP  1
Warning: Can't assign MMFF type to atom number    12  OD2 on residue ASP  1
Warning: force field setup failed due to missing parameters for molecule new_res
```

jchodera:
Any chance you have yet to update `openmoltools` locally?  I thought I had fixed the OpenEye errors there.

juliebehr:
OH this is definitely possible, I didn't realize that was where that fix was made, one second...

juliebehr:
I had been hoping the first error was leading to missing information causing the second but that is not the  case.  @pgrinaway ideas how the ```new_positions``` generated [here] (https://github.com/choderalab/perses/pull/174/files#diff-2a253a943bbd76b58d717e523aebaa2cR184) would end up being the wrong length?  There might be something wrong with how I reference the new system but it fails on ```ASN-2-ASP``` after successfully making 44 other mutations.

pgrinaway:
Looks like there is an issue with the atom map that `GeometryEngine` is getting, resulting in it producing a longer-than-appropriate positions array. Will continue debugging

pgrinaway:
Oh, I think I see what happened here. If `GeometryEngine` sees no `unique_new_atoms`, it returns the old positions with a logp of 0. However, in this case, atoms were actually deleted, with no unique new ones, so the position vector was the incorrect length. Not sure if it's appropriate to have no unique new atoms, but the `GeometryEngine` should handle this better.

juliebehr:
The lack of unique atoms is based on the ```atomexpr``` going into the MCSS -- it isn't distinguishing a difference between OD1 and ND1 of ASP and ASN respectively.
```
atomexpr = oechem.OEExprOpts_Aromaticity | oechem.OEExprOpts_RingMember | oechem.OEExprOpts_HvyDegree
bondexpr = oechem.OEExprOpts_Aromaticity | oechem.OEExprOpts_RingMember
```

pgrinaway:
@juliebehr the fix for the position mismatch is pretty straightforward for me. Do you want me to commit that to this branch? Or would that not be good?

juliebehr:
Is there a reason we were omitting ```AtomicNumber``` ?

juliebehr:
@pgrinaway yes that's fine!

pgrinaway:
Ok, one sec. Not sure about your other question.

juliebehr:
I don't think it matters for the time being, since the point of this PR is to make those expressions easily modifiable if necessary to improve the quality of position proposals -- but that would be why there aren't any unique atoms being found in ASP.

pgrinaway:
Yep, I think you're right. Try it now?

juliebehr:
```
Proposed mutation: PRO 2 ARG
(<Atom 4 (C) of chain 0 residue 0 (ACE)>, <Atom 6 (N) of chain 0 residue 1 (PRO)>)
(<Atom 18 (C) of chain 0 residue 1 (PRO)>, <Atom 20 (N) of chain 0 residue 2 (NME)>)
internal atom
<Atom 6 (N) of chain 0 residue 1 (PRO)>
Adding H to N
internal atom
<Atom 18 (C) of chain 0 residue 1 (PRO)>
Adding OH to C
(<Atom 4 (C) of chain 0 residue 0 (ACE)>, <Atom 6 (N) of chain 0 residue 1 (ARG)>)
(<Atom 18 (C) of chain 0 residue 1 (ARG)>, <Atom 30 (N) of chain 0 residue 2 (NME)>)
internal atom
<Atom 6 (N) of chain 0 residue 1 (ARG)>
Adding H to N
internal atom
<Atom 18 (C) of chain 0 residue 1 (ARG)>
Adding OH to C
  File "test_geometry_engine.py", line 585, in <module>
    test_mutate_from_all_to_all()
  File "test_geometry_engine.py", line 184, in test_mutate_from_all_to_all
    new_positions, logp = geometry_engine.propose(pm_top_proposal, current_positions, beta)
  File "/home/behrj/miniconda/lib/python2.7/site-packages/perses-0.1-py2.7.egg/perses/rjmc/geometry.py", line 106, in propose
    logp_proposal, new_positions = self._logp_propose(top_proposal, current_positions, beta, direction='forward')
  File "/home/behrj/miniconda/lib/python2.7/site-packages/perses-0.1-py2.7.egg/perses/rjmc/geometry.py", line 187, in _logp_propose
    atom_proposal_order, logp_choice = proposal_order_tool.determine_proposal_order(direction='forward')
  File "/home/behrj/miniconda/lib/python2.7/site-packages/perses-0.1-py2.7.egg/perses/rjmc/geometry.py", line 1119, in determine_proposal_order
    chosen_torsion, logp_choice = self._choose_torsion(atoms_with_positions, atom)
  File "/home/behrj/miniconda/lib/python2.7/site-packages/perses-0.1-py2.7.egg/perses/rjmc/geometry.py", line 1166, in _choose_torsion
    raise Exception("No eligible torsions found for placing atom %s." % str(atom_for_proposal))
Exception: No eligible torsions found for placing atom <Atom NE [21]; In ARG 1>.
```

pgrinaway:
Oh. This is either an error in `GeometryEngine`, or there are no available torsions to match these two amino acids (I think the latter). This is an issue in #114 --I was going to solve it for small molecules by simply not proposing moves between molecules with no torsions in common.

 Is the issue here that the backbone carbon is changing? Maybe we just need to keep it?

pgrinaway:
The alternative is that `GeometryEngine` can have a proposal mode when no torsions are available (papers like [this one](http://infoscience.epfl.ch/record/200758/files/smi952.pdf) do that)

juliebehr:
Hmm, if the problem is CA I feel like there should be some sort of special condition where its position is just kept.

juliebehr:
The problem is just that it's only matching 2 atoms, so there aren't enough for a torsion.  It starts at NE because there are multiple 2 atom matches that ```np.random.choice``` chooses between, one of which is matching the C and O from the PRO backbone to CZ (the carbon between the 3 nitrogens) and the double bonded N.

juliebehr:
I wonder if there's a way to enforce MCSS aligning the backbones to each other? If so, CA can just be added manually.

jchodera:
> I wonder if there's a way to enforce MCSS aligning the backbones to each other? If so, CA can just be added manually.

You can do this with [`AddConstraint`](https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemClasses/OEMCSSearch.html#OEChem::OEMCSSearch::AddConstraint)

juliebehr:
Perfect! Trying this now.

juliebehr:
The ```AddConstraint```s [here] (https://github.com/choderalab/perses/pull/174/files#diff-bd95c4342ca1beb79ba546696bfdc262R640) are returning ```False``` and I'm not sure why?  The [documentation](https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemClasses/OEMCSSearch.html#OEChem::OEMCSSearch::AddConstraint) suggests this should only be happening if ```old_backbone_c``` is not in ```oegraphmol_current```, but it must be, because that's where I got it from?

jchodera:
I can't figure this out either. Try OpenEye support?

jchodera:
Oh, you may need to use `mcs.GetPattern()` to get a read-only version of the query molecule and find the atoms you want to constrain in that version of the molecule:

https://docs.eyesopen.com/toolkits/python/oechemtk/patternmatch.html

"A read-only reference to the query molecule OEQMolBase contained in an instance of OEMCSSearch can be obtained with the GetPattern method. Note that if the OEMCSSearch was constructed with an OEQMolBase, the returned OEQMolBase is a separate object. OEQMolBase methods can then be used to interrogate the returned OEQMolBase reference."

juliebehr:
Currently running tests, but otherwise I think this will be ready with the next commit.

juliebehr:
Good to go.

jchodera:
Awesome, thanks! 

@pgrinaway, can you review and merge?

pgrinaway:
Going through now!

pgrinaway:
Looks good!

PR #175, opened 2016-03-16
Geometry testsystems

pgrinaway:
I'm adding additional tests for the `GeometryEngine` here. I've added tests that propose all lysozyme ligands to each other, as well as all kinase inhibitor ligands to each other.

### DIFF
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -22,6 +22,7 @@ requirements:
     - numexpr
     - openmmtools
     - pymbar
+    - numba
     - autograd
 
   run:
@@ -35,6 +36,7 @@ requirements:
     - numexpr
     - openmmtools
     - pymbar
+    - numba
     - autograd
 
 test:
--- a/perses/data/T4-inhibitors.xml
+++ b/perses/data/T4-inhibitors.xml
@@ -1,0 +1,131 @@
+<ForceField>
+ <Info>
+  <DateGenerated>2016-03-21</DateGenerated>
+ </Info>
+ <HarmonicAngleForce>
+  <Angle type1="h1" type2="c3" type3="i" angle="1.8324211816688465" k="323.17216"/>
+  <Angle type1="ca" type2="n2" type3="n1" angle="2.008699436120274" k="631.3656000000001"/>
+  <Angle type1="cl" type2="c3" type3="h3" angle="1.8699457605867247" k="338.40192"/>
+  <Angle type1="cc" type2="cd" type3="sh" angle="1.9469147805996745" k="547.60192"/>
+  <Angle type1="sh" type2="cd" type3="ss" angle="2.118306113145518" k="522.49792"/>
+  <Angle type1="ca" type2="c3" type3="sh" angle="1.931381350256925" k="517.9792"/>
+  <Angle type1="ca" type2="n4" type3="o" angle="2.0546888619103245" k="555.769088"/>
+  <Angle type1="ca" type2="n4" type3="f" angle="1.9506672384914625" k="553.969968"/>
+  <Angle type1="f" type2="n4" type3="o" angle="2.0073031727186783" k="587.93568"/>
+ </HarmonicAngleForce>
+ <PeriodicTorsionForce>
+  <Proper type1="ca" type2="ca" type3="cg" type4="c1" periodicity1="2" phase1="3.141592653589793" k1="0.0"/>
+  <Proper type1="ca" type2="ca" type3="ce" type4="c2" periodicity1="2" phase1="3.141592653589793" k1="2.9288"/>
+  <Proper type1="ca" type2="ca" type3="ce" type4="ha" periodicity1="2" phase1="3.141592653589793" k1="2.9288"/>
+  <Proper type1="cc" type2="cd" type3="ss" type4="cd" periodicity1="2" phase1="3.141592653589793" k1="4.6024"/>
+  <Proper type1="cc" type2="cd" type3="ss" type4="c3" periodicity1="2" phase1="3.141592653589793" k1="4.6024" periodicity2="3" phase2="3.141592653589793" k2="2.9288"/>
+  <Proper type1="ss" type2="cd" type3="ss" type4="cd" periodicity1="2" phase1="3.141592653589793" k1="4.6024"/>
+  <Proper type1="h4" type2="cd" type3="ss" type4="cd" periodicity1="2" phase1="3.141592653589793" k1="4.6024"/>
+  <Proper type1="ss" type2="cd" type3="ss" type4="c3" periodicity1="2" phase1="3.141592653589793" k1="4.6024"/>
+  <Proper type1="ca" type2="ca" type3="cc" type4="cd" periodicity1="2" phase1="3.141592653589793" k1="2.9288"/>
+  <Proper type1="ca" type2="ca" type3="cc" type4="ha" periodicity1="2" phase1="3.141592653589793" k1="2.9288"/>
+  <Proper type1="ca" type2="ca" type3="cd" type4="cc" periodicity1="2" phase1="3.141592653589793" k1="2.9288"/>
+  <Proper type1="ca" type2="ca" type3="cd" type4="ha" periodicity1="2" phase1="3.141592653589793" k1="2.9288"/>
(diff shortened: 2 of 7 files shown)


jchodera:
Are all of these proposals intended to be valid?  Can we really propose transitions from any kinase inhibitor to any other one?  That would be awesome, but seems like it might be overly strict.

pgrinaway:
>Are all of these proposals intended to be valid? Can we really propose transitions from any kinase inhibitor to any other one? That would be awesome, but seems like it might be overly strict.

I'm trying it out--you could be correct.

pgrinaway:
Ok I think we'll have to restrict transformations. Also, the setup phase is _extremely_ slow--profiling suggests it's the AM1 calculation. Should I somehow restrict the list of inhibitors that we are running, or should we run this test elsewhere?

jchodera:
> Ok I think we'll have to restrict transformations. Also, the setup phase is extremely slow--profiling suggests it's the AM1 calculation. Should I somehow restrict the list of inhibitors that we are running, or should we run this test elsewhere?

I think we just need to pre-parameterize the list of inhibitors to `ffxml`, store it in `data/`, and preload that `ffxml` file.  That should work now using [`openmoltools.forcefield_generators.generateForceFieldFromMolecules`](https://github.com/choderalab/openmoltools/blob/master/openmoltools/forcefield_generators.py).

pgrinaway:
Yeah, that sounds like a good plan.

pgrinaway:
Ok, these tests are extremely slow. I think now is probably the time to make `GeometryEngine` faster. Working on that now.

jchodera:
Sounds good.  We can probably pare back other tests too if that would help.

pgrinaway:
Well, it might help, but these tests take > 2 hours each (!!!!!!!!).

jchodera:
Holy crap! Which tests in particular?

Also, why is travis still failing with `No module named perses`?

pgrinaway:
The T4 lysozyme and kinase inhibitor tests in this PR--basically it's purely a function of how slow the geometry engine is, since that's the only component being tested here. 

As for the Travis error, I'm not sure. I haven't had a chance to look more closely but it's on the menu.

pgrinaway:
I have several ideas for speeding this up. I'm going to try the most straightforward ones here and talk about the rest Wednesday.

pgrinaway:
I have a `numba` version of coordinate conversion now (gets completely compiled to native code). I have to clean up a few more things (mostly involving casting `float32` arrays to `float64`), but it seems to be quite a bit faster. Will rebenchmark once the loose ends are tied up.

jchodera:
Cool!

pgrinaway:
Hm, I seem to be unable to profile with `numba` (though everything runs fine). Will come up with a minimal example and post on their issue tracker.

pgrinaway:
The proposals have gotten faster but are not as fast as I would have liked--I want to see what the remaining sticking points are.

pgrinaway:
Issue raised, will continue with tests.

jchodera:
Might still be useful to brainstorm if some OpenMM functionality might help
speed things up on the opennm issue tracker.

pgrinaway:
Yep, doing that now. A bit of "poor man's profiling" to see which parts specifically are slow, then going to brainstorm a bit. Will include info in the talk tomorrow.

pgrinaway:
Preliminary tinkering shows that with the new coordinate conversion code, that step is likely no longer the bottleneck. Proposals take ~10s to go from one kinase inhibitor to another (this means over an hour for the kinase inhibitor -> kinase inhibitor test in both directions). The breakdown for an entire proposal of 45 atoms is approximately:

These things happen once per proposal:
Processing growth order/torsion selection: 34ms
Growth system generation: 514ms

These happen once per atom, but I'm providing the sum of each after proposing 45 atoms.
(total) Torsion scan coordinate conversion: 122ms (note that this is less than 3ms per torsion scan, so the switch to `numba` definitely helped--this used to be the biggest component).

(total) Energy calculation in the torsion scan: ~7.55s (~167ms/atom).

So clearly, the slowest factor here is the energy computation. Will think more about how to speed this up.

jchodera:
> Growth system generation: 514ms

We can probably do some optimization of this code at the Python level before going to an inline C++ function that dodges the unit conversion overhead.

> (total) Energy calculation in the torsion scan: ~7.55s (~167ms/atom).

If we can perform the torsion scan on the GPU, I bet we can greatly reduce the overhead here and also position ourselves to add more interactions during the torsion scan to increase RJMC acceptance rates.
* Worst case: We write a C++ `Integrator` plugin that does the torsion scan
* Best case: We work out some way to incorporate torsion scans into a fast `CustomIntegrator` scheme
* Intermediate case: We come up with a more efficient sampling scheme that uses `CustomIntegrator`.
* For now: We can just halve the number of points in the torsion scan, or use a lookup table that places more density in the most likely torsion angles.

jchodera:
For the energy computation, it would also help to know if it is the actual energy computation part or other parts (stripping of coordinates from positions when `context.setPositions()` is called, etc.).

pgrinaway:
I was actually in the process of looking at that--will post in a moment.

pgrinaway:
The time that `context.setPositions()` takes is about 20-25% of the time that `context.getState(getEnergy=True)` takes. However, on OpenCL, position set time is something like 50% of the energy calculation time (which is also longer--however, I'm not sure what device this is using on my laptop).

jchodera:
How about we post a description and timing breakdown to the OpenMM issue tracker? I have my own ideas as to the best way to proceed (via an Integrator plugin that would handle all of the torsion proposals in one shot) but it will be very helpful to get more ideas, especially from Peter.

pgrinaway:
I might prefer to save that for a bit later. At the very least, I'd like to see what the `numba` folks have to say about the profiling problem (I've already raised that issue)--if that can be easily resolved, I'd rather give Peter and the OpenMM crew more reliable numbers*. 

Furthermore, I think it might be worthwhile to explore what other algorithms we might use here that could be less demanding. I imagine that writing a plugin for OpenMM as you describe could be a pretty complex engineering task. Although exploring other proposal mechanisms is also complex, it might provide us with more insight into the problem at hand, and is easier to try once we have an algorithm. 

Finally, we should set a goal for how fast the `GeometryEngine` actually has to be to be useful. Obviously, I'll have to pare back the tests a bit here so that they run in a reasonable amount of time on travis, but once we've done that, maybe it'd be worthwhile to have the users in the lab start using `perses` and from there we can decide the best path forward. 

I'd like to chat about this at the group meeting tomorrow to get insight into priorities, goals, etc., if that's ok with you. If it's ultimately the case that we need to write some CUDA plugins for OpenMM, I'm definitely happy to do that. I'd just like to exhaust a few other options first.

jchodera:
There may be options that don't require writing plugins, but we won't necessarily know about all the possibilities (including those which take little work from our end) without picking the brains of other OpenMM developers/users.  Let's do that as soon as you get reliable benchmark numbers.

I could imagine there are many possibilities that do not involve writing platform-specific plugins but instead simply require a C++-level function but still get us the speed increase we need or use a feature that could be implemented on the OpenMM end that multiple people could use.

Let's brainstorm/discuss more at your group meeting tomorrow.

jchodera:
One other thought on "how fast does this have to be?": The primary criteria is how much computer effort is required to decorrelate in chemical space once the optimal relative free energies of different compounds are known. The metric of merit is the statistical inefficiency for hopping in compound space times the wall clock time per expanded ensemble iteration.  There are many ways to speed things up---including making the geometry proposal moves take longer but have higher acceptance probabilities via including more of the local environmental interactions in torsion proposals, parallel Monte Carlo, more clever NCMC pathways, etc.  But it there is low-hanging fruit---and a straightforward approach to speeding up torsion proposals might be one of them---this might be worth pursuing.

jchodera:
Don't forget to add `numba` to [`devtools/conda-recipe/meta.yaml`](https://github.com/choderalab/perses/blob/geometry-testsystems/devtools/conda-recipe/meta.yaml#L12-L38) and the `setup.py` if you are going to require it.

pgrinaway:
Hm, going to have to resolve conflicts, otherwise this is almost done.

pgrinaway:
Going to make the `meta.yaml` change in a quick PR after this. Conflict resolution was otherwise a bit of a pain.

PR #176, opened 2016-03-22
Attempting to fix travis tests

jchodera:


jchodera:
I'm fixing a few `*-dev` packages to remove the possibility that both `openmm-dev` and `openmm` (and `parmed-dev` and `parmed`) are brought in at the same time.

jchodera:
Testing now runs on travis. Lots of tests fail, but this is ready to be merged!

pgrinaway:
Ok. I'm happy to merge--I think I know what needs to be fixed here.

PR #177, opened 2016-03-23
[WIP] T4LysozymeTestSystem

juliebehr:
Initial commit -- right now T4LysozymeTestSystem is MybTestSystem copied and pasted with the name changed so it obviously isn't going to do anything.


juliebehr:
Why is this [here](https://github.com/choderalab/perses/blob/master/perses/tests/testsystems.py#L430)?

```
    def __init__(self):
        super(AblImatinibTestSystem, self).__init__()
        solvents = ['vacuum', 'explicit'] # TODO: Add 'implicit' once GBSA parameterization for small molecules is working
        solvents = ['vacuum'] # DEBUG
```

juliebehr:
Specifically, does the problem with allowing explicit solvent actually have to do with explicit solvent itself, or with small molecule parameterization?  ```gaff.xml``` does not contain residue information and won't be able to construct templates.

juliebehr:
(When I try to run the T4 lysozyme test system it fails when adding solvent, but because of the ligand)

perses/tests/testsystems.py
```diff
@@ -273,28 +273,30 @@ class T4LysozymeTestSystem(PersesTestSystem):
...
+        system_generators['explicit'] = SystemGenerator([gaff_xml_filename,'amber99sbildn.xml', 'tip3p.xml'],
             forcefield_kwargs={ 'nonbondedMethod' : app.CutoffPeriodic, 'nonbondedCutoff' : 9.0 * unit.angstrom, 'implicitSolvent' : None, 'constraints' : None },
             use_antechamber=False)
         system_generators['explicit-complex'] = system_generators['explicit']
-        system_generators['explicit-peptide'] = system_generators['explicit']
-        system_generators['implicit'] = SystemGenerator(['amber99sbildn.xml', 'amber99_obc.xml'],
+        system_generators['explicit-receptor'] = system_generators['explicit']
+        system_generators['implicit'] = SystemGenerator([gaff_xml_filename,'amber99sbildn.xml', 'amber99_obc.xml'],
             forcefield_kwargs={ 'nonbondedMethod' : app.NoCutoff, 'implicitSolvent' : app.OBC2, 'constraints' : None },
             use_antechamber=False)
```
pgrinaway: I think this should be `use_antechamber=True` (default) to have `SystemGenerator` generate parameters on the fly.

juliebehr: 🙌

jchodera:
> (When I try to run the T4 lysozyme test system it fails when adding solvent, but because of the ligand)

I thought I had added that `DEBUG` statement just to speed up testing, but something else seems to be going wrong when you remove that.  I'll stop by!

perses/tests/testsystems.py
jchodera: Does that fix this?

juliebehr: Sadly no, although I think it is still correct

jchodera:
@juliebehr : The problem seems to be that your PDB file (`181L.pdb`) has a ligand in it that doesn't have any protons.  Remember that `ForceField` will try to parameterize any molecules without parameters, but **these ligands need to have all hydrogens added** since it has to infer the bond orders, aromaticity, and charge state from the connectivity in `Topology` and element names.

You will need to strip out the small molecule, process it with OpenEye to add protons, and add it back.  It may be easier to just create the molecule from IUPAC name with OpenEye tools and do a shape overlay, stripping the original small molecule out.  We should probably automate that process at some point---@andrrizzi may have already done so for YANK.

To make it easier to test/debug, I've added some code that lets you just run this test directly with
```bash
python perses/tests/testsystems.py
```

Note also that we don't have GBSA for small molecules working yet, so I have commented that out.

You probably also want to strip the crystallographic waters out of the PDB file, since when we do get implicit solvent working, those won't be wanted there.

andrrizzi:
In `yank` I added back protons through `leap`, but this will likely change atom types. Also, I have never played with shape overlay in OpenEye yet. Sorry I can't be of much help.

jchodera:
I don't believe `leap` can add protons to the small molecule, which is what @juliebehr needs to do here.  (`pdbfixer` already takes care of the protein residues.)

juliebehr:
This will run, after taking a very long time, and then I get:
```
================================================================================
SAMS sampler iteration     0
--------------------------------------------------------------------------------
Expanded Ensemble sampler iteration        0
MCMC sampler iteration 0
Taking 50 steps of Langevin dynamics...
potential  = 721484656990017.875 kT
kinetic    = 230910519852.777 kT
force norm = 912290.656 kT/A/dof
Final energy is          nan kT

```

jchodera:
OH

juliebehr:
Added minimize step, still problems:
```
Minimizing 'explicit-complex'...
Initial energy is 430121724588949.062 kcal/mol
Final energy is     701937.835 kcal/mol
Minimizing 'explicit-receptor'...
Initial energy is   -41005.549 kcal/mol
Final energy is     -80619.295 kcal/mol
Minimizing 'vacuum-complex'...
Initial energy is 430121724648427.000 kcal/mol
Final energy is      -4384.998 kcal/mol
Minimizing 'vacuum-receptor'...
Initial energy is    12345.586 kcal/mol
Final energy is      -4762.848 kcal/mol
================================================================================
SAMS sampler iteration     0
--------------------------------------------------------------------------------
Expanded Ensemble sampler iteration        0
MCMC sampler iteration 0
Taking 50 steps of Langevin dynamics...
potential  = 1177427.981 kT
kinetic    = 112537.615 kT
force norm = 356.626 kT/A/dof
Final energy is          nan kT

```

juliebehr:
I'm guessing the benzene is still messed up, currently investigating

juliebehr:
Pretty sure this is just a unit error in assigning coordinates to the benzene OEMol, confirming and fixing if so...

jchodera:
Added this code to write out a PDB file after you add the benzene:
```python
        # DEBUG: Write initial coordinates
        outfile = open('initial.pdb','w')
        PDBFile.writeFile(modeller.getTopology(), modeller.getPositions(), outfile)
        outfile.close()
```
Looks like there are a few problems: 
* The benzene is way outside the pocket
* It's really tiny. Did you forget to convert from Angstroms to nanometers?

jchodera:
> Pretty sure this is just a unit error in assigning coordinates to the benzene OEMol, confirming and fixing if so...

Yes!

perses/tests/utils.py
```diff
@@ -58,6 +58,13 @@ def extractPositionsFromOEMOL(molecule):
         positions[index,:] = unit.Quantity(coords[index], unit.angstroms)
     return positions
 
+def giveOpenmmPositionsToOEMOL(positions, molecule):
+    assert molecule.NumAtoms() == len(positions)
+    coords = molecule.GetCoords()
+    for key in coords.keys():
+        coords[key] = (positions._value[key][0],positions._value[key][1],positions._value[key][2])
```
jchodera: This is very bad!  It just takes the value without converting!

Can you use `positions[key][0] / unit.angstroms`?

juliebehr:
Current status:
```
Minimizing 'explicit-complex'...
Initial energy is   -20739.364 kcal/mol
Final energy is     -20741.473 kcal/mol
Minimizing 'explicit-receptor'...
Initial energy is   -24380.706 kcal/mol
Final energy is     -81211.251 kcal/mol
Minimizing 'vacuum-complex'...
Initial energy is    15589.085 kcal/mol
Final energy is      -4442.742 kcal/mol
Minimizing 'vacuum-receptor'...
Initial energy is    11959.039 kcal/mol
Final energy is      -4767.055 kcal/mol
================================================================================
SAMS sampler iteration     0
--------------------------------------------------------------------------------
Expanded Ensemble sampler iteration        0
MCMC sampler iteration 0
Taking 50 steps of Langevin dynamics...
potential  = -34791.692 kT
kinetic    = 4357782566434.055 kT
force norm = 1591685.782 kT/A/dof
Final energy is          nan kT

```

jchodera:
Somehow, your hydrogens ended up directly on top of your carbons:
```
ATOM   2638  C1      1   1      25.978   5.327   4.779  1.00  0.00           C  
...
ATOM   2644  H1      1   1      25.978   5.327   4.779  1.00  0.00           H  
```

juliebehr:
Oh, weird! Ok, looking through openeye docs now.

jchodera:
I couldn't figure out what was going on there either.

jchodera:
Any luck?

juliebehr:
Not yet

juliebehr:
Giving up on openeye and using geometry engine to propose hydrogen locations.

pgrinaway:
seems good to merge?

juliebehr:
It's fine with me!

juliebehr:
Merging.

PR #180, opened 2016-03-25
Fix dependency

pgrinaway:
Added numba as a dependency (didn't before because of weird git conflicts).


jchodera:
Don't we need to add dependencies to `setup.py` too?

pgrinaway:
Crap, you're right. One second.

jchodera:
Looks good to me!

PR #181, opened 2016-03-25
Speed up alchemical elimination tests

jchodera:
Working on speeding up alchemical elimination tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,12 +57,13 @@ env:
     - python=2.7 CONDA_PY=27 NOSETESTS="perses.tests.test_ncmc_integrator"
     - python=2.7 CONDA_PY=27 NOSETESTS="perses.tests.test_elimination"
     # Main tests for python 3.4
+    # We skip slow tests.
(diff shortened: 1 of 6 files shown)


jchodera:
I'm quite confused as to why the NCMC tests are suddenly failing so drastically.  The energies are pretty crazy, but I'm not sure what has changed to break things.

jchodera:
Apparently, I had fixed the `alchemical_bonds` and `alchemical_angles` arguments to `AbsoluteAlchemicalFactory` so bonds and angles in the molecules were being dissolved, causing an irreversible decomposition of the molecule.  I've fixed this now.

jchodera:
Almost all the tests are passing now!

jchodera:
Lots of tests have been fixed, and test time is now down quite a bit.

The geometry tests are still slow, but I'll let @pgrinaway figure out which ones to trim / clean up.

All remaining test failures seem to be geometry engine issues (or potentially small molecule proposal engine issues).

jchodera:
Look at all those passing tests!

OK, merging this. The rest of the failures are all yours to clean up, @pgrinaway.

PR #186, opened 2016-03-28
Fixing Geometry errors

pgrinaway:
Fixed a couple of issues, still have a couple more to go. WIP, hopefully can fix these today.

### DIFF
--- a/devtools/travis-ci/install.sh
+++ b/devtools/travis-ci/install.sh
@@ -3,7 +3,7 @@ pushd .
 
 # Install Miniconda
 cd $HOME
-MINICONDA=Miniconda-latest-Linux-x86_64.sh
+MINICONDA=Miniconda2-latest-Linux-x86_64.sh
 MINICONDA_MD5=$(curl -s http://repo.continuum.io/miniconda/ | grep -A3 $MINICONDA | sed -n '4p' | sed -n 's/ *<td>\(.*\)<\/td> */\1/p')
 wget http://repo.continuum.io/miniconda/$MINICONDA
 if [[ $MINICONDA_MD5 != $(md5sum $MINICONDA | cut -d ' ' -f 1) ]]; then
--- a/perses/rjmc/geometry.py
+++ b/perses/rjmc/geometry.py
@@ -315,7 +315,8 @@ class FFAllAngleGeometryEngine(GeometryEngine):
                 PDBFile.writeFile(top_proposal.new_topology, new_positions, file=pdbfile)
                 pdbfile.close()
         total_time = time.time() - initial_time
-        logging.log(logging.DEBUG, "Proposal order time: %f s | Proposal order forward: %f s | Growth system generation: %f s | Total torsion scan time %f s | Total energy computation time %f s | Position set time %f s| Total time %f s" % (proposal_order_time, proposal_order_forward, growth_system_time , self._torsion_coordinate_time, self._energy_time, self._position_set_time, total_time))
+        if direction=='forward':
+            logging.log(logging.DEBUG, "Proposal order time: %f s | Growth system generation: %f s | Total torsion scan time %f s | Total energy computation time %f s | Position set time %f s| Total time %f s" % (proposal_order_time, growth_system_time , self._torsion_coordinate_time, self._energy_time, self._position_set_time, total_time))
         self._torsion_coordinate_time = 0.0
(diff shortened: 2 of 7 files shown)


pgrinaway:
Main test issues are fixed. Now:
* issue in #114 (and duplicates) will be fixed. 

* Adding `skipIf` statements

pgrinaway:
Ok, so I implemented a fix for the eligible torsion thing. However, it seems that the current criteria are then sufficient to cause some members of the `ValenceSmallMoleculeLibraryTestSystem` to become unreachable from the rest. I just need to correct the array there, and have a warning saying which molecules have been removed.

jchodera:
> Ok, so I implemented a fix for the eligible torsion thing. However, it seems that the current criteria are then sufficient to cause some members of the ValenceSmallMoleculeLibraryTestSystem to become unreachable from the rest. I just need to correct the array there, and have a warning saying which molecules have been removed.

Do you mean that some molecules cannot be reached from *any* molecule in the set, or that some molecules cannot be reached from *every* molecule in the set?

Put another way: If I make a graph of molecules as nodes connected by edges that correspond to transitions `GeometryEngine` can handle, do I end up with disjoint subgraphs?  Or just a graph that isn't densely connected?

pgrinaway:
It actually ends up with disjoint subgraphs. I suppose this is the tradeoff between very conservative atom maps that delete most of the old atoms, and ones that leave old atoms in at the risk of creating untenable structures.

For now, I was going to just remove those molecules with a warning. Alternatively, we could alter the atom mapping criterion such that all molecules are connected to at least one other. Finally, we could also modify the `GeometryEngine` such that it can propose with less than one torsion (though I'd save this for later).

Next, I'll implement the visualization stuff so that we can have a better look at what's going on!

jchodera:
Thanks for the clarification!

My thinking (not necessarily correct) is to prioritize:
0. bugfixes to get things to run at all
1. visualization (so we know what is going on, can have things for paper, etc.)
2. minimal modifications to get essential systems to be used in the paper (amino acids, T4 lysozyme ligands) to run
3. optimizations to make MCSS criteria more permissive for other ligand sets (kinase inhibitors, etc)
4. optimizations to ligand proposal scheme to increase acceptance probabilities
5. considering how to allow `GeometryEngine` to make more flexible proposals if item 3 still doesn't help

pgrinaway:
I agree with your priorities, thanks for the input!

pgrinaway:
Ok, so I added in code to avert the torsion disaster. But now there is another issue: what if one of the chemical states that the `ProposalEngine` removes is the starting state? I'm tempted to say throwing an exception might be practical here. After all, the user has supplied input that cannot be used.

As an aside, the entire `ValenceSmallMoleculeLibraryTestSystem` is disconnected under the current atom mapping scheme.

pgrinaway:
Perhaps we should include specification of atom mapping criteria in the constructor.

pgrinaway:
Ah, there's another issue with it. Hang on...

pgrinaway:
Ok, I see what's wrong. One second

pgrinaway:
Currently we're retaining 54/115 molecules in the T4 inhibitor test system with current atom mapping options.

jchodera:
> Currently we're retaining 54/115 molecules in the T4 inhibitor test system with current atom mapping options.

Do you think that's enough?  We might try some tinkering to see if we can improve that.

jchodera:
Is this guy good to go if you're up to 115/115 molecules now?

pgrinaway:
Yeah, I think we can merge. Going beyond that will require additional, more complicated changes.

pgrinaway:
Let's see what the current travis tests say about the last commit--should have fixed the sampler issue and topology proposal test issue.

pgrinaway:
Miniconda MD5 mismatch...

pgrinaway:
https://travis-ci.org/choderalab/perses/jobs/119378544#L178

jchodera:
Switch the devtools/install.sh to use Miniconda2-latest instead of
Miniconda-latest

pgrinaway:
ok

pgrinaway:
Nice, the `ValenceSmallMoleculeLibraryTestSystem` actually passes now!

pgrinaway:
Travis finally passes! Merging.

jchodera:
Woohoo!

PR #187, opened 2016-03-30
[WIP] Protect print statements with self.verbose check

jchodera:
This wraps `print` statements in `if self.verbose:` checks.


jchodera:
Woohoo!  All tests pass!

PR #195, opened 2016-04-07
Fix proposals

pgrinaway:
This PR has a cleaned set of molecules, as well as:

1. Methods for removing disconnected molecules
2. Chooses the atom map based on the max number of atoms in the match (maybe would be better to use highest MCSS score? not sure).


pgrinaway:
YES fixed the merge thing

pgrinaway:
Anyone have a better idea of what [this is](https://travis-ci.org/choderalab/perses/jobs/121494667#L1031)? I'm going to add the sterics in to this PR in the meantime.

jchodera:
@rafwiewiora had a PR merged into OpenMM that deals with overrides. Something must be broken with either that PR or the ResidueTemplateGenerator from openmoltools.

@rafwiewiora, can you investigate?

rafwiewiora:
Ok so this means you have a residue template named 'MOL' in the `ForceField` and you're trying to register another template with the same name - in the past the 'undocumented behavior' meant that you wouldn't have known about it and the last occurrence of the same-named residues passed to `ForceField` would have been used. Now we don't allow that, as passing a template with the same name means you wish to overload an already loaded template (which must have a higher `overloadLevel` than the previously loaded template). 

Can you guys point me to the place in the code where this is happening?

jchodera:
Is the overloading by name or by template bonds/elements?

rafwiewiora:
It's by name. It can't be by bonds/elements - you must allow loading of >1 templates that match the same on bonds/elements but have different names for different oxidations states of ions.

jchodera:
I'm trying to figure out how to handle this.

The residue is probably named `MOL` in perses because it's easier to deal with selecting it by residue name that way.

If we use a residue template generator, it currently just generates a template name based on the residue name.  Should we change that to a completely random unique ID?

jchodera:
OK, I've made a change to `openmoltools-dev` that should fix this behavior.  After the package rebuilds, I'll retrigger the tests.

rafwiewiora:
Is that the drug molecule residue that's having changes proposed and you're overriding the template in `ForceField`? Or do I have a completely wrong idea?

And for a residue template generator, you're saying there would be cases where you're registering a template X you created for residue X more than once? 

If you want to do any overloading on templates ever, then remember you can make use of the `overloadLevel` attribute now - that's of object `ForceField._TemplateData()`, set `template.overloadLevel` sequentially increasing if you're doing some sequence of overloading or to something guaranteed to be higher then current `overloadLevel` on that-named template. Default/starting level is 0.

jchodera:
> Is that the drug molecule residue that's having changes proposed and you're overriding the template in ForceField? Or do I have a completely wrong idea?

We're not intentionally trying to override anything. We're just trying to make sure that a template gets autogenerated by the forcefield template generator if there is no template found.

The molecule has the same residue name (`MOL`) in perses, but the actual molecular species changes over time, so the forcefield template generator keeps having to generate new residue templates. Previously, it just used the residue name, but since this was always `MOL`, this triggered your exception. Now, it appends a random string to the residue name, so I think it should work now.

Tests are re-running now!

rafwiewiora:
Ah I see! Ok, seems like a good solution!

jchodera:
OK, all fixed, @pgrinaway

jchodera:
Can this be merged, @pgrinaway?

pgrinaway:
Yep! looks good!

pgrinaway:
Thanks everyone!

Issue #192, opened 2016-04-03
how to equilibrate before NCMC computation?


sstolzenberg:
Hi there,

How can I properly equilibrate a system prior to an NCMC computation?

Trying to make this work for a simple test system (annihilating a proton in a water box), I came across some odd results in the computed NCMC potential energy differences (see [test.py.txt](https://github.com/choderalab/perses/files/201594/test.py.txt) and [test.log.txt](https://github.com/choderalab/perses/files/201595/test.log.txt), I can send you more input files if necessary):

When performing a "delete" NCMC computation after equilibration (via openmm.LangevinIntegrator), I obtain much larger potential energy differences (187kT in  test.log) compared to the same "delete" NCMC computation without any prior openmm-equilibration (-22kT in test.log).
The potential energy right after the equilibration and right before the NCMC computation are identical, and I have tried different equilibrated starting structures (in.pdb), and different equilibration/NCMC times.

Is there, e.g., a specific NCMC equilibration integrator I should be using instead of openmm.LangevinIntegrator?

Thank you for your help, Best,
Sebastian

jchodera:
Go ahead and post those input files and I'll try to hook you up with a working example!

sstolzenberg:
Hi,

OK, the solution for the above problem is to simply equilibrate my H-system for simulation times much longer than the inverse of my heat bath friction coefficient (1/ps) - I am able to do this now having access to cuda7.5.

Here is my updated test.py script including the input files:
[test.zip](https://github.com/choderalab/perses/files/207401/test.zip)

The reason I still post these files (and kindly ask you for help) is the following:
In test.py, you will see that I am running one equilibration and two non-equilibrium stages (alchemical transformations for NCMC) - "equ", "nonequ1", "nonequ2", each 1000 integration steps long. On average over 100 runs, each of these stages has the following runtime (with standard deviations):

dt_equ: 0.07+-0.03 min.
dt_nonequ1: 13.5+-4.2 min.
dt_nonequ2: 13.8+-5.4 min.

So for some reason, I can get the non-equilibrium transformations not even comparably as fast as the corresponding equilibration stage.
Can you qualitatively reproduce the above runtimes (differences)? Is there a tweak I can use to speed up the non-equilibrium transformations, or is this an hardware issue?

For what it's worth, I am using
Debian 3.16.7
perses-0.1-py2.7
openmm                    6.3.1
openmmtools               0.7.5
python 2.7.1

and a "GeForce GTX 980" GPU card

[...]

PR #196, opened 2016-04-08
Speed up NCMC switching by using loop inside CustomIntegrator

jchodera:
Attempt to address slow NCMC performance noted in #192 by using loops in `CustomIntegrator` instead of explicitly unrolling loop in number of steps.


jchodera:
All tests pass on a GPU machine as well.

PR #201, opened 2016-04-18
Updating complex.pdb in Abl Affinity test system

steven-albanese:
I updated the `complex.pdb` file used for the AblAffinityTestSystem. It had missing hydrogens and overlapping atoms. 

I fixed this using Ensembler to model ABL_HUMAN_D0 using `2HYY` PDB  as a template structure. I aligned this model to 2HYY to overlay imatinib in the correct place. I then manually dragged the ligand to resolve some overlap with sidechain atoms


steven-albanese:
Just added the script I'm using for the src and abl multitarget design script. Also put `src-abl` directory in the `data` directory

jchodera:
I think this should either go in `perses/tests/testsytems.py` as another test system---I think I would prefer this---or in `perses/examples/`, but we'd need to figure out a way to get the examples to be tested automatically if we go that route.

steven-albanese:
I wasn't sure that you wanted it to be a test system, happy to put it there. I should do that from a fork though, correct?

jchodera:
Hm. How about just moving it to `perses/examples/abl-src-selectivity/` or something for now?  We can figure out how to add extensive tests for examples later.

steven-albanese:
+1 will do

Issue #190, opened 2016-04-01
Improve readability of protein chemical_state_keys


juliebehr:
This isn't quite as straightforward as we were thinking, because I don't necessarily know what WT is.  It would be trivial, for instance, to make a ```new_chemical_state_key``` reflecting the changes made within that proposal with respect to the old sequence, but more difficult / less obvious to know what modifications have been made for ```old_chemical_state_key```.

jchodera:
You can change the API of `PointMutationEngine` to take whatever you want!  If it would be useful to cache a copy of the wild-type sequence, you can have the user pass in something (maybe a wild-type `Topology` object) in the constructor and extract the sequence and store it!

But `PointMutationEngine` only permits a fixed number of mutations *away* from the wild-type sequence, so I'm a little confused as to how you don't already have to track this information.

jchodera:
Might be useful to also improve the readability of small molecule chemical state keys if the molecules have a common name (e.g. `imatinib`) that can be provided along with the SMILES string.

Maybe this is a use of our `metadata` field?

juliebehr:
Currently planning on making the ```chemical_state_key``` in a point mutation system a list of strings describing the mutant -- implementing now.

jchodera:
+1!

pgrinaway:
Ok, I think this is done

PR #204, opened 2016-04-28
Modify chemical_state_keys from PointMutationEngine

juliebehr:
Addresses https://github.com/choderalab/perses/issues/190

The IP part is that I'm pretty sure I've altered the intended behavior of ```test_mutate_from_every_amino_to_every_other``` because I was previous taking advantage of the fact that I didn't actually go back to wildtype before making a new mutation, and now when I try to initiate my little protein with 20 different amino acids to start, it will actually only change 1 away from the wildtype (I think).

### DIFF
--- a/perses/rjmc/topology_proposal.py
+++ b/perses/rjmc/topology_proposal.py
@@ -268,7 +268,7 @@ class PolymerProposalEngine(ProposalEngine):
             atom_map = dict()
             for atom in modeller.topology.atoms():
                 atom_map[atom.index] = atom.index
-            print('PolymerProposalEngine: No changes to topology proposed, returning old system and topology')
+            if self.verbose: print('PolymerProposalEngine: No changes to topology proposed, returning old system and topology')
             topology_proposal = TopologyProposal(new_topology=old_topology, new_system=old_system, old_topology=old_topology, old_system=old_system, old_chemical_state_key=old_chemical_state_key, new_chemical_state_key=old_chemical_state_key, logp_proposal=0.0, new_to_old_atom_map=atom_map)
             return topology_proposal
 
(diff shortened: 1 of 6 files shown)


juliebehr:
Just kidding about the list of strings; lists aren't hashable.

jchodera:
I think immutable forms of list might be hashable. Is there a "frozenlist"
object?

andrrizzi:
I think the frozenlist is just a tuple.

juliebehr:
Is something wrong with a string?

juliebehr:
I had initially yesterday intended it to make (eg) ```['LYS99ALA', 'MET102GLN']``` and now that mutant's key would be ```'LYS99ALA-MET102GLN'```

jchodera:
Might be easier to analyze as a tuple (eg to see how many samples had MET102GLN in them) but the difference is probably minimal.

juliebehr:
```key.split('-')``` and it's a list again

juliebehr:
I just added the ability for the proposal to choose WT when allowed_mutations have been specified (previously it was only choosing from that list) and I expected this to break the tests where I count on being about to force it to make a certain mutation by making it the only allowed_mutations option but they seem to still run...

juliebehr:
Nope, I forgot to reinstall

juliebehr:
Ok, solved by finding a use for metadata, which I'm not sure if I'm crazy about -- ```metadata['always_change'] = True``` will block self-proposals, implemented only in ```PointMutationEngine```

jchodera:
`metadata` is meant to store information that is specific to each sample.  It's not meant to be a place where we add extra arguments.  If you want to add an `always_change` (or maybe `include_initial_sequence`?) argument to `PointMutationEngine`, why not just add it as an extra argument?  Those `ProposalEngine`s don't need to share the same constructors.

jchodera:
Should this get merged in, @juliebehr @pgrinaway ?

jchodera:
Should this get merged in @juliebehr @pgrinaway ?

juliebehr:
Yes

jchodera:
@pgrinaway: Can you review and merge when you get a chance?

PR #205, opened 2016-05-06
Fix VV and GHMC NCMC

jchodera:
Lots of fixes for VV and GHMC NCMC!


jchodera:
Wow, I wonder if I accidentally brought in some other changes.

jchodera:
I think there are a few more logic flow issues left over from reworking the integrators into something you have to call `nsteps` times.  Sorting those out now. Don't merge yet!

jchodera:
I've refactored the common code between `NCMCVVAlchemicalIntegrator` and `NCMCGHMCAlchemicalIntegrator` into an `NCMCAlchemicalIntegrator` base class and vastly simplified the code for the VV and GHMC NCMC integrators.  The NCMC tests pass on my machine.

Let's see if the tests pass here and then you can review and merge, @pgrinaway!

Wait...why is travis not running?

jchodera:
Ah. Travis isn't running because this is a PR into your branch, @pgrinaway

jchodera:
OK, so you'll need to review and merge and take things from here, @pgrinaway.

pgrinaway:
Thanks so much! will review and merge!

pgrinaway:
This looks great! Thanks a million!

PR #213, opened 2016-06-20
 Fix travis import hell

pgrinaway:
As the title says...


pgrinaway:
Build failure unrelated, issue added to trello. I'm going to rebuild the problematic ones now to see that this is actually working (it occasionally worked before).

jchodera:
Somehow, the other PR (removing rendering scripts) got mixed in here too.

pgrinaway:
I think we just missed a couple files on the last PR so I removed them here.

pgrinaway:
Ok, so it's still broken.

jchodera:
OK, I need to move those files into the new repo then.

pgrinaway:
Sounds good.

jchodera:
>  I need to move those files into the new repo then.

[Done](https://github.com/choderalab/persesmoviemaker/pull/1).

pgrinaway:
Anyone ever see [this before](this before)?

jchodera:
That link seems to be broken. What are you pointing to specifically?

pgrinaway:
Oops, copypasta fail. Here: https://travis-ci.org/choderalab/perses/jobs/139015837#L339

jchodera:
I think that's just a network failure. Restart the tests?

pgrinaway:
hm, restarted twice, still a problem...

jchodera:
Looks like the [syntax has changed](http://docs.eyesopen.com/toolkits/python/quickstart-python/) to:
```
pip install -i https://pypi.anaconda.org/OpenEye/simple OpenEye-toolkits
```
as of the latest release. Can you fix?

pgrinaway:
Sure! I wonder why some of them were working...

pgrinaway:
Huh, I wonder what is going on here? https://travis-ci.org/choderalab/perses/jobs/139046703#L451 ...maybe I messed up the nosetest installation...

pgrinaway:
Oh, huh: Error: HTTPError: 503 Server Error: Service Unavailable: Back-end server is at capacity for url

jchodera:
Ah, OK.

jchodera:
Maybe we should add a random sleep delay to the beginning of each travis branch so we don't hammer it with all of the test branches at the same time?

pgrinaway:
Is travis having network connectivity problems?

pgrinaway:
Maybe it's anaconda, actually.

jchodera:
Woohoo!  All tests pass!

PR #219, opened 2016-06-28
Propagation steps

juliebehr:
Creating the option in `NCMCEngine` to take multiple steps to propagate at each lambda step.

@pgrinaway

### DIFF
--- a/perses/annihilation/ncmc_switching.py
+++ b/perses/annihilation/ncmc_switching.py
@@ -15,6 +15,7 @@ default_functions = {
 default_temperature = 300.0*unit.kelvin
 default_nsteps = 1
 default_timestep = 1.0 * unit.femtoseconds
+default_steps_per_propagation = 1
 
 class NaNException(Exception):
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


jchodera:
Out of curiosity, what is the motivation here? Purely for debugging? I
think I can prove the thermodynamic length is always shorter if you take a
lambda step for every dynamics step.

pgrinaway:
> I think I can prove the thermodynamic length is always shorter if you take a
lambda step for every dynamics step.

Is this true for GHMC propagators? I could see an efficiency argument (or the length argument for VV), but why would the number of steps of GHMC increase the thermodynamic length?

pgrinaway:
Ok, thanks for the in-person explanation @jchodera . Seems that it's good to have this option but in general it will be more efficient to take smaller lambda steps than to do more propagation steps.

juliebehr:
Merge?

jchodera:
> Merge?

@pgrinaway: I think that's you!

bas-rustenburg:
@jchodera : Has that notebook link been moved? I'm getting a 404 page. Just curious since I use similar schemes and would like to take a look at it.

jchodera:
Looks like it has. Ask @gregoryross?

Issue #194, opened 2016-04-05
Have samplers store more data about simulation history


jchodera:
We should have the samplers store more information about the sampler history in a NetCDF file or something.

Ideas so far:
* current chemical state key
* proposed chemical state key
* log acceptance probability
* accept/reject statistics
* current `Topology` 
* current configuration and box volume
* log weights
* reduced potential energy

What other data should we keep track of?

pgrinaway:
potential energies?

jchodera:
Good idea! Let me update the issue with a list.

pgrinaway:
I will update this in a bit, but sort of spitballing here:

```python

class PersesOutput(object):

    def __init__(self, output_filename):
        self._filename = output_filename

    def write_configuration(self, positions, topology, reduced_potential=None, thermodynamic_state=None):
        pass

    def write_chemical_state(self, topology, chemical_state_key, reduced_potential=None, thermodynamic_state=None):
        pass

    def write_SAMS_weights(self, weights, target_weights, sams_sampler_name):
        pass

    def write_metadata(self, data, writer_name):
        pass

    def increment_iteration(self, sampler_name):
        pass

    def write_logp_accept(self, logp_accept, old_chemical_state, new_chemcial_state, accepted):
        pass

    def get_current_iteration(self, sampler_name):
        pass
```

I don't know if there should be anything more specific. I had imagined that `GeometryEngine` et al. could write out their specific debug info via the `write_metadata` method. The object would keep track of the current iterations, so that everything would be organized.

pgrinaway:
Also, I should add that I would change the base class constructors to also take a reference to a `PersesOutput` object, which would give each class an opportunity to insert whatever metadata it needs to.

steven-albanese:
+1 on the output class, this looks great

jchodera:
Can we generalize the API a bit? We may want to cram a ton of debug info in there, including logP for various substeps, configurations for various substeps or parts of substeps, timings of substeps, and other random information. If we need to extend the API for every little statistic we might need for debugging, it will hurt more than it helps to add this layer of abstraction.

What if we focus on a few general API commands to store some common types of per-iteration information, and have the API help us by creating the NetCDF variables under the hood if they don't already exist? Then we just need to standardize on the naming of various objects in the file that are important for analysis. 

We can also make use of object groups, which are like directories. Each module or Perses can use a separate group (directory) to keep its namespace isolated from others.

jchodera:
[...]

PR #220, opened 2016-06-30
Storage layer

jchodera:
Beginning implementation of #194


jchodera:
Conceptual question: How do we want to deal with the fact that there may be multiple sampler stacks inside a single `MultiTargetDesign` object? Do we want to have each sampler stack be organized inside of a separate dataset group within a single NetCDF file, or separate NetCDF files for each sampler stack (and another file for the `MultiTargetDesign` object)?

I'm guessing it's easiest to put everything in one file, but we need to figure out how to associate a unique name to each sampler stack (e.g. based on the environment name, like `complex-explicit`, `ligand-explicit`).

jchodera:
I've gone the route of putting everything in one file since nobody was around to provide feedback.

The NetCDF file is organized into groups, with variables named `/envname/modname/varname`, with `envname` the environment name (`complex-explicit`, `ligand-explicit`), `modname` the module class name (`MCMCSampler`, `ExpandedEnsembleSampler`, `SAMSSampler`), and `varname` the variable name (`logp_accept`, `logp_ncnc_annihilation`, etc.).

I've implemented writing of floats (acceptance probabilities and components) and objects (free energy and log weight `dict`s), and implemented these in all the samplers.

I'll think about how best to implement the writing of positions, but already this does what you need, so you can probably merge it and get going right away, @pgrinaway.

I've aggressively kept the [storage layer code](https://github.com/choderalab/perses/blob/855791ed2c2247946f0c5ede8fb4c7367cdcf0c6/perses/storage/storage.py) as simple and general as possible, but as always, feedback on how to further simplify things is greatly appreciated.

I haven't added code to read from the NetCDF file, since I imagine you might want to write your own analysis code. If you want me to add reader routines or a way to easily get at the data through `NetCDFStorage`, just let me know what you would like to see and I can add things ASAP.

jchodera:
Tests should pass now.

pgrinaway:
This looks great! Thanks! I would have gone with one output file as well--I think it's easier to manage.

jchodera:
Awesome.

Can I implement the remaining pieces in separate PRs? I think you can start
to use this as is right now.

Let me know if you need documentation on how to use it. See the
test_storage.py code for an example. Each sampler has a .storage attribute
so that you can control which samplers you want to write their data.

pgrinaway:
>Can I implement the remaining pieces in separate PRs?

Go for it.

>Let me know if you need documentation on how to use it. See the
test_storage.py code for an example. Each sampler has a .storage attribute
so that you can control which samplers you want to write their data.

Thanks!

PR #223, opened 2016-07-05
[WIP] Hybridtopology

juliebehr:
Current issue: `switch_logp` and `geometry_logp_reverse` are typically on the order of at least +/- billions


jchodera:
Thanks!

I think @juliebehr is reluctant to have this merged immediately, but my guess is that getting this merged and instrumenting it for the storage engine is going to be the most effective way to quickly debug what is going on here. 

I'll read through the changes and see if I can spot any errors first, but otherwise, would it be OK to merge late tonight and add storage hooks?

pgrinaway:
@jchodera I think you should be able to commit directly to that branch, since it's on the `choderalab` repo?

pgrinaway:
I think working on the `hybridtopology` branch might be best for now--we can merge into master when it's done.

perses/samplers/samplers.py
```diff
@@ -846,6 +851,136 @@ def update_state(self):
...
+                self.topology = topology_proposal.new_topology
+                self.sampler.sampler_state.positions = ncmc_new_positions
+                self.state_key = topology_proposal.new_chemical_state_key
+                self.naccepted += 1
+                if self.verbose: print("    accepted")
+            else:
+                self.nrejected += 1
+                if self.verbose: print("    rejected")
+            # Write to storage.
+            if self.storage:
```
juliebehr: Also this may need tweaked but is already here

jchodera:
> I think working on the hybridtopology branch might be best for now--we can merge into master when it's done.

Will do! Thanks for the explanation!

perses/annihilation/relative.py
```diff
@@ -676,16 +739,30 @@ def createPerturbedSystem(self):
             sys2_indices_in_system[atom2idx] = index
             topology.addAtom(name, element, residue)
 
-        for atom in topology.atoms():
-            if not atom.index in system_atoms.keys():
-                system_atoms[atom.index] = atom
+        # RENUMBERING OF ATOM.INDEX:
+        #     INDEX MUST MATCH POSITION IN TOPOLOGY.ATOMS()
+        #     FOR POSITIONS TO BE APPLIED PROPERLY
```
jchodera: Oooooh. I wonder how many other places I make that same error.

jchodera:
Just out of curiosity, the hybrid topology alchemical creation seems to have these `system.removeForce(0)` calls:
https://github.com/choderalab/perses/blob/f802ef015cdcb18ecfe5b0873793d7938eec05b1/perses/annihilation/relative.py#L195

Does this assume the force terms are always in some predetermined order?

jchodera:
> 	set lambda before comparing potentials -- bonds agree

Woohoo!

I had the same problem of `lambda` not being set correctly before comparing potentials in my own PR, and fixed that there. I worry that our branches may be pretty far out of sync now, and GitHub is complaining that your branch has too many conflicts to merge.

Maybe it's worth biting the bullet and pulling in the changes from `master` now so that you can continue debugging and resolving the conflicts at the same time?

jchodera:
Thanks!

How do I run just this test?

juliebehr:
Just  `python perses/test/test_hybrid_builder.py` , it should be self sufficient!  I just found a major bug though, which I'm fixing right now and will hopefully correct at least the angles.

jchodera:
Your guess was right, @juliebehr!  We're missing some torsions somehow:
```
molecule 1 torsions:
shared_torsions
[1, 6, 8, 10, 12, 13, 14, 15, 19, 20, 22, 23, 25, 28, 38, 39, 40, 41]
unique_torsions1
[4, 16, 17, 18, 29, 30, 31, 32, 33, 34, 37]

molecule 2 torsions:
[1, 6, 8, 10, 12, 13, 14, 15, 21, 22, 31, 32, 34, 37, 71, 72, 73, 74]
unique_torsions2
[4, 18, 19, 20, 25, 28, 29, 38, 39, 40, 41, 42, 43, 44, 45, 46, 47, 48, 49, 50, 51, 52, 55, 56, 57, 58, 59, 60, 61, 62, 63, 64, 65, 66, 67, 68, 69, 70]
```

jchodera:
Woohoo!

juliebehr:
Nonbondeds now agree at final state, not initial state.

jchodera:
Woo!

jchodera:
Cool! Does this fix the issue?

juliebehr:
No, these are the changes made Monday that result in the same energies as previously

juliebehr:
Update: this is working exactly as intended. Confirmed by flipping initial / final in the test.

jchodera:
Just to confirm, @juliebehr: You're fixing this up so it can be merged, right?

juliebehr:
This isn't ready to be merged yet, so no.  When all aspects of it have been tested, then yes.

jchodera:
> This isn't ready to be merged yet, so no. When all aspects of it have been tested, then yes.

Ah, OK. I must have misunderstood the "this is working exactly as intended" statement.

juliebehr:
Hybrid topology is doing exactly what I expect it to be doing. This is not the same as saying all aspects of this ncmc scheme are implemented as they ought to be.

jchodera:
Got it, and sorry for my confusion!

The accounting for the overall acceptance probability is likely broken for
both schemes (RJMC-NCMC-RJMC and NCMC-RJMC-NCMC), so we can probably fix
that specific aspect in a separate PR.

juliebehr:
screen shot 2016-08-17 at 3 22 28 pm

jchodera:
I can fix!

jchodera:
cc: https://github.com/choderalab/perses/pull/227 

Will merge that as soon as tests pass.

jchodera:
Merged #227. Go ahead and merge from master?

pgrinaway:
Let's not merge this until tests are passing. If it's really critical that this code be used, I can commit a fix to the `hybridtopology` branch, and the code in the branch could be used.

jchodera:
Would it be easy to do that?

pgrinaway:
>Would it be easy to do that?

Commit to the `hybridtopology` branch? Yes.

jchodera:
Great, let's go with that then. Thanks!

perses/tests/test_samplers.py
```diff
@@ -109,7 +109,39 @@ def test_samplers():
...
+    if 'TESTSYSTEMS' in os.environ:
+        testsystem_names = os.environ['TESTSYSTEMS'].split(' ')
+
+    for testsystem_name in testsystem_names:
+        import perses.tests.testsystems
+        testsystem_class = getattr(perses.tests.testsystems, testsystem_name)
+        # Instantiate test system.
+        testsystem = testsystem_class()
+        # Test MCMCSampler samplers.
+        testsystem.environments = ['vacuum']
```
jchodera: We want to change this line to find the environment(s) that begin with `vacuum`.

jchodera: Alternatively, we want to change this example to call the environment `vacuum` instead of `vacuum-imidazole`:
https://github.com/choderalab/perses/blob/master/perses/tests/testsystems.py#L1426-L1442

pgrinaway:
Nice, thanks @juliebehr

PR #227, opened 2016-08-17
Update openeye license

jchodera:
Fixes OpenEye license failures in travis.


jchodera:
OpenEye tools are openeyeing, so merging.

PR #228, opened 2016-08-19
Prevent use of sterics in GeometryEngine

pgrinaway:
This is pretty simple, but is just here to prevent accidental use of the (incomplete) sterics support (there are no exceptions, for instance). I have it raising an exception now, but if that's a bit excessive I can change it back. As long as tests pass on this and we agree it is appropriate for now, I will merge.

### DIFF
--- a/perses/rjmc/geometry.py
+++ b/perses/rjmc/geometry.py
@@ -1742,7 +1742,7 @@ class GeometrySystemGenerator(object):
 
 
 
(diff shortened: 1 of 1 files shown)


pgrinaway:
@jchodera is this ok with you (specifically for your purposes in generating movies)? It can be merged whenever if it is (or closed if it isn't).

jchodera:
Sorry, hadn't realized you needed my input!

Fine to merge if it doesn't hinder movie creation!

PR #231, opened 2016-08-23
Add writing of positions and topology to storage layer

jchodera:
This adds the writing of positions and topologies to the storage layer.

TODO:
- [ ] Add analysis tools to extract trajectories
- [ ] Add movie-making scripts

### DIFF
--- a/perses/rjmc/geometry.py
+++ b/perses/rjmc/geometry.py
@@ -228,6 +228,10 @@ class FFAllAngleGeometryEngine(GeometryEngine):
             from simtk.openmm.app import PDBFile
             prefix = '%s-%d-%s' % (self.pdb_filename_prefix, self.nproposed, direction)
             self._proposal_pdbfile = open("%s-proposal.pdb" % prefix, 'w') # PDB file for proposal probabilities
+            self._proposal_pdbfile.write('MODEL\n')
+            self._proposal_pdbfile.write('TER\n')
+            self._proposal_pdbfile.write('ENDMDL\n')
+
             if direction == 'forward':
                 pdbfile = open('%s-initial.pdb' % prefix, 'w')
                 PDBFile.writeFile(top_proposal.old_topology, old_positions, file=pdbfile)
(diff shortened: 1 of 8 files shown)


jchodera:
We could also store the [potential around each torsion drive for the geometry engine](https://github.com/choderalab/perses/blob/master/perses/rjmc/geometry.py#L857), and then render these using [PyMOL CGOs](http://pymol.sourceforge.net/newman/user/S0500cgo.html#14) to get a better idea of how to improve acceptance rates at the geometry stage.

pgrinaway:
What's the status of this PR?

jchodera:
Writing is all working. The only things that are missing are analysis tools and moviemaking scripts to facilitate doing stuff with the data, but that can certainly something that can be added in a separate PR.

pgrinaway:
Ok. I'll review this PR now. If you want me to merge it, I certainly can. Or, if you want to push more commits, feel free.

perses/rjmc/geometry.py
```diff
@@ -226,6 +227,11 @@ def _logp_propose(self, top_proposal, old_positions, beta, new_positions=None, d
             # DEBUG: Write growth stages
             from simtk.openmm.app import PDBFile
             prefix = '%s-%d-%s' % (self.pdb_filename_prefix, self.nproposed, direction)
+            self._proposal_pdbfile = open("%s-proposal.pdb" % prefix, 'w') # PDB file for proposal probabilities
```
pgrinaway: I'm curious if we can just write out the sequence of atoms in order. That way, to visualize the introduction, we'd only need to make them visible in the correct order.

jchodera: That's a great idea!

jchodera: I've been meaning to remove those PDB-writing blocks and replace them with NetCDF-writing blocks. Should we do that here or in a separate PR?

pgrinaway: Maybe we can tackle it in a separate PR, since this one is functional?

pgrinaway:
Going to merge this. Improvements can be made in a separate PR.

perses/rjmc/geometry.py
jchodera: Sure! Can you create an issue to remind us of what we wanted to add separately?

PR #234, opened 2016-08-31
Reinstate numba in setup.py because why was it removed

juliebehr:


jchodera:
Is numba actually installable via pip?

I also had some trouble with the openmm dependency in setup.py because when
a version constraint is used, pip doesn't recognize the conda installed
version.

setup.py
```diff
@@ -90,6 +90,6 @@ def find_package_data(data_root, package_root):
         'numexpr',
         'netcdf4',
         'seaborn',
-        #'numba'
+        'numba'
```
jchodera: @juliebehr : Because of https://github.com/choderalab/perses/pull/236#issuecomment-243991435 you will need to change this to `numba ==0.27.0` before merging, or else anyone installing via `pip` will run into the same [crazy `numba` errors](https://github.com/choderalab/perses/pull/236#issuecomment-243986305).

PR #239, opened 2016-09-03
Geometry engine bugfixes

jchodera:
This PR fixes three important bugs in the geometry engine:
* The correct growth index was not always being used in the valence `System` because of rounding errors
* Ring-closing torsions were not being added with their appropriate growth index
* Hydrogen atoms were being grown before heavy atoms, causing rings to sometimes not close

### DIFF
--- a/perses/rjmc/geometry.py
+++ b/perses/rjmc/geometry.py
@@ -1792,12 +1792,6 @@ class GeometrySystemGenerator(object):
         modified_torsion_force.addPerTorsionParameter("growth_idx")
         modified_torsion_force.addGlobalParameter(parameter_name, 0)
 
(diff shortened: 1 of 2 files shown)


jchodera:
Geometries look great now, except for intramolecular steric clashes, which I think can be resolved by turning on sterics and exceptions. I think this problem is solved.

jchodera:
I've also added:
* Exclusions when sterics are enabled (which should fix the high-energies and `NaN` issues we were seeing)
* (1,4) exceptions are always included in valence terms

The one other thing we may want to do is to change the simple binary `use_sterics` option to one that allows us to select either intramolecular sterics only (which should be super-fast) or all sterics (which might help increase acceptance rates at the cost of being slower).

jchodera:
I'm going to go ahead and merge this, since tests pass and bugs have been fixed.

PR #243, opened 2016-09-12
Fix travis so that proper Python version is tested

jchodera:
This fixes an issue in which only python 2.7 was being tested.


jchodera:
Can you review and merge @pgrinaway?

pgrinaway:
Will do! Sorry--missed this earlier.

pgrinaway:
I'm not going to wait for the landscape build. I will look into other CI solutions for test coverage, since this one seems so slow.

PR #245, opened 2016-09-19
Ensure travis tests actually run on correct python version

jchodera:
This PR fixes a bug where all travis testing was done with python 2.7, despite the value assigned to the `python` environment variable.

Even worse, it looks like sometimes the anaconda cloud version of `perses-dev` was being tested instead of the current source version, which could cause lots of confusion and lost debugging time.


jchodera:
Whoops---I already did this in https://github.com/choderalab/perses/pull/243

Issue #244, opened 2016-09-16
PointMutationEngine insists on self-proposals


pgrinaway:
I'm not sure if I'm just misreading the API or there is a bug, but when I do:

```python
        allowed_mutations = [[('2','PHE')]]
        proposal_metadata = {"always_change":True}
        for environment in environments:
            proposal_engines[environment] = PointMutationEngine(topologies[environment],system_generators[environment], chain_id, proposal_metadata=proposal_metadata, allowed_mutations=allowed_mutations)
```

I still get things like this:
```
Updating chemical state with ncmc-geometry-ncmc scheme...
Proposing new topology...
Proposed transformation: ALA2PHE => ALA2PHE
```

pgrinaway:
Update: I think this is a bug where any proposal from a mutation to wildtype automatically becomes a self-proposal. That is, once we leave wildtype, we can never propose to return.

juliebehr:
Addressed in https://github.com/choderalab/perses/pull/247 for cases with `allowed_mutations`.  Not exactly a bug, per se, in cases without, but probably doesn't have the intended behavior (won't go back to WT).  Changing now and will open a new PR.

juliebehr:
Addressed in https://github.com/choderalab/perses/pull/250 for other cases.

pgrinaway:
Ok, Closing.

PR #250, opened 2016-09-20
Correct behavior for always_change

juliebehr:
Addressing bug https://github.com/choderalab/perses/issues/244 for cases where `allowed_mutations` have not been specified.  Previously, would have prevented moves back to WT, after a move away from WT has been accepted.  Now, requires every mutation from the previous state to be different.

For instance, if your WT sequence is:
`1MET-2TYR-3ILE-4PHE-5SER` and `max_allowed_mutants = 2`
the first proposal will choose 2 locations at random, and change them to anything other than the WT residues, possibly:
```
locations = [2, 5]
1MET-2GLN-3ILE-4PHE-5LEU
chemical_state_key = 'TYR2GLN-SER5LEU'
```

The next proposal first chooses 2 locations at random, then changes those residues to anything other than what they were previously (therefore allowing a return to WT in the event that the same locations are chosen again). Some possibilities:
```
locations = [2, 5]
1MET-2TYR-3ILE-4PHE-5SER
chemical_state_key = 'WT'
```
OR
```
locations = [2, 5]
1MET-2PRO-3ILE-4PHE-5HIS
chemical_state_key = 'TYR2PRO-SER5HIS'
```
OR
```
locations = [2, 3]
1MET-2TYR-3ALA-4PHE-5SER
chemical_state_key = 'ILE3ALA'
```
But NOT a possibility would be:
```
locations = [2, 3]
1MET-2GLN-3ALA-4PHE-5SER
chemical_state_key = 'TYR2GLN-ILE3ALA'
```
because the `TYR2GLN` mutant from the previous iteration has to change.

### DIFF
--- a/perses/rjmc/topology_proposal.py
+++ b/perses/rjmc/topology_proposal.py
@@ -701,6 +701,13 @@ class PointMutationEngine(PolymerProposalEngine):
             ]
     always_change : Boolean -- OPTIONAL, default True
         Have the proposal engine always propose another mutation
+        If allowed_mutations is not specified, always_change will require ALL
+        point mutations to be different
+        The proposal engine will choose number of locations specified by 
+        max_point_mutants, and will require all of those residues to change
+        eg: if old topology included L99A and M102Q, the new proposal cannot
+            include L99A OR M102Q
+        (This is only relevant in cases where max_point_mutants > 1)
     """
 
     def __init__(self, wildtype_topology, system_generator, chain_id, proposal_metadata=None, max_point_mutants=None, residues_allowed_to_mutate=None, allowed_mutations=None, verbose=False, always_change=True):
@@ -828,6 +835,10 @@ class PointMutationEngine(PolymerProposalEngine):
         ---------
         modeller : simtk.openmm.app.Modeller
         chain_id : str
+        index_to_new_residues : dict
+            contains information to mutate back to WT as starting point for new mutants
+        old_key : str
+            chemical_state_key for old topology
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


juliebehr:
(failing tests are the same ones in master)

jchodera:
This looks great!

If `max_allowed_mutants = 2`, will that *always* propose two mutations per proposal? If so, what would you think about allowing the number of mutations per proposal to be set independently of the total number of mutations away from wild-type that are allowed? We may, for example, only want to propose one mutation at a time (to get high acceptance rates), but still allow up to two mutations away from wild-type.

The failing test case is hopefully fixed in #253.

juliebehr:
Yes, it will always propose 2 mutations (except in a case like the final example where one of the changes actually moves back to WT).  I had planned on doing one at a time with 2 separate runs using `max_allowed_mutants = 1` to ever actually look for double mutants.

jchodera:
>  I had planned on doing one at a time with 2 separate runs using max_allowed_mutants = 1 to ever actually look for double mutants.

Do you mean do one run with `max_allowed_mutants = 1`, record any mutants worth looking at, and then another run with `max_allowed_mutants = 1` and a different wild-type sequence with that mutant modeled in?

juliebehr:
Yes

jchodera:
Wouldn't it just be easier to separately be able to specify `max_allowed_mutants` and `mutations_per_proposal`?

juliebehr:
Sure, I'm just saying that's why it isn't like that already.  It doesn't really seem urgent.

jchodera:
Oh, no, not urgent unless we need it for Myb. Maybe just mark it as an issue?

@pgrinaway : This is ready for you to review, I think!

pgrinaway:
Ok, going to review now!

pgrinaway:
Ok, I think this is good to merge then.

PR #252, opened 2016-09-20
Fix bond constraint reverse logp

pgrinaway:
This allows `GeometryEngine` to properly handle bond constraints while calculating the `logp_reverse`. We assume that constraints have been satisfied.

### DIFF
--- a/perses/rjmc/geometry.py
+++ b/perses/rjmc/geometry.py
@@ -293,6 +293,8 @@ class FFAllAngleGeometryEngine(GeometryEngine):
             else:
                 if direction == 'forward':
                     constraint = self._get_bond_constraint(atom, bond_atom, top_proposal.new_system)
+                    if constraint is None:
+                        raise ValueError("Structure contains a topological bond [%s - %s] with no constraint or bond information." % (str(atom), str(bond_atom)))
                     r = constraint #set bond length to exactly constraint
                 logp_r = 0.0
 
--- a/perses/rjmc/geometry.py
+++ b/perses/rjmc/geometry.py
@@ -291,8 +291,9 @@ class FFAllAngleGeometryEngine(GeometryEngine):
                 logZ_r = np.log((np.sqrt(2*np.pi)*(sigma_r/units.angstroms))) # CHECK DOMAIN AND UNITS
                 logp_r = self._bond_logq(r, bond, beta) - logZ_r
             else:
-                constraint = self._get_bond_constraint(atom, bond_atom, top_proposal.new_system)
-                r = constraint #set bond length to exactly constraint
+                if direction == 'forward':
+                    constraint = self._get_bond_constraint(atom, bond_atom, top_proposal.new_system)
+                    r = constraint #set bond length to exactly constraint
                 logp_r = 0.0
 
             #propose an angle and calculate its probability
--- a/perses/tests/testsystems.py
+++ b/perses/tests/testsystems.py
@@ -307,7 +307,7 @@ class AlanineDipeptideValenceTestSystem(PersesTestSystem):
         from pkg_resources import resource_filename
         valence_xml_filename = resource_filename('perses', 'data/amber99sbildn-valence-only.xml')
         system_generators['vacuum'] = SystemGenerator([valence_xml_filename],
-            forcefield_kwargs={ 'nonbondedMethod' : app.NoCutoff, 'implicitSolvent' : None, 'constraints' : None },
+            forcefield_kwargs={ 'nonbondedMethod' : app.NoCutoff, 'implicitSolvent' : None, 'constraints' : app.HBonds },
             use_antechamber=False)
 
         # Create peptide in solvent.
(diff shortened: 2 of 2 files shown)


perses/rjmc/geometry.py
```diff
@@ -291,8 +291,9 @@ def _logp_propose(self, top_proposal, old_positions, beta, new_positions=None, d
                 logZ_r = np.log((np.sqrt(2*np.pi)*(sigma_r/units.angstroms))) # CHECK DOMAIN AND UNITS
                 logp_r = self._bond_logq(r, bond, beta) - logZ_r
             else:
-                constraint = self._get_bond_constraint(atom, bond_atom, top_proposal.new_system)
-                r = constraint #set bond length to exactly constraint
+                if direction == 'forward':
+                    constraint = self._get_bond_constraint(atom, bond_atom, top_proposal.new_system)
```
bas-rustenburg: What happens if said constraint does not exist?  And is `r` used anywhere, even if there is no constraint?

perses/tests/testsystems.py
```diff
@@ -307,7 +307,7 @@ def __init__(self, **kwargs):
         from pkg_resources import resource_filename
         valence_xml_filename = resource_filename('perses', 'data/amber99sbildn-valence-only.xml')
         system_generators['vacuum'] = SystemGenerator([valence_xml_filename],
-            forcefield_kwargs={ 'nonbondedMethod' : app.NoCutoff, 'implicitSolvent' : None, 'constraints' : None },
+            forcefield_kwargs={ 'nonbondedMethod' : app.NoCutoff, 'implicitSolvent' : None, 'constraints' : app.HBonds },
```
bas-rustenburg: Is this something you also plan to add to other test systems? 

Also, you probably want to make sure your test still runs when you're NOT using constraints.

perses/rjmc/geometry.py
pgrinaway: Well, that code is only reachable if there is a bond but no bond information (which is how `ParmEd` deals with bond constraints). If there's a bond, but neither parameters nor a constraint, the system is not going to work.

`r` is the bond length, which can either come from sampling a distribution defined by the bond parameters, or the constraint. But either way, it's necessary for the rest.

perses/tests/testsystems.py
pgrinaway: >Is this something you also plan to add to other test systems?

Yes. Consistency would probably be good.

>Also, you probably want to make sure your test still runs when you're NOT using constraints.

Yep.

jchodera: Do you want to add `app.HBonds` constraints as default to the `SmallMoleculeLibraryTestSystem` base class? You could make it an option to the constructor so that you can have some test systems both be with and without constraints.

perses/rjmc/geometry.py
bas-rustenburg: I cant guess from the code what exactly `_get_bond_constraint` would return. Should you maybe still throw a common sense exception, in case the constraint is also None?

bas-rustenburg: I'm mostly worried about bugs slipping through in case a bond as well as constraint are absent. The system might not work, but you could catch it earlier or later depending on where you have proper checks for it.

pgrinaway:
I wasn't sure whether I wanted it back before merging. We might still want to test on the constraint-less systems.

We need more tests in general, but these are on the docket.

perses/tests/testsystems.py
pgrinaway: >Do you want to add app.HBonds constraints as default to the SmallMoleculeLibraryTestSystem base class? 

Ok

bas-rustenburg:
Rolling back is good. Just know that your handling of constraints isn't complete without tests that include/validate constraint usage.

pgrinaway:
>Just know that your handling of constraints isn't complete without tests that include/validate constraint usage.

Yeah, I agree. Test coverage is fairly poor at the moment. After shoring up some smaller issues I want to make a big push on this.

perses/rjmc/geometry.py
pgrinaway: >I cant guess from the code what exactly _get_bond_constraint would return. Should you maybe still throw a common sense exception, in case the constraint is also None?

That could make sense. The system is pretty messed up if this is the case.

>I'm mostly worried about bugs slipping through in case a bond as well as constraint are absent.

If a topological bond is present, but there are no bond parameters or constraints, then yeah, that should fail. But it should be more graceful.

perses/rjmc/geometry.py
```diff
@@ -293,6 +293,8 @@ def _logp_propose(self, top_proposal, old_positions, beta, new_positions=None, d
             else:
                 if direction == 'forward':
                     constraint = self._get_bond_constraint(atom, bond_atom, top_proposal.new_system)
+                    if constraint is None:
```
bas-rustenburg: Maybe also add a case "`if type(constraint) == float/float64`" or something and an else case.

bas-rustenburg: If only python had static typing.. 😉

pgrinaway: I don't think that should be necessary. I don't see the need to type check the return value of every function.

bas-rustenburg: ok. 👍

pgrinaway:
Ok, the only errors remaining are preexisting in `master`. I'm going to merge.

PR #253, opened 2016-09-20
Fix object storage for python 3

### LINKED ISSUE #249: Storage layer test failures
I'm seeing failures like [this](https://travis-ci.org/choderalab/perses/jobs/161324417#L859) relating to the storage layer. @jchodera suggested in #247 that this might be caused by some pickling issues.

jchodera:
Fixes #249

### DIFF
--- a/perses/samplers/samplers.py
+++ b/perses/samplers/samplers.py
@@ -1356,6 +1356,7 @@ class SAMSSampler(object):
             print("SAMS sampler iteration %5d" % self.iteration)
         self.update_sampler()
         self.update_logZ_estimates()
+        if self.storage: self.storage.sync()
         self.iteration += 1
         if self.verbose:
             print("=" * 80)
@@ -1488,6 +1489,7 @@ class MultiTargetDesign(object):
         self.update_samplers()
         self.update_target_probabilities()
         self.iteration += 1
+        if self.storage: self.storage.sync()
         if self.verbose:
             print("*" * 80)
 
@@ -1589,6 +1591,7 @@ class ProtonationStateSampler(object):
             print("ProtonationStateSampler iteration %8d" % self.iteration)
         self.update_samplers()
         self.update_target_probabilities()
+        if self.storage: self.storage.sync()
         self.iteration += 1
         if self.verbose:
             print("*" * 80)
--- a/perses/storage/storage.py
+++ b/perses/storage/storage.py
@@ -25,6 +25,7 @@ import pickle
 import json
 import mdtraj
 from simtk import unit
+import codecs
 
 ################################################################################
 # LOGGER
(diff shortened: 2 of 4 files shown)


pgrinaway:
Huh, I wonder what [this is](https://travis-ci.org/choderalab/perses/jobs/161428202#L848)

pgrinaway:
Apparently a potentially [similar issue](https://github.com/tensorflow/tensorflow/issues/830) plagued tensorflow.

pgrinaway:
[Related](https://github.com/choderalab/openmoltools/pull/205)

pgrinaway:
I'm looking into fixing this

jchodera:
Thanks! Looks like the cause and solution is explained in https://github.com/swig/swig/pull/560

pgrinaway:
Seems from the `openmoltools` issue that the problem is creating the `simtk.openmm.System` object inside the generator was the culprit.

jchodera:
Let me fix.

jchodera:
Wait, we don't create a `System` object inside a generator in `perses.tests.test_storage`, do we?

pgrinaway:
No, I'm not sure what's going on here. But I can say that this function:
```python
def test_storage_hopefully_works():
    testsystem_name = 'ValenceSmallMoleculeLibraryTestSystem'
    niterations = 5 # number of iterations to run

    # Create storage.
    tmpfile = tempfile.NamedTemporaryFile()
    filename = tmpfile.name

    import perses.tests.testsystems
    # Instantiate test system.
    testsystem = perses.tests.testsystems.ValenceSmallMoleculeLibraryTestSystem(storage_filename=filename)

    # Test MCMCSampler samplers.
    #for environment in testsystem.environments:
    environment = 'vacuum'
    mcmc_sampler = testsystem.mcmc_samplers[environment]
    mcmc_sampler.verbose = False
    f = partial(run_sampler, mcmc_sampler, niterations)
    f.description = "Testing MCMCSampler for %s with environment %s" % (testsystem_name, environment)
    return f
```
works, whereas if we replace the last line with `yield f` it gives the familiar error:

```
StopIteration

During handling of the above exception, another exception occurred:

  File "/Users/grinawap/PycharmProjects/perses/perses/tests/test_storage.py", line 195, in <module>
    for t in test_storage_hopefully_works():
SystemError: <built-in function delete_System> returned a result with an error set
```

pgrinaway:
Seems like when the `StopIteration` exception is raised by the generator (standard behavior), a `SystemError` is being thrown. But I have no idea why, or what that built-in function is.

pgrinaway:
In fact, just yielding the `mcmc_sampler` causes the error:

```python
def test_storage_hopefully_works():
    testsystem_name = 'ValenceSmallMoleculeLibraryTestSystem'
    niterations = 5 # number of iterations to run

    # Create storage.
    tmpfile = tempfile.NamedTemporaryFile()
    filename = tmpfile.name

    import perses.tests.testsystems
    # Instantiate test system.
    testsystem = perses.tests.testsystems.ValenceSmallMoleculeLibraryTestSystem(storage_filename=filename)

    # Test MCMCSampler samplers.
    #for environment in testsystem.environments:
    environment = 'vacuum'
    mcmc_sampler = testsystem.mcmc_samplers[environment]
    mcmc_sampler.verbose = False
    #f = partial(mcmc_sampler.run, niterations)
    #f.description = "Testing MCMCSampler for %s with environment %s" % (testsystem_name, environment)
    yield mcmc_sampler


if __name__ == "__main__":
    for t in test_storage_hopefully_works():
        pass
```

pgrinaway:
Yielding the `testsystem` does not.

pgrinaway:
This fails:
```python
def thermostate_generator():
    for i in range(2):
        system = openmm.System()
        yield ThermodynamicState(system, 30.0*unit.kelvin)

if __name__ == "__main__":
    for t in test_storage_hopefully_works():
        pass
```

pgrinaway:
@bas-rustenburg [points out](https://github.com/swig/swig/blob/master/CHANGES#L281) that the problem may lie in the version of SWIG.

bas-rustenburg:
Yeah, its what John pointed out in swig. It just hasn't been included yet in the version of SWIG that OpenMM is built against.

bas-rustenburg:
My suggested solution for now, stop using the generator concept here until OpenMM has updated it's SWIG version.

jchodera:
What version of `swig` do we need?

I think we've [pinned the `openmm` recipe to `swig ==3.0.7`](https://github.com/omnia-md/conda-recipes/blob/master/openmm/meta.yaml#L26-L27).

jchodera:
Looks like we need `swig` 3.0.8 or later.

pgrinaway:
Might be some way around it? Interestingly this code passes:
```python
def thermostate_generator():
    for i in range(2):
        yield ThermodynamicState(openmm.System(), 30.0*unit.kelvin)
```

while this fails:
```python
def thermostate_generator():
    for i in range(2):
        system = openmm.System()
        yield ThermodynamicState(system, 30.0*unit.kelvin)
```

jchodera:
OK, a dumb fix is in.

jchodera:
OK, all tests pass.

PR #255, opened 2016-09-21
Change numba dependency to deal with anaconda cloud updates

### LINKED ISSUE #254: numba
We have [`numba` pinned to 0.28.0](https://github.com/choderalab/perses/blob/master/setup.py#L95), but anaconda [has dumped that version](https://anaconda.org/anaconda/numba/files) to replace it with 0.28.1.

jchodera:
Fixes #254


jchodera:
I don't believe this is needed.

PR #257, opened 2016-09-21
Add warning notice to readme

pgrinaway:
This just adds a warning notice to the readme for all users who want to play around with `Perses`. I put it in PR form to make sure everyone was ok with it.


jchodera:
+1

Issue #149, opened 2016-02-23
Add perses-dev builds for multiple platforms


jchodera:
Just a reminder for myself.

pgrinaway:
What's the timeframe for this?

jchodera:
Oh, I can do that now.

jchodera:
I'm deprecating the `perses-dev` packages and going for the multiplatform nightlies:
https://github.com/omnia-md/conda-dev-recipes/pull/39

jchodera:
Not sure what's going on here:
https://travis-ci.org/omnia-md/conda-dev-recipes/jobs/161515672#L506

jchodera:
@pgrinaway : Looks like I won't be able to build a `perses` `dev` conda package while you still have `numba` pinned to `0.27.0` because there are no `numba` packages for `numpy` 1.11:
```
...Can't build /io/perses due to unsatisfiable dependencies:
The following specifications were found to be in conflict:
  - numba ==0.27.0 -> numpy 1.11*
  - numba ==0.27.0 -> python 2.7*|3.4*
  - python 3.5*
```
More info [here](https://travis-ci.org/omnia-md/conda-dev-recipes/jobs/162299149).

jchodera:
Done in https://github.com/omnia-md/conda-dev-recipes/pull/39

jchodera:
I've deleted the old `perses-dev` package to avoid confusion.

New `perses` package will appear in the `dev` channel [here](https://anaconda.org/omnia/perses/files), and will be installable via
```bash
conda install -c https://conda.anaconda.org/omnia/label/dev perses
```

jchodera:
Something is wrong with `conda-dev-build` now:
https://github.com/omnia-md/conda-dev-recipes/issues/44

jchodera:
Fixed.

jchodera:
[Hooray!](https://anaconda.org/omnia/perses/files)

PR #258, opened 2016-09-23
Un-pin numba

jchodera:
This hopefully resolves the issues with #149


jchodera:
@pgrinaway : Changing from 
```
@jit(float64[:](float64[:], float64[:], float64[:], float64[:]), nopython=True, nogil=True, cache=True)
```
to just
```
@jit
```
seems to have fixed the latest `numba`-compatibility issues. Is that OK, or will that cause huge performance penalties?

pgrinaway:
That could cause pretty big performance problems. The arguments to the decorator declare the types of the input and require (via `nopython` and `nogil`) that the python code be compiled completely to native code. I don't think we can get rid of those.

jchodera:
> That could cause pretty big performance problems. The arguments to the decorator declare the types of the input and require (via nopython and nogil) that the python code be compiled completely to native code. I don't think we can get rid of those.

The [`numba` docs](http://numba.pydata.org/numba-doc/0.17.0/user/jit.html) seemed to recommend just letting `numba` compile things lazily and figure out what it needs to do. I imagine it takes some time the first time you call it with specific arguments, but then it caches the compiled versions.

The only thing it doesn't seem that `numba` can figure out on its own is whether or not to [release the gil](http://numba.pydata.org/numba-doc/0.17.0/user/jit.html#nogil). I could try just adding `gil=True`?

Otherwise, I'm not sure how long it will take to figure out why the latest `numba` can't deal with the requirements you are placing on those functions in the decorators.

jchodera:
Alternatively, we can wait until continuum produces a version of `numba` that can satisfy your constraints and continue to live without a `conda` package until then?

bas-rustenburg:
Can you switch to a different version of numpy instead of numba?

pgrinaway:
>The numba docs seemed to recommend just letting numba compile things lazily and figure out what it needs to do. I imagine it takes some time the first time you call it with specific arguments, but then it caches the compiled versions.

Yeah, but at least when I wrote the code, comparing the performance with and without the extra arguments showed a considerable speedup with the arguments.


>The only thing it doesn't seem that numba can figure out on its own is whether or not to release the gil. I could try just adding gil=True?

You can also just try `nopython=True` , which might be good enough.

>Otherwise, I'm not sure how long it will take to figure out why the latest numba can't deal with the requirements you are placing on those functions in the decorators.

I will raise an issue on `numba` Monday. But otherwise, if `numba` continues to be a problem, this code can be rewritten in (efficient) Cython without too much trouble.

jchodera:
> You can also just try nopython=True , which might be good enough.

Sure.

> Can you switch to a different version of numpy instead of numba?

Could probably do that temporarily if needed.

jchodera:
OK, it seems that `nopython=True` is the source of the issues.

jchodera:
[This comment](http://stackoverflow.com/questions/25709918/remove-the-numba-lowering-loweringerror-internal-error) suggests unrolling the `:` slice assignment may help.

pgrinaway:
Well, it was dying on line 32, right? There's an array allocation there. Maybe `numba` has been compiled against some different library than the version of `numpy` that we're using? 

What's the disadvantage in following Bas's suggestion?

jchodera:
Line 32 is
```python
rotation_matrix[0, 1] = axis[0]*axis[1]*(1.0-cos_angle) - axis[2]*sin_angle
```

jchodera:
> What's the disadvantage in following Bas's suggestion?

Pin `numba` to 0.27.0 and `numpy` to 1.09? It's a hack, rather than a solution, and may cause havoc if you try to install `perses` alongside other tools, but will do that for now.

pgrinaway:
Oops, must have been looking at an outdate version.

> It's a hack, rather than a solution,

Well, it's a patch. I never intended for that to be permanent. We just need it to work now. If `numba` proves problematic because of `numpy` version changes like this, I will code that part up in Cython, as I said above. But if we can punt on that, I think it's best that we do.

jchodera:
Hm, specifying the type for the numpy matrix allocation seems to have fixed things...

pgrinaway:
Numba can be very fickle about types,

jchodera:
Looks like we're good! I'll merge, since this shouldn't have any performance degradation.

PR #260, opened 2016-09-27
Fix zero indexing for multiple chains

juliebehr:


jchodera:
What problem does this address, out of curiosity?

PR #267, opened 2016-10-15
Overhaul NCMC integrators

jchodera:
I think this should fix the NCMC issues we previously had with the symmetric protocol (which we had disabled before now). The integrator structure for both VVVR and GHMC have been overhauled, and protocol and shadow work are also separately accumulated now.

### DIFF
--- a/perses/annihilation/ncmc_switching.py
+++ b/perses/annihilation/ncmc_switching.py
@@ -994,9 +994,10 @@ class NCMCAlchemicalIntegrator(openmm.CustomIntegrator):
         Reset step counter and total work
         """
         self.setGlobalVariableByName("total_work", 0.0)
+        self.setGlobalVariableByName("protocol_work", 0.0)
+        self.setGlobalVariableByName("shadow_work", 0.0)
         if (self.nsteps > 0):
             self.setGlobalVariableByName("step", 0)
-            self.setGlobalVariableByName("pstep", 0)
             if self.has_statistics:
                 self.setGlobalVariableByName("naccept", 0)
                 self.setGlobalVariableByName("ntrials", 0)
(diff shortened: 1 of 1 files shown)


jchodera:
I've disabled the `addUpdateContextState()` since I think it is unnecessary and would be much more complex to incorporate this flawlessly.

jchodera:
This is ready to review and merge.

jchodera:
I should note that local testing of `test_elimination.py` show a test failure that does not show up on travis, but that I don't know how to fix:
```
ERROR: Failure: UnboundLocalError (local variable 'wt_chain' referenced before assignment)
  File "/cbio/jclab/home/chodera/miniconda/lib/python2.7/site-packages/nose/loader.py", line 251, in generate
    for test in g():
  File "/cbio/jclab/home/chodera/github/choderalab/perses/perses.choderalab/perses/tests/test_elimination.py", line 160, in test_alchemical_elimination_mutation
    topology_proposal = proposal_engine.propose(system, topology)
  File "/cbio/jclab/home/chodera/github/choderalab/perses/perses.choderalab/perses/rjmc/topology_proposal.py", line 251, in propose
    old_chemical_state_key = self.compute_state_key(old_topology)
  File "/cbio/jclab/home/chodera/github/choderalab/perses/perses.choderalab/perses/rjmc/topology_proposal.py", line 937, in compute_state_key
    for wt_res, res in zip(wt_chain._residues, chain._residues):
UnboundLocalError: local variable 'wt_chain' referenced before assignment
```

perses/annihilation/ncmc_switching.py
```diff
@@ -1216,20 +1268,22 @@ def __init__(self, temperature, system, functions, nsteps=0, steps_per_propagati
         gamma = collision_rate
 
         # NCMC variables
-        self.addGlobalVariable('lambda', 0.0) # initial total energy (kinetic + potential)
-        self.addGlobalVariable('total_work', 0.0) # initial total energy (kinetic + potential)
+        self.addGlobalVariable('lambda', 0.0) # parameter switched from 0 <--> 1 during course of integrating internal 'nsteps' of dynamics        self.addGlobalVariable('lambda', 0.0) # alchemical switching parameter
+        self.addGlobalVariable('total_work', 0.0) # cumulative total work in kT
+        self.addGlobalVariable('shadow_work', 0.0) # cumulative shadow work in kT
```
pgrinaway: Why is the distinction between protocol work and shadow work necessary here?

pgrinaway:
I think it is important to document why, exactly, this change fixes some problem.

perses/annihilation/ncmc_switching.py
jchodera: It is not necessary, but it is useful. For example, for the benzene <--> naphthalene system, we could analyze the effects of omitting the shadow work for VVVR. We could also quantitatively say how much GHMC helps the NCMC acceptance rate.

pgrinaway: Sounds good.

jchodera:
> I think it is important to document why, exactly, this change fixes some problem.

There were two contributing issues:
* OpenMM bugs in `CustomIntegrator` that have been fixed in the latest dev builds: https://github.com/pandegroup/openmm/pull/1630 https://github.com/pandegroup/openmm/pull/1603
* The logical flow of the original NCMC integrators was flawed and has been restructured

In playing with various integrator structures with @gregoryross in `openmm-constph`, it looks like there may still be issues with manually unrolled integrator structures, but any problems that may appear there do not appear in the flow-controlled version we use here.

Issue #225, opened 2016-07-06
Make sure we are handling NPT correctly with MonteCarloBarostat


jchodera:
It appears that we aren't correctly treating solvated systems by adding a `MonteCarloBarostat` to them. While we _don't_ want to include volume change in NCMC (though this might be OK when GHMC is used), we definitely want to add a barostat for solvated systems, since otherwise the default water densities can be awful, making NCMC much more difficult than it should be.

jchodera:
We have to remember to switch off the barostat during NCMC

jchodera:
It looks like we have a few options here:
* Augment `SystemGenerator` to add a `MonteCarloBarostat` to NPT systems. This will require passing a `ThermodynamicState` to the `SystemGenerator`, and would mean that the generated system is specific to a single temperature and pressure. We would need to remember to disable the barostat in other parts of the code (like NCMC).
* Only use a barostat in `MCMCSampler`, where we add a barostat, run some simulation, and then remove the barostat. This keeps us from having to worry about disabling the barostat elsewhere.

Would be useful to discuss best approach.

pgrinaway:
Resolved by previous PR.

PR #269, opened 2016-10-15
MCMC sampler now adds a barostat if pressure is specified

jchodera:
This is the simplest possibility from https://github.com/choderalab/perses/issues/225#issuecomment-253998035

### DIFF
--- a/perses/samplers/samplers.py
+++ b/perses/samplers/samplers.py
@@ -219,7 +219,7 @@ class SamplerState(object):
 
         return self
 
-    def createContext(self, integrator=None, platform=None):
+    def createContext(self, integrator=None, platform=None, thermodynamic_state=None):
         """
         Create an OpenMM Context object from the current sampler state.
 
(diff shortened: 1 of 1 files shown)


perses/samplers/samplers.py
```diff
@@ -230,6 +230,8 @@ def createContext(self, integrator=None, platform=None):
            If not specified, a VerletIntegrator with 1 fs timestep is created.
         platform : simtk.openmm.Platform, optional, default=None
            If specified, the Platform to use for context creation.
+        thermodynamic_state : ThermodynamicState, optional, default=None
+            If specified, a barostat will be added to periodic systems with a pressure defined.
```
pgrinaway: I think this part of the docstring is a bit confusing. It's not the case that if `ThermodynamicState` is specified, a barostat will be added, right (that would make no sense if pressure were not specified in `ThermodynamicState`, no?).

 I think the intention is that a barostat will be added if a pressure is specified in `ThermodynamicState`, but since that argument is optional, it's ambiguous. If you agree, I can update this docstring.

jchodera:
Fixed.

pgrinaway:
Thanks!

PR #270, opened 2016-10-17
Tests for numerical parts of FFAllAngleGeometryEngine

pgrinaway:
I decided to split up the PRs for this, since this one will already be quite large. In this PR are tests for numerical parts of the `GeometryEngine`, including logp calculation and proposals.


pgrinaway:
Looks like I've got a unit bug. One sec...

jchodera:
The major problem I see is that the test docstrings are not informative. For example:
```
   Test the generation of torsion angles
```
Yes, but *how*? By what principle does the test operate?
Since the code is primarily uncommented, it would take someone a while to even figure out what the code is doing (which also makes review of the PR difficult).

What do you think about, at minimum, documenting the principle by which the test works in the docstring?

pgrinaway:
Ok, I will add more descriptive docstrings and comments.

perses/tests/test_geometry_engine.py
```diff
@@ -26,14 +26,553 @@
...
+    angle_array = np.zeros(1000)
+    for i in range(1000):
+        proposed_angle_with_units = geometry_engine._propose_angle(angle_with_units, beta)
+        angle_array[i] = proposed_angle_with_units.value_in_unit(unit.radians)
+
+    #compare the sampled angles to a normal cdf with the appropriate parameters using
+    #the Kolomogorov-Smirnov test. The null hypothesis is that they are drawn from the same
+    #distribution (the test passes).
+    (dval, pval) = stats.kstest(angle_array,'norm', args=(theta0_without_units, sigma_without_units))
+    if pval < 0.05:
```
jchodera: I don't think we should not be using p-values of 0.05 in our significance testing. Even if everything is working correctly, each test will randomly fail 5% of the time, and as the number of these tests increases, the probability of encountering a test failure approaches unity, making the tests useless. 

I'd love to discuss the optimal testing strategy for these stochastic tests, but my strong preferences is to select a p-value that is highly significant. I generally prefer something like:
```
pval_base / (ntests * ncommits)
```
where `pval_base` is a highly significant p-value (not 0.05, much smaller), `ntests` is the number of statistical tests, and `ncommits` is the number of commits we expect in the lifetime of the project (say, 5000). This essentially applies a multiple hypothesis testing correction such that when a test failure occurs, we are highly confident something is actually wrong and we need to investigate what is wrong.

perses/rjmc/geometry.py
```diff
@@ -922,13 +877,96 @@ def _torsion_logp(self, growth_context, torsion, positions, r, theta, phi, beta,
...
+
+    def __init__(self):
+        raise NotImplementedError("This functionality is incomplete.")
+
+    def _torsion_logp(self, growth_context, torsion, positions, r, theta, phi, beta, n_divisions=360):
+        """
+        Use a spline fit to the torsion profile to get a logp
+        Parameters
+        ----------
+        growth_context
```
jchodera: These function docstrings are incomplete. Parameters should be documented in [numpy format](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#docstring-standard). I think this is important even for private functions, since there are multiple developers.

perses/tests/test_geometry_engine.py
```diff
@@ -26,14 +26,553 @@
...
+class FourAtomValenceTestSystem(GeometryTestSystem):
+    """
+    This testsystem has 4 particles, and the potential for a bond, angle, torsion term.
+    The particles are 0-1-2-3 atom-bond-angle-torsion
+    """
+
+    def __init__(self, bond=True, angle=True, torsion=True):
+        #make a simple topology
+        self._default_positions = unit.Quantity(np.zeros([4,3]), unit=unit.nanometer)
+        self._default_positions[0] = unit.Quantity(np.array([ 0.10557722 ,-1.10424644 ,-1.08578826]), unit=unit.nanometers)
```
jchodera: Where do these default positions come from?

perses/tests/test_geometry_engine.py
```diff
@@ -26,14 +26,553 @@
...
+            torsion_force.addTorsion(0, 1, 2, 3, self._default_torsion_periodicity, self._default_torsion_phase, self._default_torsion_k)
+
+        self._structure = parmed.openmm.load_topology(self._topology, self._system)
+        self._positions = self._default_positions
+        self._integrator = openmm.VerletIntegrator(1)
+        self._platform = openmm.Platform.getPlatformByName("Reference")
+        self._context = openmm.Context(self._system, self._integrator, self._platform)
+        self._context.setPositions(self._positions)
+
+    @property
```
jchodera: Should the accessible properties of the class also be documented in a docstring?

perses/tests/test_geometry_engine.py
```diff
@@ -26,14 +26,553 @@
...
+    (periodicity, phase, k) = testsystem.torsion_parameters
+    periodicity_difference = np.abs(periodicity - torsion_with_units.type.per)
+    phase_difference = np.abs(phase - torsion_with_units.type.phase)
+    force_difference = np.abs(k - torsion_with_units.type.phi_k)
+    if periodicity_difference > 1.0e-6 or phase_difference.value_in_unit(unit.radians) > 1.0e-6 or force_difference.value_in_unit(k.unit) > 1.0e-6:
+        raise Exception("Did not add units correctly to torsion.")
+
+def test_torsion_scan():
+    """
+    Test the generation of torsion angles
```
jchodera: Can you document the criteria for test pass/fail?

jchodera:
We should get this PR merged too when you get a chance, though we'll want to test both variants of the `GeometrySystemGenerator`.

jchodera:
Does this test both variants of GeometrySystemGenerator?

pgrinaway:
Not yet. I was cleaning up this PR to merge the tests that have been written, and was going to open another to take care of the `GeometrySystemGenerator` variants.

jchodera:
Sounds good!

jchodera:
Can this be merged now?

pgrinaway:
I'm going to close this, fix the remaining problems (insufficient documentation) then reopen and merge.

PR #271, opened 2016-10-25
updating to a working selectivity script

steven-albanese:
Fixed an API issue and also added in netCDF storage. This is ready to merge once someone looks at it!

### DIFF
--- a/examples/abl-src-selectivity/abl_src_affinity.py
+++ b/examples/abl-src-selectivity/abl_src_affinity.py
@@ -1,16 +1,11 @@
 from simtk import openmm, unit
 from simtk.openmm import app
 import os, os.path
-import sys, math
-import numpy as np
-from functools import partial
-from pkg_resources import resource_filename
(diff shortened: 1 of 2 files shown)


jchodera:
I wonder if closely related inhibitors may still work, whoever. For example, we might see if someone has tested a series of fluorinated variants of the same inhibitor.

steven-albanese:
@jchodera: do you have any insight into how to set the netCDF storage up? Based on other test systems, it looks like a different `NetCDFStorageView` gets instantiated for every environment and passed to `SmallMoleculeSetProposalEngine`. Is that necessary? Or is it sufficient to just pass a `storage_filename` to the test system like this: `system = Selectivity(storage_filename=storage_filename)`

steven-albanese:
Actually just talked to @pgrinaway and he was able to answer my question

jchodera:
Great! All the examples in testsystems.py should be useful examples of how
to set up the storage interface.

PR #272, opened 2016-11-01
Proof-of-concept implementation of faster GeometrySystemGenerator

jchodera:
@pgrinaway had noted that the energy computation of the special `CustomForce`-encoded `System` object generated by `GeometrySystemGenerator` was a rate-limiting step that would likely make particle filtering using our current approach of drawing from `p(torsion | bond, angle)` impractical, so I attempted a different scheme to try to speed this up.

Instead of using `CustomForce` forces, this proof-of-concept implementation uses a copy of all the standard forces, and uses `updateParticleInContext` to update the parameters in the active `Context` when the particle index is incremented. This implementation is still lacking the addition of the few ring-closure torsions and angles, but these can be easily added. The implementation can also likely be sped up further by precomputing the growth indices of all bonds, angles, torsions, and nonbonded particles and exceptions.

I refactored `GeometrySystemGenerator` in order to move the function that updates the growth index for the active context in this class. That allows us to switch between the standard `GeometrySystemGenerator` (`CustomForce` version) and the new `GeometrySystemGeneratorFast` class. We can probably change this to a flag or something if we want to make it even simpler to compare.

Preliminary tests indicate the new implementation is much faster. For `ValenceSmallMoleculeLibraryTestSystem`, the geometry step for forward proposals goes from 2.4 s to 0.7 s. Commenting out the energy calculation completely in `_torsion_log_pmf` takes 0.3 s, suggesting the energy calculation is now on par with all the other overhead for this simple valence system.

For the `KinaseInhibitorsTestSystem` in vacuum, forward geometry proposals went from 10.8 s forward / 13.7 s reverse to 3.6 s / 4.4 s.

I couldn't try naphthalene or other recent test systems because I started from the `master` branch and all of those other systems are in their own branches.

### DIFF
--- a/perses/rjmc/geometry.py
+++ b/perses/rjmc/geometry.py
@@ -253,6 +253,7 @@ class FFAllAngleGeometryEngine(GeometryEngine):
         platform = openmm.Platform.getPlatformByName(platform_name)
         integrator = openmm.VerletIntegrator(1*units.femtoseconds)
         context = openmm.Context(growth_system, integrator, platform)
+        context.setParameter(growth_parameter_name, atom_proposal_order.keys()[-1])
         debug = False
         if debug:
             context.setPositions(self._metadata['reference_positions'])
--- a/perses/rjmc/geometry.py
+++ b/perses/rjmc/geometry.py
@@ -217,7 +217,6 @@ class FFAllAngleGeometryEngine(GeometryEngine):
             atom_proposal_order, logp_choice = proposal_order_tool.determine_proposal_order(direction='reverse')
             structure = parmed.openmm.load_topology(top_proposal.old_topology, top_proposal.old_system)
             atoms_with_positions = [structure.atoms[atom_idx] for atom_idx in top_proposal.old_to_new_atom_map.keys()]
-            growth_system_generator = GeometrySystemGeneratorFast(top_proposal.old_system, atom_proposal_order.keys(),growth_parameter_name, reference_topology=top_proposal.old_topology)
             growth_system_generator = GeometrySystemGeneratorFast(top_proposal.old_system, atom_proposal_order.keys(), growth_parameter_name, reference_topology=top_proposal.old_topology, use_sterics=self.use_sterics)
             growth_system = growth_system_generator.get_modified_system()
         else:
--- a/perses/rjmc/geometry.py
+++ b/perses/rjmc/geometry.py
@@ -253,7 +253,7 @@ class FFAllAngleGeometryEngine(GeometryEngine):
         platform = openmm.Platform.getPlatformByName(platform_name)
         integrator = openmm.VerletIntegrator(1*units.femtoseconds)
         context = openmm.Context(growth_system, integrator, platform)
-        context.setParameter(growth_parameter_name, len(atom_proposal_order.keys())+1)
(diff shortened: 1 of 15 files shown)


jchodera:
Also, tests would need updating since #270 hasn't been merged.

jchodera:
To make this faster and more scalable, we could:
* pre-compile lists of all bonds, angles, torsions, and nonbonded particles/exceptions involved in each growth index increment, and only update those (instead of recomputing the growth index for every term each time)
* omit bonds, angles, and torsions that are not involved in particle growth from the `_reference_system` 
* reduce the cutoff or change the `nonbondedMethod` when `sterics` are in use

jchodera:
The travis failure is due to the same issue as https://github.com/choderalab/alchemy/pull/53/files

perses/rjmc/geometry.py
```diff
@@ -2149,6 +2162,115 @@ def _calculate_growth_idx(self, particle_indices, growth_indices):
...
+    def set_growth_parameter_index(self, growth_index, context=None):
+        """
+        Set the growth parameter index
+        """
+        for (growth_force, reference_force) in zip(self._growth_system.getForces(), self._reference_system.getForces()):
+            force_name = growth_force.__class__.__name__
+            if (force_name == 'HarmonicBondForce'):
+                for bond in range(reference_force.getNumBonds()):
+                    parameters = reference_force.getBondParameters(bond)
+                    this_growth_index = self._calculate_growth_idx(parameters[:2], self._growth_indices)
```
jchodera: You can precalculate these growth indices for each of the bonds, angles, torsions, particles, and exceptions in order to further speed this up.

If you're always incrementing the growth index by 1, you can also save time by only updating the parameters that are changing.

jchodera:
Looks like you have to rebase or merge from `master` to resolve some conflicts.

pgrinaway:
Yep, doing that now.

pgrinaway:
Doing that again...

pgrinaway:
I messed up when merging stuff I think. Will fix now.

perses/rjmc/geometry.py
```diff
@@ -2142,9 +2142,9 @@ def _determine_extra_angles(self, angle_force, reference_topology, growth_indice
                     growth_idx = self._calculate_growth_idx(atom_indices, growth_indices)
                     #If this is a CustomAngleForce, we need to pass the parameters as a list, and it will have the growth_idx parameter.
                     #If it's a regular HarmonicAngleForce, there is no growth_index and the parameters are passed separately.
-                    if type(angle_force) == openmm.CustomAngleForce:
+                    if isinstance(angle_force, openmm.CustomAngleForce):
                         angle_force.addAngle(atom_indices[0], atom_indices[1], atom_indices[2], [angle_radians, angle_force_constant, growth_idx])
-                    elif type(angle_force) == openmm.HarmonicAngleForce:
+                    elif isinstance(angle_force, openmm.HarmonicAngleForce)
```
jchodera: Forgot colon at the end of the line!

pgrinaway:
The API change was not fully accommodated (instead of `context.setParameter` we call a method of the `GeometrySystemGenerator` to handle the case where there is no `growth_index`)

pgrinaway:
Ok, I think this PR is good to merge.

Issue #277, opened 2016-11-16
Debugging ImidazoleProtonationStateTestSystem


jchodera:
I'm debugging errors with `ImidazoleProtonationStateTestSystem` nosetests that occur during `test_samplers.py:test_sampler`.

It appears that the vacuum environment run fine, but then the explicit environment tests fail with
```
======================================================================
ERROR: Testing SAMS sampler with ImidazoleProtonationStateTestSystem 'explicit-imidazole'
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/cbio/jclab/home/chodera/miniconda3/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/cbio/jclab/home/chodera/github/choderalab/perses/perses.choderalab/perses/samplers/samplers.py", line 1432, in run
    self.update()
  File "/cbio/jclab/home/chodera/github/choderalab/perses/perses.choderalab/perses/samplers/samplers.py", line 1415, in update
    self.update_sampler()
  File "/cbio/jclab/home/chodera/github/choderalab/perses/perses.choderalab/perses/samplers/samplers.py", line 1370, in update_sampler
    self.sampler.update()
  File "/cbio/jclab/home/chodera/github/choderalab/perses/perses.choderalab/perses/samplers/samplers.py", line 1235, in update
    self.update_state()
  File "/cbio/jclab/home/chodera/github/choderalab/perses/perses.choderalab/perses/samplers/samplers.py", line 890, in update_state
    topology_proposal = self.proposal_engine.propose(system, topology)
  File "/cbio/jclab/home/chodera/github/choderalab/perses/perses.choderalab/perses/rjmc/topology_proposal.py", line 1201, in propose
    current_mol_smiles)
  File "/cbio/jclab/home/chodera/github/choderalab/perses/perses.choderalab/perses/rjmc/topology_proposal.py", line 1520, in _propose_molecule
    current_smiles_idx = self._smiles_list.index(molecule_smiles)
ValueError: 'c1c[nH+]c[nH]1' is not in list

======================================================================
```
The proposal engine thinks that the current state (`c1c[nH+]c[nH]1`) is not in the `_smiles_list`, which contains
```
['c1c[nH][cH+][nH]1', 'c1cnc[nH]1']
```
[...]

PR #278, opened 2016-11-16
[WIP] Fix ImidazoleProtonationStateTestSystem test failures

jchodera:
As discussed in #277, the issue with the `ImidazoleProtonationStateTestSystem` was not that the test was flawed, but that `SmallMoleculeSetProposalEngine` does not property represent molecules differing only in protonation states. The use of canonical isomeric SMILES did not appear sufficient to unambiguously identify distinct protonation states of imidazole---or more precisely, the same protonation state was represented by multiple distinct canonical isomeric SMILES strings, which seems incorrect. Using canonical isomeric SMILES with explicit hydrogens appears to work, though I have only currently verified that the `ImidazoleProtonationStateTestSystem` sampling works. 

Will run the tests on travis to see if this breaks anything.

### DIFF
--- a/perses/rjmc/topology_proposal.py
+++ b/perses/rjmc/topology_proposal.py
@@ -1189,8 +1189,9 @@ class SmallMoleculeSetProposalEngine(ProposalEngine):
         -------
         proposal : TopologyProposal object
            topology proposal object
-        """
+        """        
         current_mol_smiles, current_mol = self._topology_to_smiles(current_topology)
+        print('current_mol_smiles = %s' % current_mol_smiles) # DEBUG
 
         current_receptor_topology = self._remove_small_molecule(current_topology)
         old_mol_start_index, len_old_mol = self._find_mol_start_index(current_topology)
@@ -1303,7 +1304,7 @@ class SmallMoleculeSetProposalEngine(ProposalEngine):
         Returns
         -------
         smiles_string : string
-            an isomeric canonicalized SMILES string representing the molecule
+            an isomeric canonicalized SMILES string with explicit hydrogens representing the molecule
         oemol : oechem.OEMol object
             molecule
         """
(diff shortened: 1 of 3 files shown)


jchodera:
Looks like more extensive changes will be needed to ensure that all input SMILES strings fed to the `SmallMoleculeSetProposalEngine` by all the tests are all consistently generated using
```python
isosmiles = oechem.OECreateSmiString(molecule, oechem.OESMILESFlag_DEFAULT | oechem.OESMILESFlag_Hydrogens)
```
@pgrinaway and @juliebehr :
* Before I make more extensive changes, do you think explicitly encoding hydrogens in the SMILES string is the right approach, or is there some other way to actually make the SMILES strings unique with implicit hydrogens while correctly capturing the equivalence between forms like `c1c[nH+]c[nH]1` and `c1c[nH][cH+][nH]1`?
* We generate canonical isomeric SMILES at multiple places in the code for the purpose of having a single unique representation of molecules. Perhaps we should reduce this to a single point in the code in case we ever want to change this, so that extensive changes are not required every time we tweak the small molecule representation? If so, where should that be?

jchodera:
As an alternative to using explicit hydrogens, I wonder if enforcing a standard normalization scheme for `OEMol` molecules prior to generating standard canonical isomeric SMILES would be sufficient to solve this problem.

Even if we go this route, it's probably still a good idea to centralize the function that goes from `OEMol` to our standard SMILES representation in one place.

pgrinaway:
>As discussed in #277, the issue with the ImidazoleProtonationStateTestSystem was not that the test was flawed, but that SmallMoleculeSetProposalEngine does not property represent molecules differing only in protonation states. 

That wasn't originally part of the design, which is why it doesn't represent those as different. I think we should focus on getting the T4 ligands to work before doing this.

pgrinaway:
That said, I do agree with centralizing the SMILES generation location eventually.

jchodera:
Correctly representing protonation states of amino acids and small molecules is certainly within the scope of the design, but as discussed in slack, we can punt on this to focus on the T4 ligands.

jchodera:
Closing this PR for now. Will disable the tests in the travis PR, and consolidate the protonation state issues.

PR #280, opened 2016-11-22
Fix bug in small molecule topology proposal

jchodera:
This fixes a major bug in the proposal of new small molecule `Topology` objects in which newly-constructed `Topology` objects contained `chain` objects that incorrectly pointed to the old `Topology` object because deep copy was used instead of purely constructive `Topology` generation via the public API.

I'm not actually sure how this could have worked before. I just ran into ParmEd stack traces when I tried to run the imidazole test in explicit solvent. Perhaps it only worked in vacuum, where there was only one residue?


jchodera:
Uh, I based this off of the `hybrid-topology` branch, so I guess it's trying to bring all of those changes in.

I'll open a PR to `hybrid-topology`, but we really want to bring those changes into `master` since they are critical.

PR #281, opened 2016-11-22
Fix bugs in small molecule topology proposal

jchodera:
This fixes a major bug in the proposal of new small molecule Topology objects in which newly-constructed Topology objects contained chain objects that incorrectly pointed to the old Topology object because deep copy was used instead of purely constructive Topology generation via the public API.

I'm not actually sure how this could have worked before. I just ran into ParmEd stack traces when I tried to run the imidazole test in explicit solvent. Perhaps it only worked in vacuum, where there was only one residue?

We should really get these fixes into `master` as well, but I'm merging into `hybrid-tests` since I started with that branch to try to get protonation state sampling to work.

### DIFF
--- a/perses/rjmc/topology_proposal.py
+++ b/perses/rjmc/topology_proposal.py
@@ -1355,11 +1355,41 @@ class SmallMoleculeSetProposalEngine(ProposalEngine):
         resname = self._residue_name
         mol_residues = [res for res in topology.residues() if res.name==resname]
         if len(mol_residues)!=1:
-            raise ValueError("There can only be one residue with a specific name in the topology.")
+            raise ValueError("There must be exactly one residue with a specific name in the topology. Found %d residues with name '%s'" % (len(mol_residues), resname))
         mol_residue = mol_residues[0]
         atoms = list(mol_residue.atoms())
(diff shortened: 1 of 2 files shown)


jchodera:
Someone else can back-port this into `master`. I can't figure out the branch madness.

pgrinaway:
I'm going to close this since I've cherry-picked on another branch to accomplish that task.

jchodera:
Which branch? And shouldn't these fixes also go into hybrid-tests?

Also, what did you cherry pick? The whole fix was essential.

pgrinaway:
>Which branch?

`small-molecule-bugfix`, see #281 

>And shouldn't these fixes also go into hybrid-tests?

They will when #281 is merged and then master is merged into `hybrid-tests`. Otherwise `small-molecule-bugfix` can be merged into `hybrid-tests` directly, but since we're about to merge #281 that seems pointless.

>Also, what did you cherry pick? 

The commit `fc2fbb7`. Cherry-picks work on a commit basis, and you indicated the fix was in one commit, so of course it contains the entire thing. If there are fixes in another commit, then we'll have to include those.

PR #282, opened 2016-11-22
Small molecule bugfix

pgrinaway:
This PR brings in the bugfix commit in `protonation`. It has one extra commit (see description) because of a mistake I made while resolving a conflict.


jchodera:
This will go I to master, and then everyone will update their other branches from there?

pgrinaway:
>This will go I to master, and then everyone will update their other branches from there?

Yes, I think that makes the most sense.

PR #283, opened 2016-11-23
Add trajectory retrieval and writing to storage layer

jchodera:
This also fixes some minor bugs in NCMC snapshot writing.


jchodera:
I've made some additional tweaks here:
* GHMC collision rate for NCMC increased 10x to 9.10/ps for better thermalization
* Annihilated or inserted atom bonds, angles, and torsions are only softened at intermediate lambda values for annihilation/insertion NCMC
* Default softening for `NCMCHybridEngine` is 1.0 (which does no softening)

jchodera:
@juliebehr @pgrinaway : This is ready to review and merge.

PR #284, opened 2016-11-28
[WIP] Numerous bugfixes

jchodera:
This merges in some changes from my `protonation` branch:
* SMILES strings are now canonicalized by converting them to OpenMM `Topology` objects and back to SMILES again. Without this, the protonation state of imidzole would end up with a different SMILES strings that wasn't in the original set. This is fixed for all testsystems.
* The proposal engines would silently remove any barostats because `SystemGenerator`had no facility to add these to newly generated systems. I've added a way to specify that `SystemGenerator` should add a barostat, and added this to all testsystems.
* `removeCMMotion` is now disabled by the system generator if not explicitly specified
* `HybridTopologyFactory` would raise a puzzling exception when the number of forces differed between `system1` and `system2` (because of the missing barostat). I've added a check during the constructor that prints useful debug information if this happens.
* Barostats are disabled during NCMC switching.
* `perses.analysis.plot_ncmc_work` now handles multiple environments. We can generalize this to still work even if environments are not present, but I haven't done this yet.

I haven't run extensive tests on these changes yet---I've been focusing on the imidazole protonation state test.

Despite these changes, I haven't been able to get the `geometry-ncmc-geometry` scheme to work with the imidazole protonation state test---the coordinates become `nan`. Still debugging why.

### DIFF
--- a/perses/analysis/analysis.py
+++ b/perses/analysis/analysis.py
@@ -147,6 +147,50 @@ class Analysis(object):
             pdf.savefig()
             plt.close()
 
+    def plot_sams_weights(self, filename):
+        """Generate plots of SAMS weights.
+
+        NOTE: This currently does not work since logZ and log_weights are stored as objects containing 'dict's.
+
+        Parameters
+        ----------
+        filename : str
+            File to write PDF of NCMC work plots to.
+
+        """
+        with PdfPages(filename) as pdf:
+            for envname in self.get_environments():
+                def plot_sams_weights(pdf, logZ, log_weights, title=""):
+                    plt.figure(figsize=(12, 8))
+
+                    #
+                    # plot logZ
+                    #
+                    plt.subplot(2,1,1)
+                    plt.plot(logZ, '.')
+                    plt.xlabel('iteration')
+                    plt.ylabel('logZ estimate')
+                    
+                    #
+                    # plot log_weights
+                    #
+                    plt.subplot(2,1,2)
+                    plt.plot(log_weights, '.')
+                    plt.xlabel('iteration')
+                    plt.ylabel('log weights')
+
+                    pdf.savefig()  # saves the current figure into a pdf page
+                    plt.close()
+
+                for modname in ['SAMSSampler']:
+                    varname = '/' + envname + '/' + modname + '/' + 'logZ'
+                    print(varname)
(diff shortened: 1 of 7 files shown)


jchodera:
Here's an example of the NCMC work traces using 250 steps of `ncmc-geometry-ncmc` switching. The work variances are still quite high, though it's expected we need something like 20 ps trajectories for these to work well.

ncmc-250

jchodera:
I'm not sure why travis tests aren't running in this PR.

jchodera:
Not sure why travis tests are not running here...

jchodera:
Please review and merge.

pgrinaway:
Can we keep constant-pH-related commits to their own branch? We agreed to wait until after the first paper to tackle this. It's fine to be in a branch so we get a running start, but it doesn't belong in master yet (whereas the commits in `hybrid-test` will be merged into master shortly)

jchodera:
We can talk about this at our brief meeting today, but these are bugfixes and not feature additions.

jchodera:
For example, you are going to need a barostat for the T4 paper. And I am surprised you haven't been bitten by the fact that all the ligands are not in canonical form for those ligands. Also, we should be removing CMMotionForce

jchodera:
If new features were accidentally swept in, I can cherry pick changes for a
PR into here or master.

perses/analysis/analysis.py
```diff
@@ -229,12 +261,24 @@ def plot_work_trajectories(pdf, work, title=""):
...
-
-                    # Plot work separated out for each chemical transformation
-                    #[niterations, nsteps] = work.shape
-                    #transformations = dict()
-                    #for iteration in range(niterations):
-                    #    plot_work_trajectories(pdf, work, title='(all transformations)')
+                for modname in ['NCMCEngine', 'NCMCHybridEngine']:
+                    work = dict()
+                    for direction in ['delete', 'insert']:
+                        varname = '/' + envname + '/' + modname + '/' + 'work_' + direction
```
juliebehr: Has this been run with the  benchmark? Because it shouldn't work, hence why I had `modname = envname`

jchodera:
Closed in favor of bugfixes consolidated into #290

PR #285, opened 2016-12-01
NCMC refactor -- cherry picking from hybrid-tests

juliebehr:
Extracting changes in the `hybrid-tests` branch that are independent to testing hybrid builder; here: refactoring `ncmc_switching.py` to inherit between `NCMCEngine` and `NCMCHybridEngine`

### DIFF
--- a/perses/annihilation/ncmc_switching.py
+++ b/perses/annihilation/ncmc_switching.py
@@ -158,29 +158,6 @@ class NCMCEngine(object):
                         parameters.append(parameter_name)
         return parameters
 
-    def _updateAlchemicalState(self, context, functions, value):
-        """
-        Update alchemical state using the specified lambda value.
-
-        Parameters
-        ----------
-        context : simtk.openmm.Context
-            The Context
-        functions : dict
-            A dictionary of functions
-        value : float
-            The alchemical lambda value
- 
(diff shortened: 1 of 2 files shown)


jchodera:
Yay, travis is running again!

perses/annihilation/ncmc_switching.py
```diff
@@ -538,52 +622,9 @@ def __init__(self, temperature=default_temperature, functions=None,
...
-        # Clean up context and integrator.
-        del context, integrator
-        # Return potential energy.
-        return -self.beta * potential
-
-    def _computeAlchemicalCorrection(self, unmodified_old_system,
-                                     unmodified_new_system, alchemical_system,
-                                     initial_positions, alchemical_positions,
-                                     final_hybrid_positions, final_positions,
+    def _computeAlchemicalCorrection(self, integrator, context,
```
jchodera: This `_computeAlchemicalCorrection` is different from the other one on line 161. Can they be combined, or do they need to be distinct?

juliebehr: They are completely different and need to be distinct

juliebehr:
I think this is good to go -- the tests pass with this as-is, and there are other changes from `hybrid-tests` I'd like to pull in subsequently.

jchodera:
Thanks!

PR #287, opened 2016-12-01
ExpandedEnsembleSampler refactor (cherry picking from hybrid-tests)

juliebehr:
Breaking the three schemes of `ExpandedEnsembleSampler` into sub-methods that can also share general components.


jchodera:
This is great, thanks!

PR #290, opened 2016-12-02
Various bugfixes

jchodera:
Bugfixes culled (by hand) from #284:
* SMILES strings are now canonicalized by converting them to OpenMM `Topology` objects and back to SMILES again. Without this, the protonation state of imidzole would end up with a different SMILES strings that wasn't in the original set. This is fixed for all testsystems.
* The proposal engines would silently remove any barostats because `SystemGenerator`had no facility to add these to newly generated systems. I've added a way to specify that `SystemGenerator` should add a barostat, and added this to all testsystems.
* `removeCMMotion` is now disabled by the system generator if not explicitly specified
* `HybridTopologyFactory` would raise a puzzling exception when the number of forces differed between `system1` and `system2` (because of the missing barostat). I've added a check during the constructor that prints useful debug information if this happens.
* Barostats are disabled during NCMC switching.

### DIFF
--- a/perses/annihilation/ncmc_switching.py
+++ b/perses/annihilation/ncmc_switching.py
@@ -170,7 +170,7 @@ class NCMCEngine(object):
         ----------
         itegrator : NCMCAlchemicalIntegrator subclasses
             NCMC switching integrator to annihilate or introduce particles alchemically.
-        context : openmm.Context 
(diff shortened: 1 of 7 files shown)


jchodera:
Ready for review.

PR #294, opened 2016-12-07
Fix atom map bug in SmallMoleculeProposalEngine

pgrinaway:
There was a bug in the small molecule proposal engine that caused it to
ignore the atom and bond expressions given to it for atom matching. This
commit resolves that bug

### DIFF
--- a/perses/rjmc/topology_proposal.py
+++ b/perses/rjmc/topology_proposal.py
@@ -1168,7 +1168,7 @@ class SmallMoleculeSetProposalEngine(ProposalEngine):
 
     def __init__(self, list_of_smiles, system_generator, residue_name='MOL', atom_expr=None, bond_expr=None, proposal_metadata=None, storage=None, always_change=True):
         if not atom_expr:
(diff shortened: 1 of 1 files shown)


perses/rjmc/topology_proposal.py
```diff
@@ -1168,7 +1168,7 @@ class SmallMoleculeSetProposalEngine(ProposalEngine):
 
     def __init__(self, list_of_smiles, system_generator, residue_name='MOL', atom_expr=None, bond_expr=None, proposal_metadata=None, storage=None, always_change=True):
         if not atom_expr:
-            self.atom_expr = oechem.OEExprOpts_AtomicNumber #oechem.OEExprOpts_Aromaticity #| oechem.OEExprOpts_RingMember
+            self.atom_expr = oechem.OEExprOpts_AtomicNumber # | oechem.OEExprOpts_Aromaticity | oechem.OEExprOpts_RingMember
```
jchodera: Just the comment has changed here?

PR #298, opened 2016-12-17
Replace deepcopy of Topology with _append_topology for openmm 7.1.0 compatibility

### LINKED ISSUE #296: We cannot copy.deepcopy Topology object
This is currently done [in `NullProposalEngine.propose`](https://github.com/choderalab/perses/blob/benchmark/perses/rjmc/topology_proposal.py#L1682) and in [`PolymerProposalEngine.propose`](https://github.com/choderalab/perses/blob/benchmark/perses/rjmc/topology_proposal.py#L232-L236). See https://github.com/choderalab/perses/pull/281 for a detailed explanation of why we cannot do this and what the fix is.

Using `deepcopy` on `Topology` objects causes Perses to fail with `openmm 7.1.0` builds.

`perses` will not run correctly on `openmm 7.0.1` because of known bugs in `CustomIntegrator`. We *must* use `7.1.0`.

jchodera:
Fixes #296.


jchodera:
Unintended. Not sure what went wrong.

I'll try again in a separate PR.

PR #299, opened 2016-12-17
Fix null proposal engine

### LINKED ISSUE #296: We cannot copy.deepcopy Topology object

jchodera:
Fixes #296.

Ready for review.


jchodera:
Actually, I'm also going to pin the openmm version so we don't make this mistake yet again.

jchodera:
Pinned. Will merge when tests pass.

jchodera:
Hm. Didn't we want to just merge this into master and delete this branch instead of the other way around?

pgrinaway:
>Hm. Didn't we want to just merge this into master and delete this branch instead of the other way around?

Yeah, but there were conflicts, so I had to do a quick merge from master.

jchodera:
Thanks!

Issue #297, opened 2016-12-17
Branches are not being Travis tested


jchodera:
Unless development is done in a PR, Travis tests are not being run on branches except master, and potentially erroneous conclusions are being drawn during development because of incomplete testing. We need a solution to this to incorporate into the git flow development model if we wish to continue using it.

pgrinaway:
I suggest that we add all branches to the `.travis.yml` like this:

```yaml

# blocklist
branches:
  except:
  - /^(?i:notest)-.*$/

```

This will automatically cause travis to build on all branches, and allow a nice dashboard-like view on travis of what's going on.

pgrinaway:
Going to make sure that these changes make it through / troubleshoot if not. I will close this with a link to the functioning dashboard when it works.

pgrinaway:
Ah, seems the `.travis.yml` needs to percolate into the other branches in order to have effect (the "master only" overrides everything).

pgrinaway:
You can now see both [here](https://github.com/choderalab/perses/branches/active) and [here](https://travis-ci.org/choderalab/perses/branches) that branches with the new `.travis.yml` are being built outside of PRs.

jchodera:
Great! OK, let's give this a try for now.

PR #300, opened 2016-12-17
Allow travis builds on all branches

pgrinaway:
@jchodera noticed a flaw with the git-flow model in that travis builds were not triggered, leading to introduction of new errors. This PR would enable travis for all branches except those with `notest` in them. (I don't know at the moment why we'd want that, but it might be useful at some point). We'd then be able to check the status of all branches on the nice travis dashboard, while keeping the clean development model of git-flow.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,8 @@ addons:
       #- g++-multilib
       #- gcc-multilib
 branches:
-  only:
-    - master
+  except:
+    - /^(?i:notest)-.*$/
 
 install:
   #add random sleep from 1-10s to try to prevent overloading the anaconda servers


pgrinaway:
Addresses #297

jchodera:
When I click on `Details` next to the travis tests on this PR, I only see `master`. Where do the other branch tests show up?

pgrinaway:
>When I click on Details next to the travis tests on this PR, I only see master. Where do the other branch tests show up?

I don't think they're being tested right now, since the `.travis.yml` currently specifies that only `master` should be built (though you'll [notice](https://travis-ci.org/choderalab/perses/branches) that this branch has now been added in addition to the others).

jchodera:
Ah, OK, so we have to merge it to take effect.

pgrinaway:
I think so. I am happy to make sure this does what we expect.

PR #302, opened 2016-12-19
Complete overhaul of acceptance criteria computation to fix benchmark issues - DO NOT MERGE

jchodera:
This is a complete overhaul of the acceptance criteria computation to fix issues raised by @juliebehr's benchmark. The changes are extensive, so may have ramifications for many tests I haven't had a chance to adapt yet.

Includes:
* Overhaul of NCMC integrators
* Correction of many inconsistencies between docstrings and actual return values
* Correction of many inconsistencies between use of log probabilities and energies
* Addition of missing logP contribution for two-stage causing the benchmark test to fail
* Fixes to failing tests based on `PointMutationEngine` 
* Updates to which acceptance probability components are written to storage
* Changes to log acceptance criteria contribution definitions, so that these are all additive
* And much much more

@juliebehr's `hybrid` method was likely working just fine in the benchmark test---the issue seems to have been with the `two-stage` method.

Note that the NCMC delete and insert methods no longer return a "switch" logP. This contribution is now folded into the NCMC, requiring that the deletion and annihilation for the `two-stage` method always be paired. You can no longer expect to use one without the other.


jchodera:
The benchmark consistency check now works:
```
[LSKI1497:choderalab/perses/perses] choderaj% python perses/tests/benchmark.py

Now testing propane null transformations
Running 0 hybrid ExpandedEnsemble steps for 10 iterations
hybrid    logp_accept = -4.9415e+00 [ncmc = -4.2558e+00, geometry = -6.8571e-01]
two-stage logp_accept = -4.9415e+00 [ncmc = -4.2558e+00, geometry = -6.8571e-01]
...

Now testing naphthalene null transformations
Running 0 hybrid ExpandedEnsemble steps for 10 iterations
hybrid    logp_accept = -7.4741e+03 [ncmc = -7.4621e+03, geometry = -1.2011e+01]
two-stage logp_accept = -7.4741e+03 [ncmc = -7.4621e+03, geometry = -1.2011e+01]
...

Now testing butane null transformations
Running 0 hybrid ExpandedEnsemble steps for 10 iterations
hybrid    logp_accept = -1.9845e+01 [ncmc = -1.6875e+01, geometry = -2.9693e+00]
two-stage logp_accept = -1.9845e+01 [ncmc = -1.6875e+01, geometry = -2.9693e+00]
...
```
I did have to pare this test down to only work for 0 NCMC steps, but I don't understand how this test could have possibly worked for more than 0 steps. If someone could explain the concept of how that would work, I could restore it to work for > 0 steps.

jchodera:
I'm a bit surprised at this test failure since I had seen it pass locally. Let me look into this.

Also, somehow TESTSYSTEMS is now set to only test the valence test systems on travis. Not sure how that happened. Is that desired behavior?

pgrinaway:
>Also, somehow TESTSYSTEMS is now set to only test the valence test systems on travis. Not sure how that happened.

I think you changed that a month ago, because the travis tests take so long. Maybe we could ditch py27 and py34 in return for more systems being tested?

https://github.com/choderalab/perses/commit/b032698141fa6f0d3c9d7388f77770307da4e5cd

jchodera:
Thanks! I'll revisit this.

I think I'd be fine with ditching the other Pythons, but it would be useful to figure out a way to test more systems. I simply recall that it is just really slow without GPUs. Still hoping there is some easy way to test on GPUs in the cloud since it seems like the Jenkins server is never coming back to life, given the fights I've had with the infosec people who are freaking out that we can't disable form autofill on Jenkins. Migrating to cloud hosted Jenkins and EC2 GPU instance workers might be one solution.

perses/samplers/samplers.py
```diff
@@ -1087,39 +1087,38 @@ def _ncmc_geometry_ncmc(self, topology_proposal, positions, old_log_weight, new_
...
-
-        ncmc_new_positions, ncmc_introduction_logp, potential_insert = self._ncmc_insert(topology_proposal, geometry_new_positions)
 
-        # Compute change in eliminated potential contribution.
-        switch_logp = - (potential_insert - potential_delete)
+        ncmc_new_positions, ncmc_introduction_logp = self._ncmc_insert(topology_proposal, geometry_new_positions)
 
         # Compute total log acceptance probability, including all components.
-        logp_accept = topology_proposal.logp_proposal + geometry_logp + switch_logp + ncmc_elimination_logp + ncmc_introduction_logp + new_log_weight - old_log_weight
+        logp_accept = (topology_proposal.logp_proposal + ncmc_elimination_logp + geometry_logp_propose + geometry_logp_reverse + ncmc_introduction_logp) + (new_log_weight - old_log_weight)
```
pgrinaway: Hm, what is the new component of the acceptance criterion here that you said was missing?

pgrinaway:
>. Still hoping there is some easy way to test on GPUs in the cloud since it seems like the Jenkins server is never coming back to life

Yeah, I think we're going to have to figure out something with that. I was just search last weekend again to see if there was a ready-made GPU CI solution, but it doesn't seem like it (other than setting up Jenkins on EC2, as you suggest, which might be the best solution).

juliebehr:
> I did have to pare this test down to only work for 0 NCMC steps, but I don't understand how this test could have possibly worked for more than 0 steps. If someone could explain the concept of how that would work, I could restore it to work for > 0 steps.

It was never at any point running with > 0 steps; the test in question that is running is literally called [benchmark_logp_0ncmc_schemes](https://github.com/choderalab/perses/blob/benchmark/perses/tests/benchmark.py#L291)

juliebehr:
Also I'm not sure why this is an explanation of why the values would have been changed and why they would be more accurate now.

> The benchmark consistency check now works

Is the switch calculated in the same way but in a different place? Why would that change the two-stage ncmc values at all? If it does, how is it verified that the newly calculated values are now accurate, rather than agreeing with hybrid because the calculation from hybrid has been duplicated?

If the two schemes give logp values that are _not_ consistent, this is a sign that at least one must have an error somewhere, but just being consistent is not verification that both are correct.

jchodera:
> It was never at any point running with > 0 steps; the test in question that is running is literally called benchmark_logp_0ncmc_schemes

The function doing the printing is called [`_geometry_once_ncmc_twice`](https://github.com/choderalab/perses/blob/benchmark/perses/tests/benchmark.py#L85) and appears in a class that explicitly allows the number of NCMC steps to be set via [`options['nsteps']`](https://github.com/choderalab/perses/blob/benchmark/perses/tests/benchmark.py#L83). The function has no documentation or `assert`s that enforce `options['nsteps'] = 0`, so I had no idea if it was designed with the intent of also being used with `nsteps > 0` after testing with `nsteps = 0` was complete.

To clarify: This function is *only* ever intended to be used with `nsteps = 0`, yes?

> Also I'm not sure why this is an explanation of why the values would have been changed and why they would be more accurate now.

It's not an explanation, but it explains what I've done. I was not able to understand how the acceptance rate was currently being computed due to the large number of inconsistencies in documentation and variable naming. Instead, I overhauled the whole of acceptance probability calculation to be consistent: Each component now computes its own contribution toward the acceptance probability in a manner that is purely additive. The scheme matches what I drew on the whiteboard when we discussed this months ago, and recapitulated at the Rutgers visit with Zhiqiang Tan. 

I also eliminated the special "0 steps" NCMC case, since even if we did debug the separate "0 steps" branch, we would still have to separately debug the ">0 steps" case, which would be much more difficult.

> Is the switch calculated in the same way but in a different place?

Each stage now computes its own contribution toward the total log acceptance probability so they are purely additive. 

For the two-stage `delete`, for example, we accumulate
* The work in switching from a fully-interacting system to an alchemically-modified system
* The work in switching lambda from 1 to 0 via NCMC
* Half the work in switching from pre-geometry to post-geometry at lambda=0 (which previously was a contribution to `logp_switch`). This is more appropriately moved to `geometry`, but it is currently in the NCMC component so that geometry doesn't need to create alchemically-modified systems.

For the two-stage `insert`, we accumulate
* The other half the work in switching from pre-geometry to post-geometry at lambda=0
* The work in switching lambda from 0 to 1 via NCMC
* The work in switching from alchemically-modified to fully-interacting

> Why would that change the two-stage ncmc values at all? 

I suspect it had to do with some inconsistencies in how we were extracting the lambda=0 and lambda=1 energies from the NCMC integrator.

* If it does, how is it verified that the newly calculated values are now accurate, rather than agreeing with hybrid because the calculation from hybrid has been duplicated?

On my machine, the `test_ncmc` and `test_elimination` tests ran fine, but I'm trying to figure out why the failed on travis. 

But we really need the valence test system analytical tests from @pgrinaway to verify that the acceptance probability is correct. There is no other test that ensures correctness of the overall acceptance probability.

If the two schemes give logp values that are not consistent, this is a sign that at least one must have an error somewhere, but just being consistent is not verification that both are correct.

pgrinaway:
>Addition of missing logP contribution for two-stage causing the benchmark test to fail

What exactly is this? I can't find it in the code.

jchodera:
OK, I can replicate the hybrid test failures in `test_elimination.py` on my iMac. Investigating.

jchodera:
It looks like:
* The [wrong `lambda` functions were being used for `check_hybrid_null_elimination`](https://github.com/choderalab/perses/blob/561b725e81b7bd99cc96ce0d9c7ebb7b4b47fa8d/perses/tests/test_elimination.py#L154-L160). They all should be set to `lambda`, but were set to the same as the two-stage protocol.
* This was missed because the tests were disabled in travis until my PR.

After correcting this, the test is still failing because the alchemical correction contribution I compute [here](https://github.com/choderalab/perses/blob/561b725e81b7bd99cc96ce0d9c7ebb7b4b47fa8d/perses/annihilation/ncmc_switching.py#L644-L652) is surprisingly large, even for simple systems like pentane. I am currently including the potential energy change to switch from single-real-topology to hybrid-alchemical-topology at the beginning, as well as the potential energy change to switch from hybrid-alchemical-topology to single-real-topology at the end, which I am now suspecting may be incorrect. This contribution cancels for `nsteps = 0`, but for even `nsteps = 1`, the thermal energy that partitions into degrees of freedom that are created or deleted seems to increase the variance of the logP through this effect. I was under the impression that this should *cancel* due to the geometry stage, however, so I'm not sure why it isn't...

jchodera:
Ah, that would be why! We don't include the geometry stages in this test, hence it won't cancel.

I think this suggests we either need to refactor NCMC and the geometry engine to move the logP_correction into the geometry engine, or we need to test the whole scheme (geometry-ncmc-geometry or ncmc-geometry-ncmc) in this test.

pgrinaway:
>They all should be set to lambda

Hm, I don't think we want this, since we don't want some bonds going to `lambda=0`. Even in the hybrid case, we'll have bonds that are disappearing entirely.

>This contribution cancels for nsteps = 0, but for even nsteps = 1, the thermal energy that partitions into degrees of freedom that are created or deleted seems to increase the variance of the logP through this effect.

Yeah, that makes sense. But this isn't strictly incorrect, right? It's just unfavorable.

>I think this suggests we either need to refactor NCMC and the geometry engine to move the logP_correction into the geometry engine, or we need to test the whole scheme (geometry-ncmc-geometry or ncmc-geometry-ncmc) in this test.

I do think we should test the whole scheme, but shouldn't the NCMC null elimination still work by itself? I'd think that it should still preserve the equilibrium distribution of the configurations of whatever molecule is being "null" tested.. Perhaps it's not a very useful test in the long run, but I'd imagine that if the test is there and the NCMC bits are working, it should pass without the RJ stuff, unless I'm missing something.

jchodera:
Forgot to specify that this is specifically the hybrid test that is failing. There's we MUST have all lambdas start at 0 and end at 1, and the protocol must be symmetric.

pgrinaway:
>There's we MUST have all lambdas start at 0 and end at 1, and the protocol must be symmetric.

Won't that cause an efficiency problem with some bonds being eliminated entirely?

jchodera:
No bonds are eliminated; all valence terms are retained at full strength throughout if softening = 1, as it must to avoid large geomtrey penalities.

jchodera:
I spoke with @pgrinaway today, who pointed out some major issues with my overhaul and areas in which I had messed things up. I've now rederived the hybrid scheme from scratch, and will put this derivation and the two-stage derivation in our [overleaf document](https://www.overleaf.com/read/nrkwjwjskpkd), and then make sure this implementation matches.

jchodera:
Will split this into two separate PRs.

jchodera:
Closing in favor of #307.

PR #303, opened 2016-12-19
Fix sams

pgrinaway:
This PR includes a couple fixes for the `SAMSSampler`, as well as for the `ProposalEngine`:

* Add a property in the base class `ProposalEngine` to retrieve the set of possible chemical states
* Ensure that the list of chemical states in the `SAMSSampler` contains the same elements as on initialization, and only elements that are able to be proposed by the `ProposalEngine`
* Add a subtraction step as specified in the paper.


perses/samplers/samplers.py
```diff
@@ -1374,19 +1374,43 @@ def __init__(self, sampler, logZ=None, log_target_probabilities=None, update_met
...
-        else:
-            self.log_target_probabilities = dict()
+            for (chemical_state, log_target_probability) in log_target_probabilities:
+                if chemical_state not in self.chemical_states:
+                    raise ValueError("Provided a log target probability for an un-proposable chemical state.")
+                self.log_target_probabilities[chemical_state] = log_target_probability
+
+            #normalize target probabilities
+            #this is likely not necessary, but it is copying the algorithm in Ref 1
+            log_sum_target_probabilities = np.log(np.sum(np.exp(list(self.log_target_probabilities.values()))))
```
jchodera: Don't use this idiom! It is not numerically stable.

Use [`logsumexp`](https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.misc.logsumexp.html) instead.

pgrinaway: Ok, changing

pgrinaway:
Thanks! One question, actually. I just realized from the test--I've implemented the property to return the list of chemical states from `SmallMoleculeProposalEngine` but not the `PolymerProposalEngines`. That's going to make this test fail (as it should), since the `SAMSSampler` now uses that property to get the list of chemical states. 

* I can try to implement that for the `PolymerProposalEngines` here (though I'm not exactly sure how.

* Right now, the base class implementation just returns an empty list. Should I instead raise a `NotImplementedError` or something more informative like that?

jchodera:
> I can try to implement that for the PolymerProposalEngines here (though I'm not exactly sure how.

We're going to need to make that change if you want to have the SAMS sampler operate this way.

What about, instead of eliminating the old SAMS sampler behavior, you allow the old or new behavior to be selectable as an option? If the ProposalEngine returns a `NotImplementedError`, the SAMS sampler will use the old behavior (no subtraction to keep lowest free energy zero); otherwise, it will use the new behavior. Alternatively, this could be selected via a flag.

> Right now, the base class implementation just returns an empty list. Should I instead raise a NotImplementedError or something more informative like that?

Yes!

pgrinaway:
> If the ProposalEngine returns a NotImplementedError, the SAMS sampler will use the old behavior (no subtraction to keep lowest free energy zero); otherwise, it will use the new behavior. Alternatively, this could be selected via a flag.

Ok. Perhaps I can pair this with a logger warning to alert the user as to what happened, and then we can fix the other `ProposalEngines` in a future PR.

pgrinaway:
Ok, I see what happened here. I will fix this.

pgrinaway:
Ok, tests pass, going to merge this.

jchodera:
Awesome, thanks!

PR #304, opened 2016-12-24
[WIP] NCMC update

jchodera:
This is just the NCMC updates (and consequent required changes to tests and samplers) from #302.
* Cleans up and refactors the NCMC integrators
* Remove special zero-step cases to make code easier to test
* No longer computes `logp_switch` separately due to confusion over components. We may want to break things out even finer, however, once we have assigned unique expressions to all components of the acceptance criteria in our document.
* Fixes some issues with hybrid elimination tests using wrong alchemical `lambda` functions.

### DIFF
--- a/perses/annihilation/ncmc_switching.py
+++ b/perses/annihilation/ncmc_switching.py
@@ -195,6 +195,7 @@ class NCMCEngine(object):
         initial_unmodified_potential = self.beta * compute_potential(system, initial_positions, platform=self.platform)
 
         if direction == 'insert':
+            # Part of Eq. 31 of acceptance criteria from perses document
             logP = - (final_unmodified_potential - final_alchemical_potential) - initial_alchemical_potential
 
             # Check potentials are finite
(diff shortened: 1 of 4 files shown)


jchodera:
I had to bring in some fixes for `PointMutationEngine` so I could run all the tests in `test_elimination.py`.

jchodera:
Whoops! I had accidentally not merged in @pgrinaway's additions for retrieving the `chemical_state_list` from #303.

@pgrinaway : Did you make those changes for `PointMutationEngine` as well? I think I fixed the other issues with `PointMutationEngine`---caused by using the `Topology` internal API that broke---so it would be great if we could re-enable those tests like I have done in this PR to make sure the NCMC stuff works.

jchodera:
I think the remaining issue is that I have to make sure the NCMC annihilation tests are still testing the correct thing given that have included the real/alchemical energy switching contributions in with the exponentiated NCMC work contribution. I realize now that this might be a mistake because the geometry forward/backward contributions may be needed for cancellation. I might try this one more time in a separate PR that maintains the separation of those two so we can test the NCMC work separately from the geometry + switching contributions (which should cancel).

jchodera:
I'm going to create a new PR that preserves the separation between work and energy changes so that (1) we can test the work separately, and (2) we can use a division of acceptance probability terms that follows the document.

PR #312, opened 2017-01-08
[WIP] Test on AWS using travis deploy

jchodera:
I'm opening a new PR to discuss testing on AWS.

The `pr` tests won't deploy, but the `push` tests should deploy. 

One issue will be to figure out how to ensure the branches don't overwrite each other if they are pushing at the same time. Alternatively, we can restrict testing to the `master` branch.

The next step is to see if we can launch a container to test in.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,16 +68,13 @@ after_success:
   - if [ "$TRAVIS_SECURE_ENV_VARS" == true ]; then source devtools/travis-ci/after_success.sh; fi
 
 deploy:
-  # Don't delete build artifacts
-  skip_cleanup: true
-  # Upload to S3 bucket
+  skip_cleanup: true # Don't delete build artifacts
   provider: s3
   access_key_id:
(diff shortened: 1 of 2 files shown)


jchodera:
Looks like the next steps are considerably more difficult. We'd want to spin up an EC2 instance, launch a docker image on it, run the tests in the docker image, capture the output, and then spin everything down.

pgrinaway:
I think this is something that will benefit multiple projects, so I'm happy to discuss plans when others are back. I think there are a few alternate approaches we can take on AWS:

* We can have travis send a message to an AWS SQS queue or directly to an AWS lambda function, triggering the above events from within the AWS infrastructure, which might be easier.

* We can use AWS lambda to write a webhook (at least getting the webhook together is pretty simple) that directly triggers these builds from pushes

* Jenkins, but this might be a PITA

I think it's worth collaborating with other interested people in the lab so that we can figure out what the most straightforward and effective solution will be.

jchodera:
OK, I'll put this on hold and we can discuss when we get back. I need to use the S3 deployment scheme to capture build artifacts for OpenMM, so this was a useful exercise.

I think there may be advantages to capturing the test output into the Travis log, so certain schemes may work better than others.

jchodera:
Closing this, since we will engineer a different scheme at some point in future.

PR #316, opened 2017-01-19
Storage fixes

pgrinaway:
This is a small PR with a few storage fixes:

* `get_object` had the variable name hardcoded, which prevented it from being used. This is fixed, but I had to add two extra parameters to `get_object`: `envname` and `modname`. 

* The test happened to use the exact same name as the hardcoded variable, so it was passing. I changed it to use some names we might encounter in simulation (though maybe even just a random string would suffice)

I debated for a bit about how to do this exactly in the spirit of the API, but I decided on the quick patch now so that we could keep going. I'm happy to take a look at this again later.

### DIFF
--- a/perses/storage/storage.py
+++ b/perses/storage/storage.py
@@ -211,7 +211,9 @@ class NetCDFStorage(object):
             The retrieved object
 
         """
+
         nc_path = "/{envname}/{modname}/{varname}".format(envname=envname, modname=modname, varname=varname)
+        
         if iteration is not None:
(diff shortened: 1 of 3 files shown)


jchodera:
I swear that was fixed in another PR!

pgrinaway:
Maybe it was accidentally undone?

jchodera:
Oof, that seems like a possibility!

pgrinaway:
Accidentally created a conflict here yesterday, just resolved + will merge when tests pass.

PR #324, opened 2017-02-21
Fix name of variable error in NCMCIntegrator

### LINKED ISSUE #323: Non-specified kinetic energy variable in addGHMCStep
In https://github.com/choderalab/perses/blob/master/perses/annihilation/ncmc_switching.py#L911-L922 there are two kinetic energy variables, "kinetic" and "ke". However ke is never actually added to the set of global variables so it causes the NCMCGHMCAlchemicalIntegrator to behave improperly.

pgrinaway:
This changes the unused `ke` to `kinetic`. Let me know if something is amiss. Should resolve #323


jchodera:
Odd that the `imatinib` NCMC test is still failing, but this fix is certainly correct, so please merge! I'll look into the NCMC failures separately.

jchodera:
Merge away!

PR #325, opened 2017-03-22
Fix travis builds

jchodera:
This pins `conda-build` to 2.1.5 until https://github.com/conda/conda-build/issues/1825 is resolved.


jchodera:
@pgrinaway : Please review when able.

PR #348, opened 2017-05-31
Relative free energy roll-up

pgrinaway:
This PR is rather large (sorry), but it contains/will contain several major changes:

- [x] New hybrid topology builder base

- [x] Docstrings and comments to explain what's going on (mostly complete, but some are still missing. I also appreciate comments regarding where code is unclear or potentially wrong--I will fix and document).

- [x] Fix existing tests

- [x] New tests for hybrid topology builder

To elaborate on some of the changes:

* Some changes in the new hybrid topology builder:
    * Use set arithmetic where possible to avoid large, confusing loops
    * Try to make code as readable as possible
    * Document what it does and why, so that future developers (including future me) can understand what it is doing.
    * Fix some issues with nonbonded exception handling (treat as valence terms, not nonbonded terms)

* Issues with tests:
    * The hybrid elimination test was generating an atom map that contained only the carbons from biphenyl. This means that at `lambda=0`, we have biphenyl, but at `lambda=1`, we have biphenyl without hydrogens. I wouldn't expect the free energy difference there to be zero. In its place, I will put a test using anthracene where two rings are mapped to each end state--this makes the `lambda=0` and `lambda=1` states the same.


pgrinaway:
@andrrizzi @Lnaden when you have a chance, I'd love to see your input. I realize it's large--sorry about that.

Also, I will rename the new hybrid builder before merging.

jchodera:
It looks like you've added a new file `new_relative.py`, which makes it difficult to compare changes. Is it so different that it is better for us to just review the new code on its own rather than the changes?

perses/annihilation/new_relative.py
```diff
@@ -0,0 +1,1213 @@
...
+
+    def __init__(self, topology_proposal, current_positions, new_positions):
+        """
+        Initialize the Hybrid topology factory.
+
+        Parameters
+        ----------
+        topology_proposal : perses.rjmc.topology_proposal.TopologyProposal object
+            TopologyProposal object rendered by the ProposalEngine
+        current_positions : [n,3] np.ndarray of float
```
jchodera: Do these *have* to be stripped of units, or are unit-bearing quantities allowed too?
Same for new_positions.

pgrinaway:
>It looks like you've added a new file new_relative.py, which makes it difficult to compare changes. Is it so different that it is better for us to just review the new code on its own rather than the changes?

Yes, it is actually quite different.

perses/annihilation/new_relative.py
pgrinaway: Hm, I will go through and take a look. Would you prefer them to be unit-bearing?

perses/annihilation/new_relative.py
```diff
@@ -0,0 +1,1213 @@
...
+        """
+        This function adds the appropriate PeriodicTorsionForce terms to the system. Core torsions are interpolated,
+        while environment and unique torsions are always on.
+        """
+        energy_expression  = '(1-lambda_torsions)*U1 + lambda_torsions*U2;'
+        energy_expression += 'U1 = K1*(1+cos(periodicity1*theta-phase1));'
+        energy_expression += 'U2 = K2*(1+cos(periodicity2*theta-phase2));'
+
+        #create the force and add the relevant parameters
+        custom_core_force = openmm.CustomTorsionForce(energy_expression)
```
jchodera: How does this work if the torsions in A and B have different numbers of torsion terms?

perses/annihilation/new_relative.py
```diff
@@ -0,0 +1,1213 @@
...
+        electrostatics_energy_expression : str
+            The energy expression for electrostatics
+        """
+        # soft-core Lennard-Jones
+        sterics_energy_expression = "U_sterics = 4*epsilon*x*(x-1.0); x = (sigma/reff_sterics)^6;"
+        # soft-core Coulomb
+        electrostatics_energy_expression = "U_electrostatics = ONE_4PI_EPS0*chargeprod/reff_electrostatics;"
+        return sterics_energy_expression, electrostatics_energy_expression
+
+    def _nonbonded_custom_cutoff(self, epsilon_solvent, r_cutoff):
```
jchodera: This seems to be missing a way to copy over how the dispersion correction is handled for Lennard-Jones interactions. This is why we have to split the `NonbondedForce` into two `CustomNonbondedForces` in `AbsoluteAlchemicalFactory`: The LJ replacement has a dispersion correction and electrostatics does not. We *may* be able to get away with omitting it for small compounds, but it will cause problems for larger compounds.

perses/annihilation/new_relative.py
```diff
@@ -0,0 +1,1213 @@
...
+        Returns
+        -------
+        sterics_energy_expression : str
+            The energy expression for U_sterics
+        electrostatics_energy_expression : str
+            The energy expression for electrostatics
+        """
+        # soft-core Lennard-Jones
+        sterics_energy_expression = "U_sterics = 4*epsilon*x*(x-1.0); x = (sigma/reff_sterics)^6;"
+        if alpha_ewald == 0.0:
```
andrrizzi: `alpha_ewald` here may be a `Quantity` object, in which case could always evaluates to `False` even if its value is 0.0. Probably worth checking.

perses/annihilation/new_relative.py
jchodera: It would be great to be able to handle both `Quantity`-wrapped numpy arrays and naked numpy arrays.

perses/annihilation/new_relative.py
```diff
@@ -0,0 +1,1213 @@
...
+        #first, loop through all the angles in the old system to determine what to do with them. We will only use the
+        #custom angle force if all atoms are part of "core." Otherwise, they are either unique to one system or never
+        #change.
+        for angle_index in range(old_system_angle_force.getNumAngles()):
+            angle_parameters = old_system_angle_force.getAngleParameters(angle_index)
+
+            #get the indices in the hybrid system
+            hybrid_index_set = {self._old_to_hybrid_map[old_index] for old_index in angle_parameters[:3]}
+            #make a list because python set objects can't be indexed
+            hybrid_index_list = list(hybrid_index_set)
```
andrrizzi: I may not understand the logic here, but I think that building the list this way may randomly scramble the order of the atom indices, which could mean you'd end up controlling another angle.

perses/annihilation/new_relative.py
```diff
@@ -0,0 +1,1213 @@
...
+                                                                            angle_parameters[4])
+
+        #finally, loop through the new system force to add any unique new angles
+        for angle_index in range(new_system_angle_force.getNumAngles()):
+            angle_parameters = new_system_angle_force.getAngleParameters(angle_index)
+
+            #get the indices in the hybrid system
+            hybrid_index_set = {self._old_to_hybrid_map[new_index] for new_index in angle_parameters[:3]}
+            #make a list because python set objects can't be indexed
+            hybrid_index_list = list(hybrid_index_set)
```
andrrizzi: Here too.

perses/annihilation/new_relative.py
```diff
@@ -0,0 +1,1213 @@
...
+        #first, loop through all the torsions in the old system to determine what to do with them. We will only use the
+        #custom torsion force if all atoms are part of "core." Otherwise, they are either unique to one system or never
+        #change.
+        for torsion_index in range(old_system_torsion_force.getNumTorsions()):
+            torsion_parameters = old_system_torsion_force.getTorsionParameters(torsion_index)
+
+            #get the indices in the hybrid system
+            hybrid_index_set = {self._old_to_hybrid_map[old_index] for old_index in torsion_parameters[:4]}
+            #make a list because python set objects can't be indexed
+            hybrid_index_list = list(hybrid_index_set)
```
andrrizzi: Here too.

perses/annihilation/new_relative.py
```diff
@@ -0,0 +1,1213 @@
...
+                                           new_torsion_parameters[4], new_torsion_parameters[5],
+                                           new_torsion_parameters[6]]
+                self._hybrid_system_forces['core_torsion_force'].addTorsion(hybrid_index_list[0], hybrid_index_list[1], hybrid_index_list[2], hybrid_index_list[3], hybrid_force_parameters)
+
+            #otherwise, just add the parameters to the regular force:
+            else:
+                self._hybrid_system_forces['standard_torsion_force'].addTorsion(hybrid_index_list[0], hybrid_index_list[1],
+                                                                            hybrid_index_list[2], hybrid_index_list[3], torsion_parameters[4],
+                                                                            torsion_parameters[5], torsion_parameters[6])
+
```
andrrizzi: Are the torsions of the unique atoms of the new system missing here, or it just doesn't make sense to add them.

perses/annihilation/new_relative.py
```diff
@@ -0,0 +1,1213 @@
...
+                                                                                 [chargeProd_old, sigma_old,
+                                                                                  epsilon_old, chargeProd_old,
+                                                                                  sigma_old, epsilon_old])
+
+            #If the exception particles are neither solely old unique, solely environment, nor contain any unique old atoms, they are either core/environment or core/core
+            #In this case, we need to get the parameters from the exception in the other (new) system, and interpolate between the two
+            else:
+                #first get the new indices.
+                index1_new = hybrid_to_old_map[index1_hybrid]
+                index2_new = hybrid_to_old_map[index2_hybrid]
```
andrrizzi: Typo in these two? Should it be `hybrid_to_new_map`?

perses/annihilation/new_relative.py
```diff
@@ -0,0 +1,1213 @@
...
+
+        #now, loop through the new system to collect remaining interactions. The only that remain here are
+        #uniquenew-uniquenew, uniquenew-core, and uniquenew-environment.
+        for exception_index in range(new_system_nonbonded_force.getNumExceptions()):
+            [index1_new, index2_new, chargeProd_new, sigma_new,
+             epsilon_new] = old_system_nonbonded_force.getExceptionParameters(exception_index)
+
+            #get hybrid indices:
+            index1_hybrid = self._old_to_hybrid_map[index1_new]
+            index2_hybrid = self._old_to_hybrid_map[index2_new]
```
andrrizzi: Should it be `self._new_to_hybrid_map` here?

perses/annihilation/new_relative.py
```diff
@@ -0,0 +1,1213 @@
...
+        self._old_system = copy.deepcopy(topology_proposal.old_system)
+        self._new_system = copy.deepcopy(topology_proposal.new_system)
+        self._old_to_hybrid_map = {}
+        self._new_to_hybrid_map = {}
+        self._hybrid_system_forces = dict()
+        self._old_positions = current_positions
+        self._new_positions = new_positions
+
+        self.softcore_alpha=0.5
+        self.softcore_beta=12*unit.angstrom**2
```
pgrinaway: Ok, that sounds good. Will change.

andrrizzi:
Also, I haven't seen any call to `CustomNonbondedForce.addExclusion()` that correspond to the standard `NonbondedForce` exception, but I guess that's part of the [WIP] status?

perses/annihilation/new_relative.py
```diff
@@ -0,0 +1,1213 @@
...
+        hybrid_positions_array = np.zeros([n_atoms_hybrid, 3])
+
+        #loop through the old system indices, and assign positions.
+        for old_index, hybrid_index in self._old_to_hybrid_map.items():
+            hybrid_positions_array[hybrid_index, :] = old_positions_without_units[old_index, :]
+
+        #Do the same for new indices. Note that this overwrites some coordinates, but as stated above, the assumption
+        #is that these are the same.
+        for new_index, hybrid_index in self._new_to_hybrid_map.items():
+            hybrid_positions_array[hybrid_index, :] = old_positions_without_units[new_index, :]
```
andrrizzi: Should it be `new_positions_without_units` here?

perses/annihilation/new_relative.py
```diff
@@ -0,0 +1,1213 @@
...
+                particle_index_in_new_system = self._topology_proposal.old_to_new_atom_map[particle_idx]
+                self._new_to_hybrid_map[particle_index_in_new_system] = hybrid_idx
+
+        #Next, add the remaining unique atoms from the new system to the hybrid system and map accordingly.
+        #As before, this does not copy interactions, only particle indices and masses.
+        for particle_idx in self._topology_proposal.unique_new_atoms:
+            particle_mass = self._new_system.getParticleMass(particle_idx)
+            hybrid_idx = self._hybrid_system.addParticle(particle_mass)
+            self._new_to_hybrid_map[particle_idx] = hybrid_idx
+
```
pgrinaway: Should we copy virtual sites that are part of the environment though? I'd imagine that we should to support things like TIP4P?

perses/annihilation/new_relative.py
```diff
@@ -0,0 +1,1213 @@
...
+        #assign atoms to one of the classes described in the class docstring
+        self._atom_classes = self._determine_atom_classes()
+
+        #verify that no constraints are changing over the course of the switching.
```
pgrinaway: Good catch.

perses/annihilation/new_relative.py
```diff
@@ -0,0 +1,1213 @@
...
+        self._constraint_check()
+
+        #call each of the methods to add the corresponding force terms and prepare the forces:
+        self._add_bond_force_terms()
+        self._add_angle_force_terms()
+        self._add_torsion_force_terms()
+        self._add_nonbonded_force_terms()
+
+        #call each force preparation method to generate the actual interactions that we need:
```
pgrinaway: Another good catch.

perses/annihilation/new_relative.py
pgrinaway: Good point. For small molecules and GAFF, I _think_ that doesn't happen, if I recall the paper correctly. But in general it will so we should deal with it.

perses/annihilation/new_relative.py
```diff
@@ -0,0 +1,1213 @@
...
+        """
+        Get a custom electrostatics expression that is common to all nonbonded methods
+
+        Returns
+        -------
+        electrostatics_addition : str
+            The common electrostatics energy expression
+        """
+        electrostatics_addition = "chargeprod = (1-lambda_electrostatics)*chargeprodA + lambda_electrostatics*chargeprodB;" #interpolation
+        electrostatics_addition += "reff_electrostatics = sqrt(softcore_beta*lambda_beta + r^2);" # effective softcore distance for electrostatics
```
pgrinaway: +1

perses/annihilation/new_relative.py
pgrinaway: Huh, ok, I'll take a look at the implementation in `AbsoluteAlchemicalFactory`.

perses/annihilation/new_relative.py
jchodera: Yes! We don't have to support virtual sites in the alchemical region (probably should throw an exception) but we should definitely support vsites elsewhere.

perses/annihilation/new_relative.py
jchodera: @andrrizzi is right. You want the opposite idiom, where you first build the list (to preserve atom ordering) and then build the set from the list (to use set operations):
```python
hybrid_index_list = [self._old_to_hybrid_map[old_index] for old_index in angle_parameters[:3]]
hybrid_index_set = set(hybrid_index_list)
```

jchodera:
@pgrinaway : Can you also expose the softcore parameters `alpha, a, b, c` (for Lennard-Jones) and `beta, d, e, f` as global parameters, just as we do for [`AbsoluteAlchemicalFactory`](https://github.com/choderalab/openmmtools/blob/master/openmmtools/alchemy.py#L1709-L1717)? I'm certain we'll need to optimize these later.

perses/annihilation/new_relative.py
pgrinaway: Yes! thanks!

perses/annihilation/new_relative.py
pgrinaway: I don't think this is an issue, since OpenMM represents additional multiplicities as additional terms.

perses/annihilation/new_relative.py
```diff
@@ -0,0 +1,1213 @@
...
+
+        #prepare dicts of forces, which will be useful later
+        self._old_system_forces = {type(force).__name__ : force for force in self._old_system.getForces()}
+        self._new_system_forces = {type(force).__name__ : force for force in self._new_system.getForces()}
+
+        #get and store the nonbonded method from the system:
+        self._nonbonded_method = self._old_system_forces['NonbondedForce'].getNonbondedMethod()
+
+        #start by creating an empty system. This will become the hybrid system.
+        self._hybrid_system = openmm.System()
```
pgrinaway: I realize what you are saying, but I think it's a tradeoff. I actually think that it's easier to make a mistake assigning parameters to particles (as well as leading to more confusing code, as the previous hybrid implementation) if we copy a system and then modify it. On the other hand, it's easy to forget the barostat, etc. if we create a new system. However, it's much easier (in my view) to make sure that we remember those things (barostat, virtual sites, etc.) than to ensure that we have deleted parameters correctly. 

With regard to weird forces, I think we should just throw an exception (or maybe a warning). I don't think we want to try to handle unknown forces.

perses/annihilation/new_relative.py
```diff
@@ -0,0 +1,1213 @@
...
+        -------
+        sterics_energy_expression : str
+            The energy expression for U_sterics
+        electrostatics_energy_expression : str
+            The energy expression for electrostatics
+        """
+        # soft-core Lennard-Jones
+        sterics_energy_expression = "U_sterics = 4*epsilon*x*(x-1.0); x = (sigma/reff_sterics)^6;"
+
+        electrostatics_energy_expression = "U_electrostatics = ONE_4PI_EPS0*chargeprod*(reff_electrostatics^(-1) + k_rf*reff_electrostatics^2 - c_rf);"
```
jchodera: I'm pretty sure we need to force `c_rf = 0` for any interactions involving the alchemical atoms in order to not have reaction field generate garbage.

For atoms that remain in `standard_nonbonded_force`, we're probably fine.

perses/annihilation/new_relative.py
```diff
@@ -0,0 +1,1351 @@
...
+        sterics_custom_nonbonded_force.addPerParticleParameter("sigmaB") # Lennard-Jones sigma final
+        sterics_custom_nonbonded_force.addPerParticleParameter("epsilonB") # Lennard-Jones epsilon final
+
+
+
+        self._hybrid_system.addForce(sterics_custom_nonbonded_force)
+        self._hybrid_system_forces['core_sterics_force'] = sterics_custom_nonbonded_force
+
+        #Add a regular nonbonded force for all interactions that are not changing.
+        standard_nonbonded_force = openmm.NonbondedForce()
```
jchodera: Where do you copy over any of the reference `NonbondedForce` settings, such as whether the dispersion correction is in use, PME error tolerance, cutoff, nonbonded method, or all the other stuff?

jchodera: If this was missed, it means the tests are not catching things that are sensitive to this. Can you add a test that is not just a molecule in vacuum, but maybe is a small molecule in some environment that contains other interacting particles that aren't part of the alchemical region? It could even be a larger molecule where only some atoms are involved in the alchemical part.

perses/annihilation/new_relative.py
jchodera: We want to make sure that the `standard_nonbonded_force` has the same dispersion correction (and any other settings) as the reference `NonbondedForce`. 

You should also set the `setUseLongRangeCorrection()` for the `CustomNonbondedForce` you create to handle sterics, but allow it to be forced off by a factory-level setting as in `AbsoluteAlchemicalFactory` [here](https://github.com/choderalab/openmmtools/blob/master/openmmtools/alchemy.py#L680-L685) so we can turn it off for NCMC use for efficiency.

perses/annihilation/new_relative.py
```diff
@@ -0,0 +1,1351 @@
...
+            sterics_energy_expression, electrostatics_energy_expression = self._nonbonded_custom_ewald(alpha_ewald, delta, r_cutoff)
+        else:
+            raise Exception("Nonbonded method %s not supported yet." % str(self._nonbonded_method))
+        sterics_energy_expression += self._nonbonded_custom_sterics_common()
+        electrostatics_energy_expression += self._nonbonded_custom_electrostatics_common()
+
+        sterics_mixing_rules, electrostatics_mixing_rules = self._nonbonded_custom_mixing_rules()
+
+        # Create CustomNonbondedForce to handle interactions between alchemically-modified atoms and rest of system.
+        electrostatics_custom_nonbonded_force = openmm.CustomNonbondedForce("U_electrostatics;" + electrostatics_energy_expression + electrostatics_mixing_rules)
```
jchodera: Where do you call `electrostatics_custom_nonbonded_force.setNonbondedMethod()` and `sterics_custom_nonbonded_force.setNonbondedMethod()` to ensure these match their original reference `NonbondedForce` nonbonded methods?

scripts/setup_relative_calculation.py
```diff
@@ -0,0 +1,151 @@
...
+    ifs = oechem.oemolistream()
+    ifs.open(mol_filename)
+    #get the list of molecules
+    mol_list = [mol for mol in ifs.GetOEMols()]
+    #we'll always take the first for now
+    return mol_list[0]
+
+
+
+class RelativeFreeEnergySetup(object):
```
jchodera: Do you have any tests for this class?

pgrinaway: It's not finished. I think I'm going to finish it in another PR. I can remove it for now if that's better.

perses/tests/test_hybrid_builder.py
```diff
@@ -242,16 +242,90 @@ def compute_logP(system, positions, parameter=None):
...
+
+    integrator = openmm.VerletIntegrator(1)
+    platform = openmm.Platform.getPlatformByName("Reference")
+    context = openmm.Context(alchemical_system, integrator, platform)
+
+    context.setPositions(alchemical_positions)
+
+    functions = {
+        'lambda_sterics' : '2*lambda * step(0.5 - lambda) + (1.0 - step(0.5 - lambda))',
+        'lambda_electrostatics' : '2*(lambda - 0.5) * step(lambda - 0.5)',
```
jchodera: These `lambda_sterics` and `lambda_electrostatics` functions are only appropriate for absolute calculations, aren't they? Don't you want to change both simultaneously as `lambda`? The protocol is also asymmetric. 
image

perses/tests/test_hybrid_builder.py
```diff
@@ -242,16 +242,90 @@ def compute_logP(system, positions, parameter=None):
...
+        'lambda_bonds' : '1.0', # don't soften bonds
+        'lambda_angles' : '1.0', # don't soften angles
```
jchodera: Don't you want to have bonds and angles change from molecule A to molecule B as `lambda`, like torsions?
You're using the HybridTopologyFactory here, not the Absolute factory.

perses/tests/test_hybrid_builder.py
```diff
@@ -242,16 +242,90 @@ def compute_logP(system, positions, parameter=None):
...
+
+    initial_energy = context.getState(getEnergy=True).getPotentialEnergy()
+
+    #set all to one
+    for parm in functions.keys():
+        context.setParameter(parm, 1.0)
+
+    final_energy = context.getState(getEnergy=True).getPotentialEnergy()
+
+    if np.abs(final_energy - initial_energy) > 1.0e-6*unit.kilojoule_per_mole:
```
jchodera: This might be too tight a criteria if you are using, say, non-deterministic PME CUDA.

perses/tests/test_hybrid_builder.py
pgrinaway: I copied this from the old hybrid topology stuff. I can change it to something else.

jchodera: Ah, I see---these particular functions may not matter if you're just checking lambda 0 -> 1.

perses/tests/test_elimination.py
```diff
@@ -136,7 +286,117 @@ def check_alchemical_null_elimination(topology_proposal, positions, ncmc_nsteps=
...
+        The topology proposal to test.
+        This must be a null transformation, where topology_proposal.old_system == topology_proposal.new_system
+    ncmc_steps : int, optional, default=50
+        Number of NCMC switching steps, or 0 for instantaneous switching.
+    NSIGMA_MAX : float, optional, default=6.0
+    """
+    functions = {
+        'lambda_sterics' : 'lambda',
+        'lambda_electrostatics' : 'lambda',
+        'lambda_bonds' : 'lambda', # don't soften bonds
```
jchodera: Remove comments here?

perses/tests/test_hybrid_builder.py
pgrinaway: I also copied this from the old hybrid, but I can change it.

perses/tests/test_elimination.py
pgrinaway: Ok

perses/tests/test_hybrid_builder.py
pgrinaway: Ok. What do you recommend?

perses/tests/test_elimination.py
```diff
@@ -51,6 +51,156 @@ def simulate(system, positions, nsteps=500, timestep=1.0*unit.femtoseconds, temp
...
+    platform = openmm.Platform.getPlatformByName("Reference")
+
+    hybrid_system = factory.hybrid_system
+    hybrid_topology = factory.hybrid_topology
+    initial_hybrid_positions = factory.hybrid_positions
+
+    n_iterations = 50 #number of times to do NCMC protocol
+
+    #alchemical functions
+    functions = {
```
jchodera: These functions are only appropriate for absolute factory, not relative.

pgrinaway: Ok, once again, copied from the old hybrid. I will change. What do you suggest?

perses/annihilation/new_relative.py
jchodera: Docstrings need to be updated if these can handle quantities.

perses/annihilation/new_relative.py
```diff
@@ -0,0 +1,1433 @@
...
+        self._old_to_hybrid_map = {}
+        self._new_to_hybrid_map = {}
+        self._hybrid_system_forces = dict()
+        self._old_positions = current_positions
+        self._new_positions = new_positions
+
+        self._use_dispersion_correction = use_dispersion_correction
+
+        self.softcore_alpha=0.5
+        self.softcore_beta=12*unit.angstrom**2
```
jchodera: I think we still want to change this to unitless to match absolute factory.

perses/annihilation/new_relative.py
```diff
@@ -0,0 +1,1433 @@
...
+        #Next, add the remaining unique atoms from the new system to the hybrid system and map accordingly.
+        #As before, this does not copy interactions, only particle indices and masses.
+        for particle_idx in self._topology_proposal.unique_new_atoms:
+            particle_mass = self._new_system.getParticleMass(particle_idx)
+            hybrid_idx = self._hybrid_system.addParticle(particle_mass)
+            self._new_to_hybrid_map[particle_idx] = hybrid_idx
+
+        #check that if there is a barostat in the original system, it is added to the hybrid.
+        #We copy the barostat from the old system.
+        if "MonteCarloBarostat" in self._old_system_forces.keys():
```
jchodera: What about other forces, like AndersonThermostat or something weird? Can you just check to see if there are no forces you don't explicitly handle here and throw an exception if either system contains something not explicitly copied? Otherwise we are going to run into trouble with this construction at some point, and it will take a long time to debug.

perses/annihilation/new_relative.py
jchodera: By unitless, I mean that softcore_beta is not just lacking units, but uses the LJ effective `sigma` to set the scale of the interaction. See here:
https://github.com/choderalab/openmmtools/blob/master/openmmtools/alchemy.py#L1472

perses/annihilation/new_relative.py
```diff
@@ -0,0 +1,1433 @@
...
+            barostat = copy.deepcopy(self._old_system_forces["MonteCarloBarostat"])
+            self._hybrid_system.addForce(barostat)
+
+        #initialize unitless softcore beta
+        self.softcore_beta = self.softcore_beta / self.softcore_beta.in_unit_system(unit.md_unit_system).unit
```
jchodera: Sorry, this isn't what I meant about storing unitless beta. It uses the LJ sigma to set the lengthscale of the softening in computing the effective separation:
https://github.com/choderalab/openmmtools/blob/master/openmmtools/alchemy.py#L1472

perses/annihilation/new_relative.py
jchodera: Did we resolve the case where the total number of torsion terms*multipicities may differ between the two molecules?

perses/annihilation/new_relative.py
```diff
@@ -0,0 +1,1433 @@
...
+        Get a custom sterics expression that is common to all nonbonded methods
+
+        Returns
+        -------
+        sterics_addition : str
+            The common softcore sterics energy expression
+        """
+        sterics_addition = "epsilon = (1-lambda_sterics)*epsilonA + lambda_sterics*epsilonB;" #interpolation
+        sterics_addition += "reff_sterics = sigma*((softcore_alpha*lambda_alpha + (r/sigma)^6))^(1/6);" # effective softcore distance for sterics
+        #sterics_addition += "softcore_alpha = %f;" % self.softcore_alpha
```
jchodera: Remnant comment line?

perses/annihilation/new_relative.py
pgrinaway: Yes.

perses/annihilation/new_relative.py
jchodera: I'm still worried about this, but it's OK for now if we (1) throw an `Exception` if there are any `Force` objects we aren't sure we explicitly know how to handle, and (2) are vigilant about API changes/additions to force changes we *do* know how to handle.

perses/annihilation/new_relative.py
jchodera: Don't do this. It's unsafe to just strip units if the user can later change the class field.

If you're going to keep unit-bearing `softcore_beta`, just strip the units when you need to use it, but don't store it with implicit units.

perses/tests/test_hybrid_builder.py
jchodera: I'm not sure I have a recommendation, but we might have to adjust this later if it starts failing for systems of nontrivial size.

PR #349, opened 2017-06-21
Hybrid topology documentation

pgrinaway:
This PR will provide documentation for the `HybridTopologyEngine`.

### DIFF
--- a/docs/annihilation.rst
+++ b/docs/annihilation.rst
@@ -58,8 +58,82 @@ In order to properly assign parameters to the particles in the hybrid system, th
 - Unique new: These atoms are not in the map, and are present only in the new system
     - Example: The chlorine atom that appears when a benzene is transformed into a chlorobenzene.
     
-Creation of Force Terms
-^^^^^^^^^^^^^^^^^^^^^^^
+Treatment of different interaction groups
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 For each supported force, we have to create at least one custom force, as well as an unmodified force. In general, the interactions are handled as follows:
+
+- Environment-Environment: These interactions are handled by regular unmodified forces, as their parameters never change.
+
+- Environment-{Core, Unique old, Unique new}: These interactions are handled by Custom*Forces. For interactions involving the core atoms, the interaction scales from the parameter in the old system to the parameter in the new system as lambda is switched from 0 to 1. For interactions involving the unique atoms, the interaction scales from full to zero as lambda switches from 0 to 1 in unique old atoms, and vice versa for unique new atoms.
+
+- Core-Core interactions: These interactions scale from old system parameters to new system parameters as lambda is switched from 0 to 1.
+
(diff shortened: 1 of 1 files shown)


docs/annihilation.rst
```diff
@@ -58,8 +58,82 @@ In order to properly assign parameters to the particles in the hybrid system, th
...
+Force terms
+^^^^^^^^^^^
+
+Given the above division of atoms into different groups, the assignment of interactions to different force classes is straightforward. 
+
+
+Bonds
+"""""
+
+Bonds are handled with a HarmonicBondForce and a CustomBondForce.
```
jchodera: You might want to back-tick protect OpenMM classes like `HarmonicBondForce`

I think sphinx uses two backticks before and after: http://www.sphinx-doc.org/en/stable/rest.html

docs/annihilation.rst
```diff
@@ -136,4 +147,51 @@ Any force that is not in one of the above-mentioned classes will cause an except
...
+
+Basic Construction of Factory
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+In order to use the HybridTopologyFactory, several objects are necessary:
+
+- :py:class:`perses.rjmc.topology_proposal.TopologyProposal` is a container class which holds so-called new and old systems, topologies, and respective atom maps. It can be created manually using the constructor, or can be obtained through a subcalss of :py:class:`perses.rjmc.topology_proposal.ProposalEngine`.
+
+- You need to supply current positions and new positions for the old and new systems, respectively. It is assumed that the atoms which are mapped will have the same positions. If you only have positions for the old system, you can generate (decent) positions for the new system using :py:class:`perses.rjmc.geometry.FFAllAngleGeometryEngine`. You should minimize the hybrid topology that results after this to avoid clashes, however.
+
+- The dispersion correct is calculated for the sterics force if `use_dispersion_correction` is set to `True`. It is false by default; whether to enable this depends on what sort of calculation you are doing:
```
jchodera: I think restructured text uses two backquotes to quote code:
http://www.sphinx-doc.org/en/stable/rest.html

docs/annihilation.rst
pgrinaway: Ok, Fixing this

pgrinaway:
@jchodera when you have a chance could you take a look at this? I am not sure if there should be more detail here/what specifically might be missing. I've corrected the code quotes, but if you see one missing let me know.

jchodera:
@pgrinaway : This is great!

Are there any other assumptions made on the input `System` objects, like the order of forces, number of particles, particle ordering matching up? Might be good to document this too, but otherwise, looks good to merge!

pgrinaway:
Thanks!

>Are there any other assumptions made on the input System objects, like the order of forces, number of particles, particle ordering matching up? 

1. No assumption about force order (they are used by name)
2. The number of particles can differ, but anything not mapped will be duplicated
3. No particular order is necessary for the particles, since an atom map is required.

I think those are covered by what's in there (it explains how particles are grouped).

pgrinaway:
I'm going to go ahead and merge. Feel free to point out any other problems, and I'll fix them in a future PR.

PR #351, opened 2017-07-10
Fix bug + add feature

pgrinaway:
This PR does two things:

* Fixes a bug where the wrong range is used for looping through the bond force and angle force.

* Allows (but does not expect by default) specification of the alchemical functions in the hybrid. The advantage of this is that we can get openmm to calculate dU/dlambda easily. This would presumably also allow for alchemical functions to be defined using the OpenMM spline classes as well, which are accessible from the force side but not the integrator side.

### DIFF
--- a/perses/annihilation/new_relative.py
+++ b/perses/annihilation/new_relative.py
@@ -602,24 +602,52 @@ class HybridTopologyFactory(object):
         custom_nonbonded_method = self._translate_nonbonded_method_to_custom(self._nonbonded_method)
 
         # Create CustomNonbondedForce to handle interactions between alchemically-modified atoms and rest of system.
-        electrostatics_custom_nonbonded_force = openmm.CustomNonbondedForce("U_electrostatics;" + electrostatics_energy_expression + electrostatics_mixing_rules)
-        electrostatics_custom_nonbonded_force.addGlobalParameter("lambda_electrostatics", 0.0)
+        total_electrostatics_energy = "U_electrostatics;" + electrostatics_energy_expression + electrostatics_mixing_rules
+        if self._functions:
+            try:
+                total_electrostatics_energy += 'lambda_electrostatics = ' + self._functions['lambda_electrostatics']
+            except KeyError as e:
+                print("Functions were provided, but there is no entry for electrostatics")
+                raise e
+
+        electrostatics_custom_nonbonded_force = openmm.CustomNonbondedForce(total_electrostatics_energy)
         electrostatics_custom_nonbonded_force.addGlobalParameter("softcore_beta", self.softcore_beta)
         electrostatics_custom_nonbonded_force.addPerParticleParameter("chargeA") # partial charge initial
         electrostatics_custom_nonbonded_force.addPerParticleParameter("chargeB") # partial charge final
+
+        if self._functions:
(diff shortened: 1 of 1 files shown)


jchodera:
> Allows (but does not expect by default) specification of the alchemical functions in the hybrid. The advantage of this is that we can get openmm to calculate dU/dlambda easily. This would presumably also allow for alchemical functions to be defined using the OpenMM spline classes as well, which are accessible from the force side but not the integrator side.

Awesome! I think this would be a great feature to add to `AbsoluteAlchemicalFactory` too.

pgrinaway:
Looks like this PR is about ready to be merged.

perses/annihilation/new_relative.py
```diff
@@ -55,6 +55,10 @@ def __init__(self, topology_proposal, current_positions, new_positions, use_disp
             The positions of the "new system"
         use_dispersion_correction : bool, default False
             Whether to use the long range correction in the custom sterics force. This is very expensive for NCMC.
+        functions : dict, default None
+            Alchemical functions that determine how each force is scaled with lambda. The keys must be strings with
+            names beginning with lambda_ and ending with each of bonds, angles, torsions, sterics, electrostatics.
+            If none, then the integrator will need to set each of these and parameter derivatives will be unavailable.
```
bas-rustenburg: Do you always have to provide all lambda functions, and set the ones you don't use to `None`, or only the ones you use? Not sure what `if none` indicates.

bas-rustenburg: Ah, so it looks like you either provide all functions, or don't provide functions at all. What happens if you provide
`lambda_bonds = None`? This would probably be a bug, right?
`core_energy_expression += 'lambda_bonds = ' + self._functions['lambda_bonds']`
would be 
`core_energy_expression += 'lambda_bonds = ' + None`

bas-rustenburg: Okay, I think I misunderstood the initial explanation. You don't set any functions to None, you set the entire dict to None. You might just to say "If `functions` is not provided" instead of `if None`. I was a little confused by the change in subject in between sentences

perses/annihilation/new_relative.py
```diff
@@ -70,6 +75,12 @@ def __init__(self, topology_proposal, current_positions, new_positions, use_disp
         self.softcore_alpha=0.5
         self.softcore_beta=12*unit.angstrom**2
```
jchodera: Are you sure you want to use the fused softcore here, where the charges are softened at intermediate lambda, or did you just want to linearly change charges with `lambda`?

pgrinaway: Hm, that decision happened in the previous PR, and I assumed this strategy (which is what was in the previous hybrid implementation as well) was ok since the PR was approved. We can discuss which is better, but I think a different PR is appropriate for changing it (if necessary).

pgrinaway: I can add the units back in too, but let's only discuss the parts that changed in this PR to keep things organized. Definitely feel free to open an issue if you see a problem though!

jchodera: No need to alter this immediately, but something to keep an eye out for in case it causes problems in the future.

PR #386, opened 2017-08-31
Disable slow tests and switch to py35 and py36

pgrinaway:
Just trying to disable the slow tests. Not sure why `skipIf` isn't skipping some tests...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,6 +60,7 @@ env:
     # Location of decrypted OpenEye license file
     - OE_LICENSE="$HOME/oe_license.txt"
     - OPENEYE_CHANNEL="-i https://pypi.anaconda.org/OpenEye/simple"
+    - TRAVIS="true"
     # encrypted BINSTAR_TOKEN for push of dev package to binstar
     - secure: "RRvLDPu9mPoNaRWIseaJdgShOXI+PaHPWKAIJvW7VYWcAS6iEN7W4Fj4zD5hkocQxc3ou97EtkgID+ApH10bSGKxCykyU0urSY9jsSOJX2m0AE19X0dVr6ySIQkwCWE6kUMVlvQYQo80fM2EMElD+btr4G9XBAhArAO7HvZHkoQ="
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ branches:
 install:
   #add random sleep from 1-10s to try to prevent overloading the anaconda servers
   - sleep $[ ( $RANDOM % 10 )  + 1 ]s
+  - env
   - source devtools/travis-ci/install.sh
   - export PYTHONUNBUFFERED=true
   - export CC=gcc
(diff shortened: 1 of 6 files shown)


pgrinaway:
Looks like the reason those slow tests weren't being disabled was because the conda build was failing, and it was installing + testing an old version (thanks for discovering this @jchodera). We've removed the packages from the cloud, and have commented out the command to push to the cloud.

pgrinaway:
See #387 for conda issue discussion.

jchodera:
Looks like you fixed it by pinning `conda-build` to 2.1.17! Tests are running to completion (though it looks like some tests have caught issues). 

Those issues should probably be addressed in separate PRs as needed.

It may be worth dropping python 2.7 and 3.4 and adding 3.6 so that we only support 3.5 and 3.6 going forward. That will simplify the code.

jchodera:
Note, for example, that new `openmmtools` releases are no longer built for Python 3.4, so we can no longer support it here either.

pgrinaway:
>It may be worth dropping python 2.7 and 3.4 and adding 3.6 so that we only support 3.5 and 3.6 going forward. That will simplify the code.

Agreed, I'll do that now.

pgrinaway:
@jchodera In this PR I've:

* Pinned the conda build version to the same one as `yank`
* Changed the environments to include only Python 3.5 and 3.6
* Unpinned numpy (which does not seem to be giving a problem)

If you could take a quick pass through and make sure that nothing seems terribly wrong, that would be great. Thanks!

pgrinaway:
cc: #387 , resolved by this PR.

pgrinaway:
Hm, nope, seems py36 is still failing.

pgrinaway:
Oh, I need to unpin another one.

jchodera:
The tests seem to be running, but the NCMC tests also seem to be running---was that intended?
```
Testing VV NCMC switching using harmonic oscillator with 0 NCMC steps ...    0.043 +- 0.156 kT
ok (34.4397s)
Testing VV NCMC switching using harmonic oscillator with 1 NCMC steps ...    0.133 +- 0.173 kT
ok (34.4379s)
Testing VV NCMC switching using harmonic oscillator with 50 NCMC steps ...   -0.009 +- 0.017 kT
ok (48.3885s)
Testing GHMC NCMC switching using harmonic oscillator with 0 NCMC steps ...   -0.314 +- 0.144 kT
ok (34.6554s)
Testing GHMC NCMC switching using harmonic oscillator with 1 NCMC steps ... 
```

jchodera:
It's [this test](https://github.com/choderalab/perses/blob/3f4b744403cd11626959ab43627cecf94107b54d/perses/tests/test_ncmc_integrator.py#L142-L151) that's still running NCMC tests to test `NCMCVVAlchemicalIntegrator` and `NCMCGHMCAlchemicalIntegrator`. Do you still use those in the current application? If not, maybe add the travis skip there too?

pgrinaway:
Ok, looks like the errors are currently unrelated (will raise issue to document), and this is complete.

PR #389, opened 2017-09-01
Remove temperature defined at top-level of application file

jchodera:
The `relative_setup.py` file defined `temperature`, `kT`, and `beta` at the top-level of the file, and used this top-level `beta` but a function argument `temperature` (could be different) potentially leading to contradictory temperatures. This eliminates top-level temperature definitions.

Note that I can't find anywhere the function `generate_vacuum_hybrid_topology` is either used or tested in `perses`. Should this be deleted?


pgrinaway:
Yes, that should be deleted. That was just there for quick testing.

jchodera:
Remove in a separate PR?

pgrinaway:
Sounds good.

PR #391, opened 2017-09-01
Reset NCMC statistics before nonequilibrium switching

### LINKED ISSUE #368: Reset nonequilibrium integrators before use in nonequilibrium tasklet
Nonequilibrium integrators need to call `integrator.reset()` before switching in order to work since they are being retrieved from a cached `Context`.

jchodera:
Fixes #368.

Since #1870, the development version of OpenMM has `Context.getIntegrator()` return the correct Python subclass, so the `restore_properties()` workaround should no longer be needed (and would also interfere with the ability to call `integrator.reset()`).

We may also have to update `openmmtools.ContextCache()` to streamline the `RestorableIntegrator` scheme, but this will only work with OpenMM 7.2 and later, so we should hold off on that if we can.

Docstrings have also been clarified.

### DIFF
--- a/perses/distributed/feptasks.py
+++ b/perses/distributed/feptasks.py
@@ -87,11 +87,10 @@ def run_protocol(self, starting_positions, nsteps, thermodynamic_state, integrat
     """
     integrator = restore_properties(integrator)
     switching_ctx, integrator_neq = self._cache.get_context(thermodynamic_state, integrator)
-    integrator_neq = restore_properties(integrator_neq)
     switching_ctx.setPositions(starting_positions)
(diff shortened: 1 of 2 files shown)


jchodera:
Can you double check this branch runs locally with the latest openmm dev version? Or does the Travis testing actually covers the application?

pgrinaway:
>Can you double check this branch runs locally with the latest openmm dev version? Or does the Travis testing actually covers the application?

Will check.

pgrinaway:
Looks like this runs fine locally. Merging.

PR #396, opened 2017-09-05
Switch to using MCMove framework in feptasks.py

pgrinaway:
In this PR, I'm switching the `feptasks.py` tasklets to use the MCMove framework. This will also involve some substantial changes to `relative_setup.py` in order to accommodate the new (and cleaner, I think) API.

### DIFF
--- a/perses/distributed/feptasks.py
+++ b/perses/distributed/feptasks.py
@@ -21,6 +21,7 @@ import openmmtools.integrators as integrators
 import redis
 import numpy as np
 import mdtraj as md
+import mdtraj.utils as mdtrajutils
 import pickle
 
 broker_name_server = "redis://localhost"
@@ -209,6 +210,8 @@ def run_protocol(thermodynamic_state, sampler_state, ne_mc_move, topology, n_ite
 
     #create a numpy array for the trajectory
     trajectory_positions = np.zeros([n_iterations, n_atoms, 3])
+    trajectory_box_lengths = np.zeros([n_iterations, 3])
+    trajectory_box_angles = np.zeros([n_iterations, 3])
 
     #create a numpy array for the work values
     cumulative_work = np.zeros(n_iterations)
@@ -224,11 +227,16 @@ def run_protocol(thermodynamic_state, sampler_state, ne_mc_move, topology, n_ite
         #record the positions as a result
         trajectory_positions[iteration, :, :] = sampler_state.positions
 
+        #get the box angles and lengths
+        a, b, c, alpha, beta, gamma = mdtrajutils.unitcell.box_vectors_to_lengths_and_angles(sampler_state.box_vectors)
+        trajectory_box_lengths[iteration, :] = [a, b, c]
+        trajectory_box_angles[iteration, :] = [alpha, beta, gamma]
+
         #record the cumulative work as a result
         cumulative_work[iteration] = ne_mc_move.current_total_work
 
     #create an MDTraj trajectory with this data
-    trajectory = md.Trajectory(trajectory_positions, topology)
+    trajectory = md.Trajectory(trajectory_positions, topology, unitcell_lengths=trajectory_box_lengths, unitcell_angles=trajectory_box_angles)
 
     return trajectory, cumulative_work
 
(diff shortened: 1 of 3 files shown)


pgrinaway:
@jchodera Note that the commits up to this point are modifying the tasklets. I will now modify the driver to suit the new API.

jchodera:
Cool! Is your idea to save a frame each time a `MCMCMove` is applied, so that changing the save interval is just a matter of adjusting the number of steps in your move and how many times you want to apply it each tasklet?

pgrinaway:
>Cool! Is your idea to save a frame each time a MCMCMove is applied, so that changing the save interval is just a matter of adjusting the number of steps in your move and how many times you want to apply it each tasklet?

Yes, exactly!

perses/distributed/relative_setup.py
```diff
@@ -486,8 +486,8 @@ def __init__(self, topology_proposal, pos_old, new_positions, use_dispersion_cor
         self.minimize()
 
         #initialize the trajectories for the lambda 0 and 1 equilibrium simulations
-        self._lambda_zero_traj = md.Trajectory(self._lambda_zero_sampler_state.positions, self._factory.hybrid_topology)
-        self._lambda_one_traj = md.Trajectory(self._lambda_one_sampler_state.positions, self._factory.hybrid_topology)
+        self._lambda_zero_traj = md.Trajectory(np.array(self._lambda_zero_sampler_state.positions), self._factory.hybrid_topology)
```
jchodera: I'm not sure you should do this. Shouldn't `sampler_state` automatically store things in `simtk.unit.Quantity` wrapped numpy arrays? What you're doing here runs the risk of silently dropping units.

If you need to convert from Quantity-wrapped numpy arrays to nanometers for MDTraj, use an explicit unit conversion:
```python
positions / unit.nanometers
```
If sampler_state isn't storing things in Quantity-wrapped numpy arrays, let's fix that in `openmmtools.mcmc`.

pgrinaway: It looks like it's stored as a unit bearing list, which MDTraj apparently can't handle.

jchodera: For now, what about dealing with both contingencies (`Quantity`-wrapped `list` or `numpy.array`) with a `try...except` clause?

pgrinaway:
>Do we really want the tasklets to write disk directly? 

They aren't writing to disk directly...

pgrinaway:
If you take a closer look, you'll see that the part writing to disk is a driver class.

pgrinaway:
Currently dealing with [this issue](https://github.com/choderalab/openmmtools/issues/289)

jchodera:
> If you take a closer look, you'll see that the part writing to disk is a driver class.

Whoops! You're right---the changes had just folded in a weird way on mobile, and I missed that!

perses/distributed/relative_setup.py
```diff
@@ -307,135 +314,350 @@ class NonequilibriumSwitchingFEP(object):
...
+        if self._ncmc_nsteps % self._nsteps_per_iteration != 0:
+            logging.warning("The number of ncmc steps is not divisible by the number of steps per iteration. You may not have a full protocol.")
+        self._n_iterations_per_call = self._ncmc_nsteps // self._nsteps_per_iteration
+
+        #create the thermodynamic state
+        lambda_zero_alchemical_state = alchemy.AlchemicalState.from_system(self._hybrid_system)
+        lambda_one_alchemical_state = copy.deepcopy(lambda_zero_alchemical_state)
+
+        #ensure their states are set appropriately
+        lambda_zero_alchemical_state = self._set_all_parameters_of_state(lambda_zero_alchemical_state, 0.0)
```
andrrizzi: You should call `lambda_zero_alchemical_state.set_alchemical_parameters(0.0)`. I think you'll get an error later if you set to `0.0` a lambda parameter that is not defined in the `System` (like `lambda_bonds` in most cases). Parameters that are not defined in the system should be `None` (the rationale for the error is to make it clear that you're not modifying anything even if you change the state's parameter).

perses/distributed/feptasks.py
```diff
@@ -20,7 +39,129 @@ def get_broker_name():
...
+        if self.context_cache is None:
+            context_cache = cache.global_context_cache
+        else:
+            context_cache = self.context_cache
+
+        #Get the integrator from the context cache
+        context, integrator = context_cache.get_context(thermodynamic_state, self._integrator)
+
+        #Reset the statistics on the integrator
+        integrator.reset()
```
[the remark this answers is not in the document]
pgrinaway: Ok, great!

perses/distributed/feptasks.py
```diff
@@ -20,7 +39,129 @@ def get_broker_name():
...
+
+        Parameters
+        ----------
+        context : openmm.Context
+            The OpenMM context which is performing the integration
+        thermodynamic_state : openmmtools.states.ThermodynamicState
+            The relevant thermodynamic state for this context and integrator
+        """
+        integrator = context.getIntegrator()
+        self._current_total_work += integrator.get_total_work(dimensionless=True)
```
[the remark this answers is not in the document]
pgrinaway: Oh, good call.

perses/distributed/relative_setup.py
pgrinaway: ok, will do. I have all of those parameters, but to be safe I'll use the proper API.

pgrinaway:
Thanks!

PR #397, opened 2017-09-11
Fix topology creation to use the residue from the hybrid topology

pgrinaway:
This problem did not affect actual system creation, but resulted in the
MDTraj topology that was generated being unable to be deepcopied. To
resolve it, for each unique new atom that gets added, we find the
corresponding residue and extract it from the hybrid, then use that to
add the new atom.


pgrinaway:
Thanks! Merging (small PR)

PR #398, opened 2017-09-12
Faster constraint check

### LINKED ISSUE #363: Exclude hydrogens from being mapped
In order to allow us to run `perses` with bonds to hydrogen constrained, we are forced to exclude hydrogens from mapped atoms so that they are always `unique`. Otherwise, we would not be able to permit constrained `H-X` bonds of different lengths in initial and final states.

pgrinaway:
This PR updates the constraint check to use the system XMLs and lxml; this is much faster (on the order of 15sec instead of 15min) and allows us to re-introduce the constraint check. This should resolve #363 by raising an exception if a constraint is changing.

### DIFF
--- a/perses/annihilation/new_relative.py
+++ b/perses/annihilation/new_relative.py
@@ -410,6 +410,16 @@ class HybridTopologyFactory(object):
                 if constraint != constraint_from_old_system:
                     raise ValueError("Constraints are changing during switching.")
 
+    def _constraint_check_fast(self):
+        """
+        This method will check for changing constraints by first serializing the new and old systems to xml, then using
+        that xml to check for constraint changes. Using lxml and XPATH, this should be considerably faster than the
+        OpenMM API. If a constraint is found to be changing, an exception will be raised, as this cannot currently be
+        handled by the HybridTopologyFactory.
+        """
+        old_system_xml = openmm.XmlSerializer.serialize(self._topology_proposal.old_system)
+        new_system_xml = openmm.XmlSerializer.serialize(self._topology_proposal.new_system)
+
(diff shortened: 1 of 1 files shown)


perses/annihilation/new_relative.py
```diff
@@ -410,6 +412,50 @@ def _constraint_check(self):
...
+        old_system_tree = etree.parse(old_system_io)
+        new_system_tree = etree.parse(new_system_io)
+
+        #get the list of constraints from new and old systems:
+        old_system_constraint_list = old_system_tree.xpath(constraint_string)
+        new_system_constraint_list = new_system_tree.xpath(constraint_string)
+
+        #convert the list of constraint elements to dictionaries. By using frozenset, we can do this independent of the order of
+        old_system_constraints = {frozenset((o_h_map[int(constraint.attrib['p1'])], o_h_map[int(constraint.attrib['p2'])])) : float(constraint.attrib['d']) for constraint in old_system_constraint_list}
+        new_system_constraints = {frozenset((n_h_map[int(constraint.attrib['p1'])], n_h_map[int(constraint.attrib['p2'])])) : float(constraint.attrib['d']) for constraint in new_system_constraint_list}
```
andrrizzi: I think it should be safe to use simply a `tuple` instead of a `frozenset` for the keys.

andrrizzi: Ah! Or is it to avoid the case when you have duplicate `(p1, p2)` and `(p2, p1)` pairs?

pgrinaway: But if I do that, then the order of the indices will matter, right?

pgrinaway: Oops, didn't see your other message.

Issue #400, opened 2017-09-18
redis output buffer limit error in distributed module


brycestx:
I am experiencing an error where the celery tasks stall when running [relative_setup.py](https://github.com/choderalab/perses/blob/master/perses/distributed/relative_setup.py) on the complex topology (see attached files). 
[relative_setup_error_091817.zip](https://github.com/choderalab/perses/files/1312207/relative_setup_error_091817.zip)

Redis outputs a message:

```
29179:M 18 Sep 16:08:54.168 # Client id=28 addr=127.0.0.1:38640 fd=23 name= age=77 idle=56 flags=N db=0 sub=14 psub=0 multi=-1 qbuf=0 qbuf-free=0 obl=88 oll=1 omem=33554472 events=r cmd=subscribe scheduled to be closed ASAP for overcoming of output buffer limits.
29179:M 18 Sep 16:09:05.823 # Client id=30 addr=127.0.0.1:38646 fd=23 name= age=7 idle=7 flags=N db=0 sub=14 psub=0 multi=-1 qbuf=0 qbuf-free=0 obl=88 oll=1 omem=33554472 events=r cmd=subscribe scheduled to be closed ASAP for overcoming of output buffer limits.
29179:M 18 Sep 16:21:29.887 # Client id=32 addr=127.0.0.1:38650 fd=23 name= age=741 idle=741 flags=N db=0 sub=14 psub=0 multi=-1 qbuf=0 qbuf-free=0 obl=89 oll=1 omem=167772200 events=r cmd=subscribe scheduled to be closed ASAP for overcoming of output buffer limits.
29179:M 18 Sep 16:21:39.433 # Client id=34 addr=127.0.0.1:38750 fd=23 name= age=10 idle=10 flags=N db=0 sub=14 psub=0 multi=-1 qbuf=0 qbuf-free=0 obl=89 oll=1 omem=167772200 events=r cmd=subscribe scheduled to be closed ASAP for overcoming of output buffer limits.
29179:M 18 Sep 16:23:54.955 # Client id=36 addr=127.0.0.1:38754 fd=23 name= age=135 idle=135 flags=N db=0 sub=14 psub=0 multi=-1 qbuf=0 qbuf-free=0 obl=89 oll=1 omem=167772200 events=r cmd=subscribe scheduled to be closed ASAP for overcoming of output buffer limits.
[...]

PR #402, opened 2017-09-20
Strip water from trajectories returned by tasklets

pgrinaway:
Should address #400 by requiring significantly less memory per trajectory.

### DIFF
--- a/perses/distributed/feptasks.py
+++ b/perses/distributed/feptasks.py
@@ -207,7 +207,7 @@ def run_protocol(thermodynamic_state, sampler_state, ne_mc_move, topology, n_ite
 
     """
     #get the atom indices we need to subset the topology and positions
-    atom_indices = topology.select('not solvent')
+    atom_indices = topology.select('not water')
     subset_topology = topology.subset(atom_indices)
 
     n_atoms = subset_topology.n_atoms
@@ -276,7 +276,7 @@ def run_equilibrium(thermodynamic_state, sampler_state, mc_move, topology, n_ite
         Unitless reduced potential (kT) of the final frame of the trajectory
     """
     #get the atom indices we need to subset the topology and positions
-    atom_indices = topology.select('not solvent')
+    atom_indices = topology.select('not water')
     subset_topology = topology.subset(atom_indices)
 
     n_atoms = subset_topology.n_atoms
--- a/perses/distributed/feptasks.py
+++ b/perses/distributed/feptasks.py
@@ -181,7 +181,7 @@ def update_broker_location(broker_location, backend_location=None):
     app.conf.update(broker=broker_location, backend=broker_location)
 
 @app.task(serializer="pickle")
-def run_protocol(thermodynamic_state, sampler_state, ne_mc_move, topology, n_iterations):
+def run_protocol(thermodynamic_state, sampler_state, ne_mc_move, topology, n_iterations, atom_indices_to_save=None):
     """
     Perform a nonequilibrium switching protocol and return the nonequilibrium protocol work. Note that it is expected
     that this will perform an entire protocol, that is, switching lambda completely from 0 to 1, in increments specified
@@ -199,6 +199,8 @@ def run_protocol(thermodynamic_state, sampler_state, ne_mc_move, topology, n_ite
         An MDtraj topology for the system to generate trajectories
     n_iterations : int
         The number of times to apply the specified MCMove
+    atom_indices_to_save : list of int, default None
+        list of indices to save (when excluding waters, for instance). If None, all indices are saved.
 
     Returns
     -------
@@ -207,8 +209,12 @@ def run_protocol(thermodynamic_state, sampler_state, ne_mc_move, topology, n_ite
 
     """
     #get the atom indices we need to subset the topology and positions
-    atom_indices = topology.select('not water')
-    subset_topology = topology.subset(atom_indices)
+    if atom_indices_to_save is None:
+        atom_indices = range(topology.n_atoms)
+        subset_topology = topology
+    else:
+        subset_topology = topology.subset(atom_indices_to_save)
+        atom_indices = atom_indices_to_save
 
     n_atoms = subset_topology.n_atoms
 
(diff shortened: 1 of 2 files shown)


perses/distributed/feptasks.py
```diff
@@ -206,7 +206,11 @@ def run_protocol(thermodynamic_state, sampler_state, ne_mc_move, topology, n_ite
         Trajectory containing n_iterations frames
 
     """
-    n_atoms = topology.n_atoms
+    #get the atom indices we need to subset the topology and positions
+    atom_indices = topology.select('not water')
```
jchodera: Make the atom indices a parameter to make this tasklet more general. You can do the topology selection just once in the main driver.

perses/distributed/feptasks.py
```diff
@@ -225,7 +229,7 @@ def run_protocol(thermodynamic_state, sampler_state, ne_mc_move, topology, n_ite
         ne_mc_move.apply(thermodynamic_state, sampler_state)
 
         #record the positions as a result
-        trajectory_positions[iteration, :, :] = sampler_state.positions
+        trajectory_positions[iteration, :, :] = sampler_state.positions[atom_indices, :]
```
jchodera: Isn't `sampler_state.positions` a unit-bearing quantity that you are forcing into a numpy array (`trajectory_positions`) without explicit unit conversion? This will break if `SamplerState` default units ever change. Can you at least explicitly convert all these quantities into the appropriate MDTraj units explicitly?

jchodera: Actually, it may be even better to add a method to `SamplerState` like `.to_mdtraj_trajectory(topology, atom_indices=None)` where it will automatically populate a `Topology` object for you using the specified atom indices.

perses/distributed/feptasks.py
```diff
@@ -271,7 +275,11 @@ def run_equilibrium(thermodynamic_state, sampler_state, mc_move, topology, n_ite
     reduced_potential_final_frame : float
         Unitless reduced potential (kT) of the final frame of the trajectory
     """
-    n_atoms = topology.n_atoms
+    #get the atom indices we need to subset the topology and positions
+    atom_indices = topology.select('not water')
```
jchodera: Just take atom indices as an argument (defaulting to `None`, which means all atoms).

perses/distributed/relative_setup.py
```diff
@@ -399,8 +399,14 @@ def __init__(self, topology_proposal, pos_old, new_positions, use_dispersion_cor
...
 
-        self._lambda_zero_traj = md.Trajectory(np.array(self._lambda_zero_sampler_state.positions), self._factory.hybrid_topology, unitcell_lengths=[a_0, b_0, c_0], unitcell_angles=[alpha_0, beta_0, gamma_0])
-        self._lambda_one_traj = md.Trajectory(np.array(self._lambda_one_sampler_state.positions), self._factory.hybrid_topology, unitcell_lengths=[a_1, b_1, c_1], unitcell_angles=[alpha_1, beta_1, gamma_1])
+        #subset the topology appropriately:
+        atom_selection = self._factory.hybrid_topology.select("not water")
+        subset_topology = self._factory.hybrid_topology.subset(atom_selection)
+        lambda_zero_positions = self._lambda_zero_sampler_state.positions[atom_selection, :]
+        lambda_one_positions = self._lambda_one_sampler_state.positions[atom_selection, :]
+
+        self._lambda_zero_traj = md.Trajectory(np.array(lambda_zero_positions), subset_topology, unitcell_lengths=[a_0, b_0, c_0], unitcell_angles=[alpha_0, beta_0, gamma_0])
```
jchodera: Don't implicitly strip units from `sampler_state.positions` this way---explicitly convert them to MDTraj units.

jchodera:
We should probably get this cleaned up and merged!

perses/distributed/feptasks.py
```diff
@@ -199,14 +200,24 @@ def run_protocol(thermodynamic_state, sampler_state, ne_mc_move, topology, n_ite
...
     Returns
     -------
     trajectory : mdtraj.Trajectory
         Trajectory containing n_iterations frames
 
     """
-    n_atoms = topology.n_atoms
+    #get the atom indices we need to subset the topology and positions
+    if atom_indices_to_save is None:
+        atom_indices = range(topology.n_atoms)
```
bas-rustenburg: Do you want a `range` object here (python3 behavior)? The user would normally provide a list. Perhaps it doesn't make a difference, but just want to make sure to convert to list, or face a potential bug.

pgrinaway: Thanks, good call

perses/distributed/feptasks.py
```diff
@@ -271,7 +284,15 @@ def run_equilibrium(thermodynamic_state, sampler_state, mc_move, topology, n_ite
     reduced_potential_final_frame : float
         Unitless reduced potential (kT) of the final frame of the trajectory
     """
-    n_atoms = topology.n_atoms
+    #get the atom indices we need to subset the topology and positions
+    if atom_indices_to_save is None:
+        atom_indices = range(topology.n_atoms)
```
bas-rustenburg: Same comment as above, cast to list?

perses/distributed/relative_setup.py
```diff
@@ -312,7 +312,7 @@ class NonequilibriumSwitchingFEP(object):
 
     def __init__(self, topology_proposal, pos_old, new_positions, use_dispersion_correction=False,
                  forward_functions=None, ncmc_nsteps=100, nsteps_per_iteration=1, concurrency=4, platform_name="OpenCL",
-                 temperature=300.0 * unit.kelvin, trajectory_directory=None, trajectory_prefix=None):
+                 temperature=300.0 * unit.kelvin, trajectory_directory=None, trajectory_prefix=None, atom_selection="not water"):
```
bas-rustenburg: Is the atom_selection an md_traj selection string of some kind? Might be helpful to add a comment about it, or perhaps add to docstring.

pgrinaway: Ok, adding docstring. Just noticed it was missing. There's an issue open to document this class better, but will add docstring here and address other docs in another PR.

jchodera:
Do you really want to do this for the equilibrium trajectories? You need the fully solvated snapshots with box vectors to shoot new equilibrium or nonequilibrium trajectories.

pgrinaway:
>Do you really want to do this for the equilibrium trajectories? You need the fully solvated snapshots with box vectors to shoot new equilibrium or nonequilibrium trajectories.


Do what? Can you point to the code where this happens? I don't use the trajectories to restart equilibrium or nonequilibrium calculations.

Issue #403, opened 2017-09-22
new relative perturbation system NaN and index errors


brycestx:
I have been trying new perturbations from the JACS paper with the same CDK2 system in [relative_setup.py](https://github.com/choderalab/perses/blob/master/perses/distributed/relative_setup.py) and have received two errors which I believe to be related after running through a variety of perturbations (see below).

When defining molecule indices as 0 and 1 in [NonequilibriumFEPSetup](https://github.com/choderalab/perses/blob/master/perses/distributed/relative_setup.py#L679) function of relative_setup.py I received a NaN error:
```
[2017-09-21 18:44:42,096: ERROR/ForkPoolWorker-1] Potential energy is NaN after 0 attempts of integration with move LangevinSplittingDynamicsMove
[2017-09-21 18:44:42,100: ERROR/ForkPoolWorker-1] Task perses.distributed.feptasks.run_equilibrium[c323162f-8b98-482f-b125-21a203802aa8] raised unexpected: IntegratorMoveError('Potential energy is NaN after 0 attempts of integration with move LangevinSplittingDynamicsMove',)
Traceback (most recent call last):
  File "/home/ballen/anaconda2/envs/oepython3/lib/python3.6/site-packages/celery/app/trace.py", line 374, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/home/ballen/anaconda2/envs/oepython3/lib/python3.6/site-packages/celery/app/trace.py", line 629, in __protected_call__
    return self.run(*args, **kwargs)
  File "/home/ballen/github/perses/perses/distributed/feptasks.py", line 291, in run_equilibrium
    mc_move.apply(thermodynamic_state, sampler_state)
  File "/home/ballen/anaconda2/envs/oepython3/lib/python3.6/site-packages/openmmtools-0.13.1-py3.6.egg/openmmtools/mcmc.py", line 1103, in apply
    super(LangevinDynamicsMove, self).apply(thermodynamic_state, sampler_state)
  File "/home/ballen/anaconda2/envs/oepython3/lib/python3.6/site-packages/openmmtools-0.13.1-py3.6.egg/openmmtools/mcmc.py", line 703, in apply
    raise IntegratorMoveError(err_msg, self, context)
openmmtools.mcmc.IntegratorMoveError: Potential energy is NaN after 0 attempts of integration with move LangevinSplittingDynamicsMove
[2017-09-21 18:44:47,255: INFO/MainProcess] missed heartbeat from worker1@alchemist
[...]

PR #404, opened 2017-09-25
Fix typo in atom map

pgrinaway:
Addresses the second bug in #403


jchodera:
Add a test which would have caught this bug?

pgrinaway:
Test implemented.

Issue #362, opened 2017-08-18
Ensure core atoms are correctly identified when there are no unique atoms


pgrinaway:
In transformations such as 1-fluorobenzene to 1-chlorobenzene, we are unclear as to whether the current atom-matching scheme would flag there the ligand residue does not have any atoms that are deleted or added, resulting in no `unique` atoms found. We should refine the logic used to identify `environment` atoms to make sure these corner cases are correctly handled.

pgrinaway:
We should move the determination of whether an atom is in core or environment to the atom mapping facility to avoid this problem. We can also add a second stage to de-map atoms that would change a constraint length.

jchodera:
I'm diving into this now!

jchodera:
Pinging this thread to note the remainder of this solution needs to be implemented now that #420 is in.

jchodera:
Fixed by naming residues `MOL`

PR #420, opened 2017-10-04
Better handling of core/environment atoms in small molecule atom mappings

jchodera:
Addresses part of https://github.com/choderalab/perses/issues/362

- [x] Update MCSS rules
- [x] Add support to`TopologyProposal` for identifying environment/alchemical atoms
- [x] Have the atom mapping routines identify environment/alchemical atoms

### DIFF
--- a/perses/rjmc/topology_proposal.py
+++ b/perses/rjmc/topology_proposal.py
@@ -1244,8 +1244,8 @@ class SmallMoleculeSetProposalEngine(ProposalEngine):
         self.atom_expr = atom_expr or DEFAULT_ATOM_EXPRESSION
         self.bond_expr = bond_expr or DEFAULT_BOND_EXPRESSION
 
-        list_of_smiles = list(set(list_of_smiles))
-        self._smiles_list = [self._canonicalize_smiles(smiles) for smiles in list_of_smiles]
+        # Canonicalize all SMILES strings
+        self._smiles_list = [self._canonicalize_smiles(smiles) for smiles in set(list_of_smiles)]
         self._n_molecules = len(self._smiles_list)
 
         self._residue_name = residue_name
@@ -1279,7 +1279,6 @@ class SmallMoleculeSetProposalEngine(ProposalEngine):
         proposal : TopologyProposal object
            topology proposal object
         """
-        self.verbose = True
         current_mol_smiles, current_mol = self._topology_to_smiles(current_topology)
         current_receptor_topology = self._remove_small_molecule(current_topology)
         old_mol_start_index, len_old_mol = self._find_mol_start_index(current_topology)
--- a/perses/rjmc/topology_proposal.py
+++ b/perses/rjmc/topology_proposal.py
@@ -677,8 +677,7 @@ class PolymerProposalEngine(ProposalEngine):
         oegraphmol_proposed = oechem.OEGraphMol(proposed_molecule)
         mcs = oechem.OEMCSSearch(oechem.OEMCSType_Exhaustive)
 
-        # Select atom mapping scheme
-        atomexpr = oechem.OEExprOpts_Aromaticity | oechem.OEExprOpts_RingMember | oechem.OEExprOpts_HvyDegree
+        atomexpr = oechem.OEExprOpts_Aromaticity | oechem.OEExprOpts_RingMember | oechem.OEExprOpts_HvyDegree | oechem.OEExprOpts_AtomicNumber
         bondexpr = oechem.OEExprOpts_Aromaticity | oechem.OEExprOpts_RingMember
         mcs.Init(oegraphmol_current, atomexpr, bondexpr)
 
--- a/perses/rjmc/topology_proposal.py
+++ b/perses/rjmc/topology_proposal.py
@@ -163,7 +163,7 @@ class TopologyProposal(object):
         self._old_to_new_atom_map = {old_atom : new_atom for new_atom, old_atom in new_to_old_atom_map.items()}
         self._unique_new_atoms = list(set(range(self._new_topology._numAtoms))-set(self._new_to_old_atom_map.keys()))
         self._unique_old_atoms = list(set(range(self._old_topology._numAtoms))-set(self._new_to_old_atom_map.values()))
-        self._old_alchemical_atoms = old_alchemical_atoms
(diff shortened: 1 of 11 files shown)


jchodera:
I'm waiting to merge #426 before wrapping this up, but should be able to finish it today.

jchodera:
@pgrinaway : To have the atom mapping identify core/environment atoms, we need to add more information to the `topology_proposal`, which will necessitate specifying this information in anything that creates a `TopologyProposal`. What's the best way to do this? 

The minimal amount of information needed to make things easy is likely to specify one additional atom map which is either `environment_atom_mapping` (from old to new), or `transformation_atom_mapping` (which is just non-environment atoms). 

If we add `environment_atom_mapping`, would we want the existing `atom_mapping` map to still hold *all* the atom mapping (so that we don't need to change any existing code, but we can use `environment_atom_mapping` where handy to do so)? Or would we want to split this into environment- and non-environment atom maps for convenience of downstream processing (at the cost of updating code in a few places)?

jchodera:
I guess we can simply have all three available:
* complete atom map
* environment atom map
* non-environment atom map

All we have to do is to pass in some information about which old or new atoms are the environment when creating the `TopologyProposal` (which is easy) and it can construct all of these convenience maps automatically.

perses/rjmc/topology_proposal.py
```diff
@@ -1177,6 +1180,9 @@ def build_system(self, new_topology):
         new_system : openmm.System
             A system object generated from the topology
         """
+        if self.verbose: print('Generating System...')
```
pgrinaway: Can we use logging? I realize it's not the best thing, but it's better than `print` because that way the user can control the verbosity level very easily.

perses/rjmc/topology_proposal.py
```diff
@@ -1474,6 +1451,9 @@ def _build_new_topology(self, current_receptor_topology, oemol_proposed):
         mol_start_index : int
             The first index of the small molecule
         """
+        if self.verbose: print('Building new Topology object...')
```
pgrinaway: logging here too. Maybe level `INFO`?

perses/rjmc/topology_proposal.py
```diff
@@ -1483,6 +1463,8 @@ def _build_new_topology(self, current_receptor_topology, oemol_proposed):
         if current_receptor_topology._periodicBoxVectors != None:
             new_topology._periodicBoxVectors = copy.deepcopy(current_receptor_topology._periodicBoxVectors)
 
+        if self.verbose: print('Topology generation took %.3f s' % (time.time() - timer_start))
```
pgrinaway: logging `INFO`?

perses/rjmc/topology_proposal.py
```diff
@@ -1541,6 +1532,8 @@ def _get_mol_atom_map(current_molecule, proposed_molecule, atom_expr=oechem.OEEx
             old_index = matchpair.pattern.GetIdx()
             new_index = matchpair.target.GetIdx()
             new_to_old_atom_map[new_index] = old_index
+
+        if verbose: print('Atom map took %.3f s' % (time.time() - timer_start))
```
pgrinaway: `INFO` here as well`

jchodera:
Thanks for converting these to logging! I had intended to do that after the logging PR was merged.

jchodera:
@pgrinaway : Did you want me to finish up this PR or did you want to finish it yourself?

jchodera:
I'll get this finished up today. Apologies for the delay!

pgrinaway:
> If we add environment_atom_mapping, would we want the existing atom_mapping map to still hold all the atom mapping (so that we don't need to change any existing code, but we can use environment_atom_mapping where handy to do so)? Or would we want to split this into environment- and non-environment atom maps for convenience of downstream processing (at the cost of updating code in a few places)?

I think retaining all atoms in the atom map would be a good thing, since that way we don't have to refactor anything else. I don't think the cost of adding an extra `dict` to the `TopologyProposal` is very high (though if it turns out to be we can certainly revisit this)

pgrinaway:
>I'll get this finished up today. Apologies for the delay!

No worries!

jchodera:
@pgrinaway : I've added support for specifying alchemical/environment atoms to `TopologyProposal`:
https://github.com/choderalab/perses/pull/420/files#diff-bd95c4342ca1beb79ba546696bfdc262R121

For convenience, you only have to specify `old_alchemical_atoms` and the rest are computed.

Do we want to make this *optional*, so that all atoms are assumed to be potentially alchemically modified by default (the current behavior)? Or do we want to make it *mandatory* where an exception is raised if no list of `old_alchemical_atoms` is specified?

The following properties are added, all of which return list copies of the relevant atom indices:
```
+    new_alchemical_atoms : list
 +        List of all atoms in new system that are being transformed
 +    new_environment_atoms : list
 +        List of all atoms in new system that are not transformed, just mapped
 +    old_alchemical_atoms : list
 +        List of all atoms in old system that are being transformed
 +    old_environment_atoms : list
 +        List of all atoms in old system that are not transformed, just mapped
```

jchodera:
I've indeed made this optional so all atoms are assumed to be alchemically modified by default unless the proposal engine specifies `old_alchemical_atoms`. For `SmallMoleculeSetProposalEngine`, this is trivial to provide, so I've added this. We still have to modify the downstream methods to make use of this information to speed things up, but we can do that later.

pgrinaway:
>I've indeed made this optional so all atoms are assumed to be alchemically modified by default unless the proposal engine specifies old_alchemical_atoms. For SmallMoleculeSetProposalEngine, this is trivial to provide, so I've added this. We still have to modify the downstream methods to make use of this information to speed things up, but we can do that later.

Alright, that sounds good.

jchodera:
@pgrinaway : Let me handle the constrained atoms exclusion in a separate PR because it's going to take a bit more time to build something around `_constraint_check_fast` to minimize code reuse.

I'm super surprised that this is much faster than using the OpenMM API. Any idea why this might be?

jchodera:
@pgrinaway Can you review? I'll tackle the remainder of #362 (don't map atoms that change constraint status or length) in a separate PR so we can get this one merged.

PR #423, opened 2017-10-04
Change print statements to logger statements

pgrinaway:
The most verbose stuff is relegated to the level "debug," and the rest
is in "INFO." The logger name used is "geometry."

### DIFF
--- a/perses/rjmc/geometry.py
+++ b/perses/rjmc/geometry.py
@@ -14,6 +14,7 @@ import openeye.oechem as oechem
 import openeye.oeomega as oeomega
 import simtk.openmm.app as app
 import time
+import logging
 
 class GeometryEngine(object):
     """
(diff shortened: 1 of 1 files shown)


perses/rjmc/geometry.py
```diff
@@ -93,6 +94,7 @@ def __init__(self, metadata=None, use_sterics=False, verbose=True):
         self._position_set_time = 0.0
         self.verbose = verbose
         self.use_sterics = use_sterics
+        self._logger = logging.getLogger("geometry")
```
andrrizzi: It is also common to instantiate a single `logger` at the module level. This works too though!

pgrinaway: I gave it a run and it seems to respond as I expected from the calling module, so I think we're good there. INFO isn't too bad here--just the DEBUG stuff is crazy.

PR #425, opened 2017-10-04
Disable tests that should not work yet

jchodera:


jchodera:
PR #426 will replace this since I was able to resolve all the test failures without needing to disable them.

PR #426, opened 2017-10-04
Always use canonical isomeric SMILES with explicit hydrogens

### LINKED ISSUE #394: Explicit hydrogen SMILES is causing failures in SmallMoleculeSetProposalEngine
The travis tests [are failing](https://travis-ci.org/choderalab/perses/jobs/270885563#L1944) because the `SmallMoleculeSetProposalEngine` is now using explicit hydrogens in the SMILES string but is being fed SMILES without explicit hydrogens. 

Short term:
 * We could add hydrogens to SMILES without them
 * We could make sure the input is explicit-hydrogen SMILES (I think this is better for now*)

Long term:
 * We can think a bit more about the way the ProposalEngines work and how to internally represent chemical states.

* I imagine that adding hydrogens to the SMILES strings could lead us to a problem where the hydrogens are inappropriately rearranged as before.

jchodera:
Attempts to fix #394

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,10 +32,10 @@ script:
   - conda create --yes -n test python=$python
   # Activate the test environment
   - source activate test
-  # Add org channel
-  - conda config --add channels ${ORGNAME}
+  # Add omnia
+  - conda config --add channels omnia --add channels conda-forge
   # Add omnia dev channels
-  - conda config --add channels https://conda.anaconda.org/omnia/label/dev
+  - conda config --add channels omnia/label/dev
   # Build the recipe
   - conda build devtools/conda-recipe
   # Install the package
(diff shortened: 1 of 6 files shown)


jchodera:
The last remaining NetCDF error is that python 3.5 (which is failing) uses [`netcdf4` 1.1.9](https://travis-ci.org/choderalab/perses/jobs/283508397#L1517) while python 3.6 (which is succeeding) uses [`netcdf4` 1.2.4](https://travis-ci.org/choderalab/perses/jobs/283508399#L1517).

jchodera:
Ah, the problem is that @pgrinaway hasn't yet followed the [omnia install instructions](http://www.omnia.md/install/) that `conda-forge` is now required.

PR #428, opened 2017-10-25
Overlap test

pgrinaway:
Added two overlap tests. Neither are running on travis, since even with the pentane <-> butane system, the build times out.

### DIFF
--- a/perses/tests/test_hybrid_builder.py
+++ b/perses/tests/test_hybrid_builder.py
@@ -273,14 +273,29 @@ def run_hybrid_endpoint_overlap(topology_proposal, current_positions, new_positi
 def test_simple_overlap():
     """Test that the variance of the endpoint->nonalchemical perturbation is sufficiently small for pentane->butane in vacuum"""
     topology_proposal, current_positions, new_positions = generate_vacuum_topology_proposal()
-    run_hybrid_endpoint_overlap(topology_proposal, current_positions, new_positions)
+    [df, ddf, Neff_max] = run_hybrid_endpoint_overlap(topology_proposal, current_positions, new_positions)
+    
+    if Neff_max < 10:
+        raise Exception("Number of effective samples is less than 10; cannot produce reliable estimates of free energy")
+
+    if ddf > 3.0:
+        msg = "Overlap test in vacuum failed for pentane->butane transformation"
+        msg += "standard deviation %f is greater than 3kT" % ddf
+        raise Exception(msg)
 
 @skipIf(istravis, "Skip expensive test on travis")
 def test_difficult_overlap():
     """Test that the variance of the endpoint->nonalchemical perturbation is sufficiently small for imatinib->nilotinib in solvent"""
     topology_proposal, solvated_positions, new_positions = generate_solvated_hybrid_test_topology(mol_name='imatinib', ref_mol_name='nilotinib')
-    run_hybrid_endpoint_overlap(topology_proposal, solvated_positions, new_positions)
+    [df, ddf, Neff_max] = run_hybrid_endpoint_overlap(topology_proposal, solvated_positions, new_positions)
 
+    if Neff_max < 10:
+        raise Exception("Number of effective samples is less than 10; cannot produce reliable estimates of free energy")
+        
+    if ddf > 3.0:
+        msg = "Overlap test in vacuum failed for pentane->butane transformation"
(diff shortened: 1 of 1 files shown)


jchodera:
What about, for Travis, using a vacuum version of this test, just to make sure (1) the code runs, and (2) there is not something grossly wrong with the relative factory in vacuum? 

I think it would just involve passing a flag to `generate_solvated_hybrid_test_topology()` that would control whether the [`addSolvent` step](https://github.com/choderalab/perses/pull/428/files#diff-9a04b7cf5c843a2cea097e856c8904dfR143) is skipped. You can probably leave PME on.

perses/tests/test_hybrid_builder.py
```diff
@@ -99,6 +123,173 @@
 
 forcefield = app.ForceField('amber99sbildn.xml')
 
+def generate_solvated_hybrid_test_topology(mol_name="naphthalene", ref_mol_name="benzene"):
```
jchodera: Might still be good to add docstrings even for test methods, since others often need to understand what they do for review or reuse.

perses/tests/test_hybrid_builder.py
```diff
@@ -99,6 +123,173 @@
...
+
+        w[iteration] = nonalchemical_reduced_potential - hybrid_reduced_potential
+
+    [t0, g, Neff_max] = timeseries.detectEquilibration(w)
+    w_burned_in = w[t0:]
+
+    [df, ddf] = pymbar.EXP(w_burned_in)
+    ddf_corrected = ddf * np.sqrt(g)
+
+    assert ddf_corrected < 3.0
```
jchodera: Perhaps the threshold should be a constant specified as an optional argument for whatever function does the error checking?

perses/tests/test_hybrid_builder.py
```diff
@@ -99,6 +123,173 @@
...
+        nonalchemical_sampler_state = SamplerState(nonalchemical_positions, box_vectors=new_sampler_state.box_vectors)
+
+        #compute the reduced potential at the nonalchemical system as well:
+        nonalchemical_context, integrator = cache.global_context_cache.get_context(nonalchemical_thermodynamic_state)
+        nonalchemical_sampler_state.apply_to_context(nonalchemical_context, ignore_velocities=True)
+        nonalchemical_reduced_potential = nonalchemical_thermodynamic_state.reduced_potential(nonalchemical_context)
+
+        w[iteration] = nonalchemical_reduced_potential - hybrid_reduced_potential
+
+    [t0, g, Neff_max] = timeseries.detectEquilibration(w)
```
jchodera: Should we also check the `Neff_max` is large enough? If the test is failing badly, there's the danger that `Neff_max ~ 1` and the `ddf` might be erroneously small.

perses/tests/test_hybrid_builder.py
```diff
@@ -99,6 +123,173 @@
...
+
+    #Now create the compound states with different alchemical states
+    lambda_zero_thermodynamic_state = CompoundThermodynamicState(thermodynamic_state, composable_states=[lambda_zero_alchemical_state])
+    lambda_one_thermodynamic_state = CompoundThermodynamicState(thermodynamic_state, composable_states=[lambda_one_alchemical_state])
+
+    return nonalchemical_zero_thermodynamic_state, nonalchemical_one_thermodynamic_state, lambda_zero_thermodynamic_state, lambda_one_thermodynamic_state
+
+def run_endpoint_perturbation(lambda_thermodynamic_state, nonalchemical_thermodynamic_state, initial_hybrid_sampler_state, mc_move, n_iterations, factory, lambda_index=0):
+    """
+
```
jchodera: Document docstring when possible?

perses/tests/test_hybrid_builder.py
jchodera: Perhaps this method should just return `ddf_corrected` and the calling function should test the uncertainty is below a specified threshold, returning helpful error statements if it isn't?

pgrinaway:
>What about, for Travis, using a vacuum version of this test, just to make sure (1) the code runs, and (2) there is not something grossly wrong with the relative factory in vacuum?

Done. Looks like that runs fine on travis. I'll address the other points now and then merge.

pgrinaway:
Requested changes have been made here, merging.

PR #439, opened 2017-11-29
Setup script

pgrinaway:
This PR containts:

- [x] Yaml-based interface for running calculations

- [x] Simple analysis library

- [x] jupyter notebook for quickly examining results


pgrinaway:
I think we can probably merge this. It has a basic yaml interface as well as a rudimentary notebook. It also contains some fixes for the dask stuff, as well as the "run locally" ability, so it would be good to have in master.

Issue #448, opened 2018-01-12
MCL1 perturbation 23~66


brycestx:
Per our conversation during the meeting, I think we should also use MCL1 as a test system for 'validation' because it almost 3x smaller than CDK2. The perturbation I am working with is simple and is located at indices 19 > 13 in [MCL1_ligands.sdf.zip](https://github.com/choderalab/perses/files/1627783/MCL1_ligands.sdf.zip). 
[MCL1_protein_fixed.pdb.zip](https://github.com/choderalab/perses/files/1627785/MCL1_protein_fixed.pdb.zip)
is the protein that I prepared using pdbfixer. 

However after preparing the topology proposal with [NonequilibriumFEPSetup](https://github.com/choderalab/perses/blob/master/perses/dispersed/relative_setup.py#L31), I get an error when trying to build the [HybridTopologyFactory](https://github.com/choderalab/perses/blob/master/perses/annihilation/new_relative.py#L14):
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-17-9565cf892879> in <module>()
      1 snefep = NonequilibriumSwitchingFEP(sfesetup[0], sfesetup[1], sfesetup[2], ncmc_nsteps=1000000, 
      2                                     nsteps_per_iteration=1000, trajectory_directory='mcl1_solvated_011118',
----> 3                                     trajectory_prefix='traj')

~/PycharmProjects/perses/perses/dispersed/relative_setup.py in __init__(self, topology_proposal, pos_old, new_positions, use_dispersion_correction, forward_functions, n_equil_steps, ncmc_nsteps, nsteps_per_iteration, temperature, trajectory_directory, trajectory_prefix, atom_selection, scheduler_address)
    362 
    363         #construct the hybrid topology factory object
--> 364         self._factory = HybridTopologyFactory(topology_proposal, pos_old, new_positions, use_dispersion_correction=use_dispersion_correction)
    365 
    366         #use default functions if none specified

~/PycharmProjects/perses/perses/annihilation/new_relative.py in __init__(self, topology_proposal, current_positions, new_positions, use_dispersion_correction, functions)
    146 
    147         #verify that no constraints are changing over the course of the switching.
--> 148         self._constraint_check_fast()
    149 
    150         #construct dictionary of exceptions in old and new systems

~/PycharmProjects/perses/perses/annihilation/new_relative.py in _constraint_check_fast(self)
    455         for constraint_pair in overlapping_constraints:
[...]

PR #449, opened 2018-01-16
[BUGFIX] Small bugs in 'enumerate_cycle_basis' and relative_setup 'run'

brycestx:
Correcting small bugs in '[enumerate_cycle_basis](https://github.com/choderalab/perses/blob/master/perses/rjmc/topology_proposal.py#L1578)' function in topology_proposal.py and '[run](https://github.com/choderalab/perses/blob/master/perses/dispersed/relative_setup.py#L468)' function in relative_setup.py that I found while trying to debug #448 

The correction in the run function was referenced here #445 but I think it may have never been added to master.

### DIFF
--- a/perses/rjmc/topology_proposal.py
+++ b/perses/rjmc/topology_proposal.py
@@ -1604,7 +1604,7 @@ class SmallMoleculeSetProposalEngine(ProposalEngine):
                 for i in range(len(cycle)):
                     atom_index_1 = cycle[i]
                     atom_index_2 = cycle[(i+1)%len(cycle)]
-                    edge = g.edges[atom_index_1,atom_index_2]
+                    edge = g.edge[atom_index_1][atom_index_2]
                     bond = edge['bond']
                     bond_cycle.append(bond)
                 bond_cycle_basis.append(bond_cycle)
--- a/perses/dispersed/relative_setup.py
+++ b/perses/dispersed/relative_setup.py
@@ -498,7 +498,7 @@ class NonequilibriumSwitchingFEP(object):
                 noneq_trajectory_filenames = [None, None]
 
             #run a round of equilibrium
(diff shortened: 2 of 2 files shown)


perses/rjmc/topology_proposal.py
```diff
@@ -1604,7 +1604,7 @@ def enumerate_cycle_basis(molecule):
                 for i in range(len(cycle)):
                     atom_index_1 = cycle[i]
                     atom_index_2 = cycle[(i+1)%len(cycle)]
-                    edge = g.edges[atom_index_1,atom_index_2]
+                    edge = g.edge[atom_index_1][atom_index_2]
```
jchodera: @brycestx : Are you using `networkx` 1.x or 2.x? I thought the API changed for 2.0, and that's what I was using here. Or are both `Graph.edge` and `Graph.edges` available in 2.0?

brycestx: It does appear that my networkx is 1.11. I just upgraded and it works now. My bad, we should add a requirement in setup.py to address this

jchodera: We've had some trouble with requirements in `setup.py` that can't be installed via `pip`---I wonder if there's a way to just add a function to `setup.py` that checks to make sure all the requirements in `devtools/conda/meta.yaml` are installed and suggest you install them via `conda` if they aren't present.

brycestx:
Gonna close this due to fix of updating networkx. I'll just file an issue for the #445 referenced bug due to the fact that it looks like dask throws an error due to it. However I've been having a bunch of issues with dask, where my workers just get killed.. so I will try to put together an issue for that

jchodera:
Great! Please file a separate issue for each problem/bug/question if possible---it makes it easier to track progress toward resolution that way!

PR #451, opened 2018-01-25
Fix some bugs, add vacuum host-guest example

pgrinaway:
This PR does several things:

* It fixes a major bug where the wrong alchemical state is used for lambda=0 equilibrium simulation
* It fixes another bug where ions can be accidentally added as alchemical atoms
* It fixes a more minor bug where the setup object could not be serialized

It also adds a host-guest example, and a script to set up the host guest in vacuum.


perses/data/host-guest/prepare_host_guest.py
```diff
@@ -0,0 +1,188 @@
...
+import simtk.openmm as openmm
+import openeye.oechem as oechem
+import openeye.oeiupac as oeiupac
+import tempfile
+import os
+import openeye.oeomega as oeomega
+
+#used SAMPL6 sampling challenge inputs
+#git revision 93f642ef9e2c478985cd14f3084d899ae3ee1c6d
+host_guest_path = "/Users/grinawap/SAMPL6/host_guest/SAMPLing/CB8-G3-0/PDB/complex.pdb"
```
jchodera: Can you eliminate the hard-coded paths so that this could be run in this directory, with the required input files provided (and their provenance documented in a `README.md` in this directory)? Otherwise, it won't be easy to tweak/fix these files if needed in future.

perses/dispersed/relative_setup.py
```diff
@@ -669,23 +719,32 @@ def run_setup(setup_options):
...
     pressure = setup_options['pressure'] * unit.atmosphere
     temperature = setup_options['temperature'] * unit.kelvin
     solvent_padding_angstroms = setup_options['solvent_padding'] * unit.angstrom
 
+
     setup_pickle_file = setup_options['save_setup_pickle_as']
 
-    fe_setup = NonequilibriumFEPSetup(protein_pdb_filename, ligand_file, old_ligand_index, new_ligand_index, forcefield_files, pressure=pressure, temperature=temperature, solvent_padding=solvent_padding_angstroms)
+    #fe_setup = NonequilibriumFEPSetup(ligand_file, old_ligand_index, new_ligand_index, forcefield_files, protein_pdb_filename=protein_pdb_filename, receptor_mol2_filename=receptor_mol2, pressure=pressure, temperature=temperature, solvent_padding=solvent_padding_angstroms, solvate=solvate)
+    fe_setup_file = open("/Users/grinawap/solvent_test_5/fesetup.pkl", 'rb')
```
jchodera: This should not be hardcoded to an absolute path.

scripts/basic_setup.yaml
```diff
@@ -38,17 +38,17 @@ forward_functions:
...
-n_steps_ncmc_protocol: 50000
+n_steps_ncmc_protocol: 50
 
 #The number of NCMC steps per move application. This controls the output frequency
 #1 step/move application means writing out the work at every step.
 n_steps_per_move_application: 1
 
 #where to put the trajectories
-trajectory_directory: /Users/grinawap/solvent_test
+trajectory_directory: /Users/grinawap/solvent_test_4
```
jchodera: We should not be using hardcoded absolute paths.

If `dask` needs absolute paths, it can check the current working directory from the submitting client and prepend this to all specified relative paths to generate absolute paths.

scripts/hg_setup.yaml
```diff
@@ -0,0 +1,64 @@
+#provide the full path of the protein PDB file
+receptor_mol2: /Users/grinawap/PycharmProjects/perses/perses/data/host-guest/host.tripos.mol2
```
jchodera: We shouldn't be using absolute paths.

pgrinaway:
@jchodera can you re-review when you have the chance? Thanks!

jchodera:
Sorry, hadn't realized this was ready for re-review! Looking now.

examples/host-guest/hg_setup.yaml
```diff
@@ -1,33 +1,33 @@
 #provide the full path of the protein PDB file
-protein_pdb: /Users/grinawap/Downloads/CDK2_fixed_nohet.pdb
+receptor_mol2: /Users/grinawap/PycharmProjects/perses/perses/data/host-guest/host.mol2
```
jchodera: This file still has hard-coded absolute paths. Can you change these to relative paths?

examples/cdk2-example/cdk2_setup.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+#Provide the list of forcefield files. Non-standard (such as gaff.xml) should
+#be provided with a full path
+forcefield_files:
+    - gaff.xml
+    - amber99sbildn.xml
+    - tip3p.xml
+
+#the temperature and pressure of the simulation
+pressure: 1.0
```
jchodera: Let's document the units here with comments for now, and create an issue to later add support for parsing unit-bearing quantities like YANK does. It just uses a [simple function](https://github.com/choderalab/yank/blob/master/Yank/utils.py#L867-L903) to parse the strings.

examples/host-guest/hg_setup.yaml
pgrinaway: Oh, oops! I thought I had fixed all of them. Sorry, changing now--good catch.

examples/cdk2-example/cdk2_setup.yaml
pgrinaway: Ok, sounds good.

pgrinaway:
Just waiting for tests to pass, then will merge.

PR #452, opened 2018-02-05
Change integrator for NE switching to use BAOAB

pgrinaway:
The integrator being used for nonequilibrium switching had defaulted to a metropolized VVVR, which is not what we want for several reasons:

* It doesn't match the equilibrium simulation
* It's extremely slow (both in terms of runtime and correlation time)

This PR changes the splitting string to `VROHORV`, which is BAOAB with the hamiltonian update in the middle.

### DIFF
--- a/perses/dispersed/relative_setup.py
+++ b/perses/dispersed/relative_setup.py
@@ -411,6 +411,8 @@ class NonequilibriumSwitchingFEP(object):
         #reverse functions to get a symmetric protocol
         self._reverse_functions = {param : param_formula.replace("lambda", "(1-lambda)") for param, param_formula in self._forward_functions.items()}
 
+        #setup splitting string:
+        self._splitting_string = "V R O H R V"
         #set up some class attributes
         self._hybrid_system = self._factory.hybrid_system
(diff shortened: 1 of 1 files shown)


pgrinaway:
This is ready to review.

pgrinaway:
Thanks! Merging.

perses/dispersed/relative_setup.py
```diff
@@ -411,6 +411,8 @@ def __init__(self, topology_proposal, pos_old, new_positions, use_dispersion_cor
         #reverse functions to get a symmetric protocol
         self._reverse_functions = {param : param_formula.replace("lambda", "(1-lambda)") for param, param_formula in self._forward_functions.items()}
 
+        #setup splitting string:
+        self._splitting_string = "V R O H R V"
```
jchodera: There is a typo here: It should be `V R O H O R V`. You're missing the second `O`.

pgrinaway: That's not a typo, as noted [here](https://github.com/choderalab/openmmtools/issues/329).

PR #456, opened 2018-02-22
Fixing shadow_work and dask gather bug.

brycestx:
This PR fixes a large bug we identified last week where multiple calls to [run](https://github.com/choderalab/perses/blob/master/perses/dispersed/relative_setup.py#L512) in `relative_setup.py` leading to integrator errors in [run_protocol](https://github.com/choderalab/perses/blob/master/perses/dispersed/feptasks.py#L144) where `ne_mc_move.apply(thermodynamic_state, sampler_state)` would return near zero work values. Upon debugging this step we found that neither `lambda` nor `step` in the integrator was increasing every time apply was called (over 100 iterations), AFTER the initial [run](https://github.com/choderalab/perses/blob/master/perses/dispersed/relative_setup.py#L512) command. I have validated that this PR fixes this issue on my local linux and lilac setup. This issue was originally raised in #445 but I think this fix is more clear and applicable to both `local` and `distributed` setups.

### DIFF
--- a/perses/dispersed/feptasks.py
+++ b/perses/dispersed/feptasks.py
@@ -5,17 +5,16 @@ import os
 
 #Add the variables specific to the Alchemical langevin integrator
 cache.global_context_cache.COMPATIBLE_INTEGRATOR_ATTRIBUTES.update({
-    "protocol_work" : 0.0,
-    "Eold" : 0.0,
-    "Enew" : 0.0,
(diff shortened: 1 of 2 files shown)


pgrinaway:
This looks good to me. I'm going to override this requirement for merging, since it's failing because of the miniconda MD5 mismatch.

pgrinaway:
Thanks!

PR #459, opened 2018-03-07
[WIP] Refactor NonequilibriumSwitchingMove

pgrinaway:
This PR refactors the `NonequilibriumSwitchingMove` to perform the entire protocol in one move application, and collect work and trajectory data within the move. It also involves some other API changes to compensate for this. Marked WIP because I am still making sure it works.

### DIFF
--- a/perses/dispersed/feptasks.py
+++ b/perses/dispersed/feptasks.py
@@ -45,20 +45,17 @@ class NonequilibriumSwitchingMove(mcmc.BaseIntegratorMove):
         The integrator that will be used for the nonequilibrium switching.
     work_write_interval : int, default None
         The frequency with which to record the cumulative total work. If None, only save the total work at the end
-    trajectory_write_interval : int, default None
-        The frequency with which to record the positions of the system. If None, only save the last frame
-    traj: md.Trajectory, default None
-        The trajectory to use to write the positions along the protocol. If None, don't write anything.
+    top: md.Topology, default None
+        The topology to use to write the positions along the protocol. If None, don't write anything.
+    subset_atoms : np.array, default None
+        The indices of the subset of atoms to write. If None, write all atoms (if writing is enabled)
     context_cache : openmmtools.cache.ContextCache, optional
         The ContextCache to use for Context creation. If None, the global cache
         openmmtools.cache.global_context_cache is used (default is None).
     Attributes
     ----------
-    n_steps : int
-    context_cache : openmmtools.cache.ContextCache
-    reassign_velocities : bool
-    restart_attempts : int or None
     current_total_work : float
+        The total work in kT.
     """
 
     def __init__(self, integrator: integrators.AlchemicalNonequilibriumLangevinIntegrator,
@@ -157,7 +154,7 @@ class NonequilibriumSwitchingMove(mcmc.BaseIntegratorMove):
         for iteration in range(self._number_of_step_moves):
 
             integrator.step(self._number_of_step_moves)
-            self._current_total_work = integrator.get_protocol_work()
+            self._current_total_work = integrator.get_protocol_work(dimensionless=True)
             self._cumulative_work[iteration+1] = self._current_total_work
 
             #if we have a trajectory, we'll also write to it
--- a/perses/dispersed/feptasks.py
+++ b/perses/dispersed/feptasks.py
@@ -31,6 +31,9 @@ import simtk.unit as unit
 EquilibriumResult = NamedTuple('EquilibriumResult', [('sampler_state', states.SamplerState), ('reduced_potential', float)])
 NonequilibriumResult = NamedTuple('NonequilibriumResult', [('cumulative_work', np.array)])
 
+class NoTrajectoryException(Exception):
+    
+
 class NonequilibriumSwitchingMove(mcmc.BaseIntegratorMove):
     """
     This class represents an MCMove that runs a nonequilibrium switching protocol using the AlchemicalNonequilibriumLangevinIntegrator.
@@ -83,11 +86,14 @@ class NonequilibriumSwitchingMove(mcmc.BaseIntegratorMove):
 
         #if we have a trajectory, set up some ancillary variables:
         if self._topology is not None:
+            self._write_frames = True
             n_atoms = self._topology.n_atoms
             n_iterations = self._number_of_step_moves
             self._trajectory_positions = np.zeros([n_iterations, n_atoms, 3])
             self._trajectory_box_lengths = np.zeros([n_iterations, 3])
             self._trajectory_box_angles = np.zeros([n_iterations, 3])
+        else:
+            self._write_frames = False
 
         self._current_total_work = 0.0
 
@@ -219,6 +225,10 @@ class NonequilibriumSwitchingMove(mcmc.BaseIntegratorMove):
             raise Exception("Tried to access a trajectory on a move that hasn't been used yet.")
         else:
             return self._trajectory
+    
+    @property
+    def cumulative_work(self):
+        return self._cumulative_work
 
     def __getstate__(self):
         dictionary = super(NonequilibriumSwitchingMove, self).__getstate__()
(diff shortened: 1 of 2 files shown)


perses/dispersed/feptasks.py
```diff
@@ -84,30 +110,97 @@ def _get_integrator(self, thermodynamic_state):
...
+
+        # Create integrator.
+        integrator = self._get_integrator(thermodynamic_state)
+
+        context, integrator = context_cache.get_context(thermodynamic_state, integrator)
+
+        # Subclasses may implement _before_integration().
+        self._before_integration(context, thermodynamic_state)
+        
+        self._cumulative_work[0] = integrator.get_protocol_work()
```
jchodera: Shouldn't this be zero? Maybe this is a good place to put an `assert` to ensure that it is.

perses/dispersed/feptasks.py
```diff
@@ -84,30 +110,97 @@ def _get_integrator(self, thermodynamic_state):
...
+        else:
+            context_cache = self.context_cache
+
+        # Create integrator.
+        integrator = self._get_integrator(thermodynamic_state)
+
+        context, integrator = context_cache.get_context(thermodynamic_state, integrator)
+
+        # Subclasses may implement _before_integration().
+        self._before_integration(context, thermodynamic_state)
```
jchodera: Does this actually reset the protocol work?

perses/dispersed/feptasks.py
```diff
@@ -84,30 +110,97 @@ def _get_integrator(self, thermodynamic_state):
...
+        #loop through the number of times we have to apply in order to collect the requested work and trajectory statistics.
+        for iteration in range(self._number_of_step_moves):
+
+            integrator.step(self._number_of_step_moves)
+            self._current_total_work = integrator.get_protocol_work(dimensionless=True)
+            self._cumulative_work[iteration+1] = self._current_total_work
+
+            #if we have a trajectory, we'll also write to it
+            if self._topology is not None:
+                context_state = context.getState(getPositions=True, getVelocities=True, getEnergy=True,
```
jchodera: You're computing the energy here, but not storing it anywhere. Did you want to?

perses/dispersed/feptasks.py
```diff
@@ -84,30 +110,97 @@ def _get_integrator(self, thermodynamic_state):
...
+                #get the box angles and lengths
+                a, b, c, alpha, beta, gamma = mdtrajutils.unitcell.box_vectors_to_lengths_and_angles(*sampler_state.box_vectors)
+                self._trajectory_box_lengths[iteration, :] = [a, b, c]
+                self._trajectory_box_angles[iteration, :] = [alpha, beta, gamma]
+
+
+        # We get also velocities here even if we don't need them because we
+        # will recycle this State to update the sampler state object. This
+        # way we won't need a second call to Context.getState().
+        context_state = context.getState(getPositions=True, getVelocities=True, getEnergy=True,
```
jchodera: This new `context_state` is not used anywhere. Do you want to update the MCMC sampler state from the current context instead?

pgrinaway: I use it in line 191.

perses/dispersed/feptasks.py
pgrinaway: I should set that to false

perses/dispersed/feptasks.py
pgrinaway: It should be zero. You asked me to have the first item in the array be zero.

perses/dispersed/feptasks.py
pgrinaway: I will override this to add the reset step.

perses/dispersed/feptasks.py
jchodera: Yes! I'm just suggesting you add an `assert`.

pgrinaway: I'm raising a `RuntimeError` with an appropriate message if it's not zero now.

pgrinaway:
BTW, the travis failure is in python 3.5 and has to do with an lxml dependency. I'm not sure what's going on there.

jchodera:
That might go away if you add, after [this line in `.travis.yml`](https://github.com/choderalab/perses/blob/master/.travis.yml#L36), 
```
conda update --yes --all
```
so that all the packages get updated to their conda-forge equivalents.

perses/dispersed/feptasks.py
```diff
@@ -84,30 +111,103 @@ def _get_integrator(self, thermodynamic_state):
...
+
+            integrator.step(self._work_save_interval)
+            self._current_total_work = integrator.get_protocol_work(dimensionless=True)
+            self._cumulative_work[iteration+1] = self._current_total_work
+
+            #if we have a trajectory, we'll also write to it
+            if self._topology is not None:
+                context_state = context.getState(getPositions=True, getVelocities=True, getEnergy=True,
+                                                 enforcePeriodicBox=thermodynamic_state.is_periodic)
+                sampler_state.update_from_context(context_state)
```
andrrizzi: You can safely pass the `context` here instead of the OpenMM `State`. `SamplerState` will take care of getting only what it needs and cache the unit-less positions/velocities in the meantime so you'll get some speedup this way.

The only reason why in the original `apply()` method we pass the `State` object is because we need the energy as part of the NaN restarting code.

perses/dispersed/feptasks.py
```diff
@@ -31,41 +32,67 @@
...
     """
 
-    def __init__(self, integrator: integrators.AlchemicalNonequilibriumLangevinIntegrator, n_steps: int, **kwargs):
-        super(NonequilibriumSwitchingMove, self).__init__(n_steps, **kwargs)
-        self._integrator = integrator
+    def __init__(self, integrator: integrators.AlchemicalNonequilibriumLangevinIntegrator,
+        work_save_interval: int=None, top: md.Topology=None, subset_atoms: np.array=None, **kwargs):
+
+        super(NonequilibriumSwitchingMove, self).__init__(**kwargs)
+        self._integrator = copy.deepcopy(integrator)
```
andrrizzi: With this `copy`, I think you'll lose the integrator Python layer. You'll have to call `RestorableIntegrator.restore_interface(integrator)` to recover the Python methods, but you'll still lose the Python attributes.

I'm fixing this right now as part of the next feature I'm implementing. It should be ready by the end of the day, and I'd expect the new `openmmtools` release to come out by the end of the week/beginning of next week so, if you can wait a little longer, this won't need changes.

perses/dispersed/feptasks.py
```diff
@@ -84,30 +111,103 @@ def _get_integrator(self, thermodynamic_state):
...
+        # will recycle this State to update the sampler state object. This
+        # way we won't need a second call to Context.getState().
+        context_state = context.getState(getPositions=True, enforcePeriodicBox=thermodynamic_state.is_periodic)
+
+        self._trajectory = md.Trajectory(self._trajectory_positions, self._topology, unitcell_lengths=self._trajectory_box_lengths, unitcell_angles=self._trajectory_box_angles)
+
+        # Subclasses can read here info from the context to update internal statistics.
+        self._after_integration(context, thermodynamic_state)
+
+        sampler_state.update_from_context(context_state)
```
andrrizzi: Same here as above. This call might crash if you don't have the `velocities` in the state. Let me know if you'd like me to add an optional `ignore_velocities` argument in `update_from_context()`.

perses/dispersed/feptasks.py
```diff
@@ -84,30 +111,103 @@ def _get_integrator(self, thermodynamic_state):
...
+        # Check if we have to use the global cache.
+        if self.context_cache is None:
+            context_cache = cache.global_context_cache
+        else:
+            context_cache = self.context_cache
+
+        # Create integrator.
+        integrator = self._get_integrator(thermodynamic_state)
+
+        context, integrator = context_cache.get_context(thermodynamic_state, integrator)
```
andrrizzi: Checking just for caution: make sure the `integrator` here has always the same serialization hash every time you call apply (i.e. all the variables that are not standardized must be the same everytime you call apply), or the `ContextCache` won't be able to match it to an existing `Context`.

We can expand the list of the serialized attributes if you need it (see here: https://github.com/choderalab/openmmtools/blob/master/openmmtools/cache.py#L438), or you can simply add your attributes to that set at runtime.

perses/dispersed/feptasks.py
pgrinaway: ah ok, thanks

perses/dispersed/feptasks.py
pgrinaway: Ok, thanks for the heads up.

perses/dispersed/feptasks.py
pgrinaway: I'm not sure I'm clear on this one. `apply` should only be called once in each protocol, and a copy of the same integrator is sent each time, so I think we're ok? Or is the serialization going to mess that up?

jchodera: Can you protect this so that subsequent calls to `apply` do nothing?

andrrizzi: It may be fine then! I just wanted to make sure you were aware of this problem in case you started to see that your `Context`s are not recycled properly.

pgrinaway: >Can you protect this so that subsequent calls to apply do nothing?

Yes. Going forward, is this adherent with the design principles of the `MCMCMove`? What's the best way to deal with this situation that doesn't result in an `MCMCMove` that behaves anomalously?

pgrinaway:
Ok, I think this is about done. Thanks to Bas, we've resolved the issue with getting `gaff.xml` to be discoverable by OpenMM. Assuming the tests pass, this is (finally) ready to merge.

pgrinaway:
Ok, I think I've fixed it by just removing the call. The integrators are no longer serialized, so this shouldn't be an issue.

andrrizzi:
Removing it should be fine in this case. For the future, there's now a general utility class `RestorableOpenMMObject` that handles both integrators and forces. You can use either `RestorableOpenMMObject.restore_interface()` or, for our integrators, `ThermostatedIntegrator.restore_interface()`.

Issue #342, opened 2017-05-04
RDKit support


jchodera:
SiliconTx says that RDKit support is high priority.

pgrinaway:
Changed to low priority now that we can use OpenEye.

pgrinaway:
Since the call indicated that this was relatively high priority, I am going to outline briefly how OpenEye is currently used and how it might be replaced:

# Reversible Jump steps

## ProposalEngines

* First, we propose a molecule given our set of molecules. This uses OpenEye to:
    * [Make a map](https://github.com/choderalab/perses/blob/master/perses/rjmc/topology_proposal.py#L1545) (with MCSS) between the current molecule and the other molecules
    * [Convert](https://github.com/choderalab/perses/blob/master/perses/rjmc/topology_proposal.py#L1416) from an OpenMM topology residue to SMILES
    * [Convert](https://github.com/choderalab/perses/blob/master/perses/rjmc/topology_proposal.py#L1509) from an OEMol to an OpenMM topology residue

 * I'll update this with how we use OpenEye in the polymer `ProposalEngines`, but I believe it is just for atom mapping.

## SystemGenerator

* This uses the [openmoltools forcefield generators](https://github.com/choderalab/openmoltools/blob/master/openmoltools/forcefield_generators.py) to parameterize the system that is being proposed. 
    * Convert an OEMol into an OpenMM Topology
    * Use antechamber + AM1-BCC (OpenEye quacpac) for charges

## GeometryEngine

* [Convert](https://github.com/choderalab/perses/blob/master/perses/rjmc/geometry.py#L358) OpenMM residue to OEMol
* Use OEMol to identify rigid bonds
    * [Add an extra torsion term](https://github.com/choderalab/perses/blob/master/perses/rjmc/geometry.py#L1506) based on reference conformer from Omega (this helps with ring closure, bond stereochem)
 * Use OEMol to identify angles for aromatic heavy atoms
    * [Add an extra angle term](https://github.com/choderalab/perses/blob/master/perses/rjmc/geometry.py#L1618) for these to further aid ring closure.

## Miscellaneous

* Molecule I/O in testsystems

# [Nonequilibrium setup](https://github.com/choderalab/perses/blob/master/perses/dispersed/relative_setup.py)

[...]

PR #461, opened 2018-03-09
Container infrastructure

pgrinaway:
To simplify debugging and provision of the code in this repo, this PR contains a Dockerfile along with some updates to conda dependencies. Right now, the user would still need to provide an OpenEye license (until #342 is resolved) to use most of the code. Marked WIP because I need to make sure this actually builds correctly (it did before, but without the required dependencies that I've now added).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,7 @@ RUN conda config --add channels openeye
 
 RUN conda install --yes conda-build jinja2 anaconda-client pip
 
-RUN git clone -b container https://github.com/choderalab/perses
+RUN git clone https://github.com/choderalab/perses
 
 RUN conda build perses/devtools/conda-recipe
 RUN conda install --yes --use-local perses-dev
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,22 @@
+FROM nvidia/cuda:9.0-cudnn7-devel
+
+SHELL ["/bin/bash", "-c"]
+
+RUN mkdir /input
+RUN mkdir /output
+RUN mkdir /license
+
+RUN apt-get update && apt-get install -y wget && apt-get install -y git
+RUN MINICONDA="Miniconda3-latest-Linux-x86_64.sh" && \
+    wget --quiet https://repo.continuum.io/miniconda/$MINICONDA && \
+    bash $MINICONDA -b -p /miniconda && \
+    rm -f $MINICONDA
+ENV PATH /miniconda/bin:$PATH
+
+# Add channels and build, then install
+RUN conda config --add channels omnia
+RUN conda config --add channels omnia/label/dev
+RUN conda install --yes conda-build==2.1.17 jinja2 anaconda-client pip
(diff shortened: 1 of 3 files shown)


jchodera:
@pgrinaway : Installation of the openeye tools can now be done with:
```bash
conda install -c openeye openeye-toolkits
```
No need to add the `openeye` channel or install via `pip`. See more [here](https://docs.eyesopen.com/toolkits/python/quickstart-python/linuxosx.html#installing-openeye-python-toolkits).

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+    rm -f $MINICONDA
+ENV PATH /miniconda/bin:$PATH
+
+# Add channels and build, then install
+RUN conda config --add channels omnia
+RUN conda update --yes -n base conda
+RUN conda config --add channels conda-forge
+RUN conda config --add channels omnia/label/dev
+RUN conda config --add channels openeye
+RUN pip install progressbar2
```
jchodera: Why can't you use the conda-installable `progressbar2` from conda-forge?
https://anaconda.org/conda-forge/progressbar2

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+RUN MINICONDA="Miniconda3-latest-Linux-x86_64.sh" && \
+    wget --quiet https://repo.continuum.io/miniconda/$MINICONDA && \
+    bash $MINICONDA -b -p /miniconda && \
+    rm -f $MINICONDA
+ENV PATH /miniconda/bin:$PATH
+
+# Add channels and build, then install
+RUN conda config --add channels omnia
+RUN conda update --yes -n base conda
+RUN conda config --add channels conda-forge
```
jchodera: After installing the `conda-forge` channel, you should use `conda update --yes --all`. 
See our recommendation for using `omnia` here:
http://www.omnia.md/install/

Dockerfile
pgrinaway: Changing now.

jchodera:
@pgrinaway : If you're using the CUDA 9 image, you'll want to pin `openmm` to be `>=7.2.0`

Dockerfile
pgrinaway: Ok

pgrinaway:
>: Installation of the openeye tools can now be done with:

Well, I wanted to add it as a dependency in the `meta.yaml` so that it is automatically installed. Presumably the channel needs to be added for that to happen?

jchodera:
> Presumably the channel needs to be added for that to happen?

Ah, yes, sorry!

pgrinaway:
>@pgrinaway : If you're using the CUDA 9 image, you'll want to pin openmm to be >=7.2.0

Ok, got it.

pgrinaway:
Making sure dockerhub can build this, then removing the `-b ` option from the `git clone` and merging.

pgrinaway:
Ok, the dockerhub build succeeded, merging.

PR #462, opened 2018-03-26
Remove use of no-longer-existing self._forcefield

pgrinaway:
There was a mistake that remained from the previous PR. I eliminated the superfluous forcefield object, but was apparently using it elsewhere. It wasn't caught by the test because the solvated setup is too long to run on travis.


pgrinaway:
This is a really minor change so I'm going to just merge it.

PR #463, opened 2018-03-27
Fix write interval

pgrinaway:
There was a bug stemming from the fact that under the previous design, a different quantity was used to specify the work write interval. This PR fixes that bug, as well as adds a test to check that the dimensions of the work writeout are correct.


pgrinaway:
Ok, I realized there's one more thing for me to do here. I still have work writing tied to configuration writing, which is horribly inefficient. I am decoupling those right now.

pgrinaway:
This is good to merge--I've run it on AWS successfully and tests pass.

PR #466, opened 2018-04-11
Improve alchemical path for atoms being created/destroyed

pgrinaway:
Instead of switching sigma to 1.0 when it is being turned off, this switches sigma to machine epsilon. It can't be switched to zero, otherwise OpenMM will complain that the number of exceptions is changing. When it is switched to 1.0, it has no effect at the endpoint as long as epsilon is zero; however, it creates a very strange protocol for intermediate states that could be inefficient.

I need to evaluate this to see if that is what happens.

### DIFF
--- a/perses/annihilation/new_relative.py
+++ b/perses/annihilation/new_relative.py
@@ -1,3 +1,4 @@
+import sys
 import simtk.openmm as openmm
 import simtk.openmm.app as app
 import simtk.unit as unit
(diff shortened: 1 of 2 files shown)


jchodera:
Instead of changing `sigma` -> machine epsilon, can you try keeping `sigma` the same in both endstates? We're already turning off `epsilon`, so there is no need to perturb `sigma`---it will likely increase the work variance to perturb it. The standard approach is to keep `sigma` constant

To be concrete, instead of
```python
self._hybrid_system_forces['core_sterics_force'].addParticle([sigma, epsilon, 1.0, 0.0])
```
we want
```python
self._hybrid_system_forces['core_sterics_force'].addParticle([sigma, epsilon, sigma, 0.0])
```
Same with
```python
self._hybrid_system_forces['core_sterics_force'].addParticle([1.0, 0.0, sigma, epsilon])
```
we want
```python
self._hybrid_system_forces['core_sterics_force'].addParticle([sigma, 0.0, sigma, epsilon])
```

pgrinaway:
When these tests pass, I'll merge this branch. After that, @brycestx can feel free to benchmark with `master`, since that already has the softcore fixes + will (after this is merged) have the `sigma` fix as well.

Issue #468, opened 2018-05-15
Deal with reciprocal space PME


pgrinaway:
Right now, we don't capture the reciprocal space in the `CustomNonbondedForce`. We have a couple workarounds:

- `CustomCVForce` interpolating between `NonbondedForce`s for electrostatics of each endpoint, as well as a `NonbondedForce` for environment interactions.
    - Pros: somewhat simple to implement
    - Cons: very large number of exceptions (is this a problem?). Either we have to exclude all the environment-environment atoms in the endpoint `NonbondedForces`, or we have to use exceptions to _include_ all the `environment-core` and `core-core` interactions.

- `CustomCVForce` interpolating between two `NonbondedForce`s for electrostatics, including the environment interactions.
    - Pros: fewer exceptions (I think)
    - Cons: It turns out to require a pretty substantial change to the way the `HybridTopology` is set up.

Thoughts? @jchodera @brycestx

jchodera:
What do you mean by "environment interactions" here? Lennard-Jones interactions, or electrostatics? And do you just mean "all interactions that don't involve the alchemical region"?

If electrostatics, remember that PME must include *all* the electrostatics so that a uniform background charge is not used. I'm not certain that splitting out environment electrostatics (if the environment carries a net charge) would have an effect that *exactly* cancels with the alchemical region interactions with the opposite net charge.

pgrinaway:
Oh, sorry, looks like I didn't post this response:

Sorry, by environment interactions I mean "all interactions that don't involve the alchemical region." That's what I named it in the code + docs. And I was only referring to electrostatics.

>If electrostatics, remember that PME must include all the electrostatics so that a uniform background charge is not used. I'm not certain that splitting out environment electrostatics (if the environment carries a net charge) would have an effect that exactly cancels with the alchemical region interactions with the opposite net charge.

So you're saying I _can't_ have 3 `NonbondedForce` objects, one for alchemical system A, another for alchemical system B, and a third for environment (defined above). 

Instead, the only way to address this is to have two `NonbondedForce` objects with electrostatics only for everything + sterics for the environment (again as defined above), with alchemical sterics handled by a now-lone `CustomNonbondedForce`?

jchodera:
> So you're saying I can't have 3 NonbondedForce objects, one for alchemical system A, another for alchemical system B, and a third for environment (defined above).
[...]

PR #469, opened 2018-05-17
[WIP] Exact pme with relative factory

pgrinaway:
This implements what was being discussed in #468 . There is an additional option, `exact_pme`, which results in the electrostatics being split into two `NonbondedForces`, one with the parameters for systemA, and the other with parameters for systemB. With the exception (no pun intended) of sterics for intra-unique interactions, these forces zero out the sterics, as this is still handled by a `CustomNonbondedForce`. The two `NonbondedForce`s are then added to a `CustomCVForce` with the same `lambda_electrostatics` parameter as before. This is marked WIP because I need to enable tests for this new arrangement and see if I messed anything up.

### DIFF
--- a/perses/annihilation/new_relative.py
+++ b/perses/annihilation/new_relative.py
@@ -1296,7 +1296,12 @@ class HybridTopologyFactory(object):
 
                 #add the particle to the hybrid custom sterics and electrostatics.
                 self._hybrid_system_forces['core_sterics_force'].addParticle([sigma, epsilon, sigma, 0.0])
-                self._hybrid_system_forces['core_electrostatics_force'].addParticle([charge, 0.0])
+
+                if not self._exact_pme:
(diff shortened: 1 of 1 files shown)


pgrinaway:
Ok, I think we might have a problem with exceptions. I think this just got a bit more complicated.When we have two `NonbondedForce` objects, the exceptions have to match on OpenCL (haven't tried it on CUDA). This was not the case for `CustomNonbondedForce`s on OpenCL and CUDA, which is why the other scheme worked. This is a bit disappointing. We'll have to brainstorm a bit more about how to get around this problem.

jchodera:
Can you ask on the OpenMM issue tracker how hard it would be for this constraint to be relaxed (explaining why)?

jchodera:
An easy workaround would be to make sure to define the same exceptions for both with nonzero (but very small) numbers used in place of zero, or full values computed if no exception should be used.

davidlmobley:
(Watching this issue)

PR #470, opened 2018-06-11
SAMS sampler and other improvements

pgrinaway:
This PR contains a set of improvements including SAMS support, as well as additional yaml options and the ability to rapidly record shadow work.

### DIFF
--- a/analyse_sams_convergence.py
+++ b/analyse_sams_convergence.py
@@ -3,32 +3,37 @@ import seaborn as sns
 import numpy as np
 import os, sys
 from netCDF4 import Dataset
-
+from glob import glob
 
 
 if __name__ == '__main__':
-    filename, plot_name = sys.argv[1:]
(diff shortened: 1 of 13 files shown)


jchodera:
@pgrinaway : Awesome! It looks like there are some conflicts with `master`.

pgrinaway:
Conflicts resolved. I'm going to make sure the tests pass and then merge.

pgrinaway:
@brycestx Do you know what the right way to deal with the method call [here](https://github.com/choderalab/perses/blob/149ef1eb9a0fe5bebe5ced191102c1bb60f9b9e3/perses/tests/utils.py#L567) is? The docs suggest that method takes only one parameter, and I'm not exactly sure what that code was trying to do.

pgrinaway:
Ah, finally, tests pass! Merging.

PR #472, opened 2018-06-29
[WIP] Modifications to use new 'offset' OpenMM feature for nonbonded electrostatics

jchodera:
@pgrinaway : I think I've made all the changes necessary to use the new ['offset' OpenMM feature](https://github.com/pandegroup/openmm/pull/2105) for near-exact nonbonded electrostatics, but I can't figure out which overlap tests I can run to test that this is actually working. Can you help?

At first, I thought [`test_simple_overlap()`](https://github.com/choderalab/perses/blob/master/perses/tests/test_hybrid_builder.py#L307-L318) may work, but it appears to use several broken functions. Is there another test that is supposed to work for checking overlap with the physical endpoints?

The `offset` feature is being built as `openmm==7.3.0` right now.

This PR would replace #469 since we would no longer need that solution.

### DIFF
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -18,7 +18,7 @@ requirements:
     - numexpr
     - autograd
     - pymbar
-    - openmm >=7.2.0
+    - openmm >=7.3.0
     - parmed
     - openmoltools
     - alchemy >=1.2.3
@@ -43,7 +43,7 @@ requirements:
     - numexpr
     - autograd
     - pymbar
-    - openmm >=7.2.0
+    - openmm >=7.3.0
     - parmed
     - openmoltools
     - alchemy >=1.2.3
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -18,7 +18,7 @@ requirements:
     - numexpr
     - autograd
     - pymbar
-    - openmm >=7.3.0
+    - openmm ==7.3.0
     - parmed
     - openmoltools
     - alchemy >=1.2.3
@@ -43,7 +43,7 @@ requirements:
     - numexpr
     - autograd
     - pymbar
-    - openmm >=7.3.0
+    - openmm ==7.3.0
     - parmed
     - openmoltools
     - alchemy >=1.2.3
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -18,6 +18,7 @@ requirements:
     - numexpr
     - autograd
     - pymbar
+    - cuda92
     - openmm ==7.3.0
     - parmed
     - openmoltools
@@ -43,6 +44,7 @@ requirements:
     - numexpr
     - autograd
     - pymbar
+    - cuda92
     - openmm ==7.3.0
     - parmed
     - openmoltools
--- a/examples/cdk2-example/cdk2_setup.yaml
+++ b/examples/cdk2-example/cdk2_setup.yaml
@@ -13,8 +13,8 @@ new_ligand_index: 15
 #be provided with a full path
 forcefield_files:
     - gaff.xml
-    - amber99sbildn.xml
-    - tip3p.xml
+    - amber14/protein.ff14SB.xml
+    - amber14/tip3p.xml
 
 #the temperature and pressure of the simulation, as well as how much solvent paddding to add
 #units:
@@ -31,7 +31,9 @@ save_setup_pickle_as: fesetup_hbonds.pkl
 
 #the type of free energy calculation.
 #currently, this could be either nonequilibrium or sams
-fe_type: nonequilibrium
+fe_type: sams
+
+n_states: 100
 
 #the forward switching functions. The reverse ones will be computed from this. Only valid for nonequilibrium switching
 forward_functions:
@@ -42,17 +44,17 @@ forward_functions:
     lambda_torsions: lambda
 
 #the number of equilibration iterations:
-n_equilibration_iterations: 100
+n_equilibration_iterations: 10
 
 #The number of equilibrium steps to take between nonequilibrium switching events (only valid for nonequiibrium switching)
-n_equilibrium_steps_per_iteration: 100
+n_equilibrium_steps_per_iteration: 1000
 
 #The length of the ncmc protocol (only valid for nonequilibrium switching)
 n_steps_ncmc_protocol: 50
 
 #The number of NCMC steps per move application. This controls the output frequency
 #1 step/move application means writing out the work at every step. (only valid for nonequilibrium switching)
-n_steps_per_move_application: 10
+n_steps_per_move_application: 2500
 
 #where to put the trajectories
 trajectory_directory: cdk2_neq_hbonds
@@ -66,6 +68,7 @@ atom_selection: not water
 #which phases do we want to run (solvent, complex, or both solvent and complex in the list are acceptable):
 phases:
     - solvent
+    - complex
 
 #timestep in fs
 timestep: 1.0
@@ -83,5 +86,5 @@ measure_shadow_work: True
 scheduler_address: null
 
 #how many iterations to run (n_cycles*n_iterations_per_cycle) (only valid for nonequilibrium switching)
-n_cycles: 5
+n_cycles: 10000
 n_iterations_per_cycle: 1
--- a/perses/annihilation/new_relative.py
+++ b/perses/annihilation/new_relative.py
@@ -45,7 +45,7 @@ class HybridTopologyFactory(object):
     _known_forces = {'HarmonicBondForce', 'HarmonicAngleForce', 'PeriodicTorsionForce', 'NonbondedForce', 'MonteCarloBarostat'}
     _known_softcore_methods = ['default', 'amber', 'classic']
 
-    def __init__(self, topology_proposal, current_positions, new_positions, use_dispersion_correction=False, functions=None, softcore_method='amber', softcore_alpha=None, softcore_beta=None, annihilate_intraunique_charges=True):
+    def __init__(self, topology_proposal, current_positions, new_positions, use_dispersion_correction=False, functions=None, softcore_method='amber', softcore_alpha=None, softcore_beta=None):
         """
         Initialize the Hybrid topology factory.
 
@@ -73,8 +73,6 @@ class HybridTopologyFactory(object):
             "alpha" parameter of softcore sterics. If None is provided, value will be set to 0.5
         softcore_beta: unit, default None
             "beta" parameter of softcore electrostatics. If None is provided, value will be set to 12*unit.angstrom**2
-        annihilate_intraunique_charges: bool, optional, default False
-            If True, unique old/new atoms will be discharged as part of the switching process
 
         """
         self._topology_proposal = topology_proposal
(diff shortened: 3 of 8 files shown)


brycestx:
@jchodera in https://github.com/pandegroup/openmm/pull/2105 you mentioned `conda install -c omnia/label/dev openmm==7.3.1`. For me `7.3.1` is working with this PR and not `7.3.0`

jchodera:
> @jchodera in pandegroup/openmm#2105 you mentioned conda install -c omnia/label/dev openmm==7.3.1. For me 7.3.1 is working with this PR and not 7.3.0

That's weird---I'm pretty sure 7.3.0 is building the appropriate branch:
https://github.com/omnia-md/conda-dev-recipes/blob/master/openmm/meta.yaml#L3-L6

> you mean that that method call is not yet available?

It's currently only available in this PR [https://github.com/pandegroup/openmm/pull/2105], but it's being built as the 7.3.0 `dev` conda package right now.

jchodera:
@pgrinaway : This should be ready to go.

jchodera:
I think I know what the problem is. For other context parameters, we allow these to be computed from a master lambda variable so that they are slaved to it via some functions. It seems we can't take advantage of that here in computing the offset scaling factors via an expression instead of a parameter.

jchodera:
@pgrinaway or @brycestx : Do you have an example YAML input file for SAMS for the CDK2 example?

It looks like we can't use the `HybridTopologyFactory` with passed functions if we want to use the offset capability yet, but I'm discussing the feature extension here:
https://github.com/pandegroup/openmm/pull/2105#issuecomment-401511479

jchodera:
OK, this is strange: In trying to address the other test failures (that don't use the `functions` capability), I don't understand why [this test](https://travis-ci.org/choderalab/perses/jobs/398490312#L2723-L2750) is failing since `openmm==7.3.0` is being installed, which should have the new features:
```
ERROR: perses.tests.test_hybrid_builder.test_compare_energies
  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/perses/tests/test_hybrid_builder.py", line 666, in test_compare_energies
    compare_energies(mol_name=mol_ref_pair[0], ref_mol_name=mol_ref_pair[1])
  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/perses/tests/test_hybrid_builder.py", line 628, in compare_energies
    factory = HybridTopologyFactory(top_proposal, positions, positions)
  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/perses/annihilation/new_relative.py", line 205, in __init__
    self.handle_nonbonded()
  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/perses/annihilation/new_relative.py", line 1259, in handle_nonbonded
    self._hybrid_system_forces['standard_nonbonded_force'].addParticleParameterOffset('lambda_electrostatics', particle_index, (charge_new - charge_old), 0, 0)
  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/simtk/openmm/openmm.py", line 10988, in <lambda>
    __getattr__ = lambda self, name: _swig_getattr(self, NonbondedForce, name)
  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/simtk/openmm/openmm.py", line 74, in _swig_getattr
    return _swig_getattr_nondynamic(self, class_type, name, 0)
  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/simtk/openmm/openmm.py", line 69, in _swig_getattr_nondynamic
    return object.__getattr__(self, name)
AttributeError: type object 'object' has no attribute '__getattr__'
-------------------- >> begin captured logging << --------------------
openmoltools.openeye: DEBUG: keep_confs was set to 1. Molecule positions will be reset.
openmoltools.openeye: DEBUG: MOL
parmed.openmm.parameters: DEBUG: Valid patch combinations:
proposal_engine: INFO: Generating atom map...
proposal_engine: INFO: Atom map took 0.001 s
--------------------- >> end captured logging << ---------------------
```
When trying to reproduce on lilac, I just get a segfault:
```
[chodera@lilac:perses]$ py.test -s --verbose perses/tests/test_hybrid_builder.py::test_compare_energies
=========================================================================================================================================================================== test session starts ============================================================================================================================================================================
platform linux -- Python 3.6.5, pytest-3.5.1, py-1.5.3, pluggy-0.6.0 -- /home/chodera/miniconda/bin/python
cachedir: .pytest_cache
rootdir: /lila/home/chodera/github/choderalab/perses/perses, inifile:
collected 1 item                                                                                                                                                                                                                                                                                                                                                           

perses/tests/test_hybrid_builder.py::test_compare_energies Warning: OEAssignPartialCharges has been deprecated please use OEAssignCharges instead!
Segmentation fault (core dumped)
```

brycestx:
> OK, this is strange: In trying to address the other test failures (that don't use the functions capability), I don't understand why this test is failing since openmm==7.3.0 is being installed, which should have the new features:

@jchodera I got that error with the 7.3.0 build but it seems to work for me with the 7.3.1 build. 

This yaml file works for me with the cdk2-example:
```
#provide the full path of the protein PDB file
protein_pdb: CDK2_fixed_nohet.pdb

#provide the path to the ligand file with coordinates
ligand_file: CDK2_ligands.mol2

#The ligand file contains multiple ligands. Choose the indices of the ligands
#between which we should compute a relative free energy
old_ligand_index: 14
new_ligand_index: 15

#Provide the list of forcefield files. Non-standard files should
#be provided with a full path
forcefield_files:
    - gaff.xml
    - amber14/protein.ff14SB.xml
    - amber14/tip3p.xml

#the temperature and pressure of the simulation, as well as how much solvent paddding to add
#units:
#pressure: atm
#temperature: Kelvin
#padding: angstroms
pressure: 1.0
temperature: 300.0
solvent_padding: 9.0


#The name of the pickle file where we will save the setup object
save_setup_pickle_as: fesetup_hbonds.pkl

#the type of free energy calculation.
#currently, this could be either nonequilibrium or sams
fe_type: sams

n_states: 100

#the forward switching functions. The reverse ones will be computed from this. Only valid for nonequilibrium switching
forward_functions:
    lambda_sterics: lambda
    lambda_electrostatics: lambda
    lambda_bonds: lambda
    lambda_angles: lambda
    lambda_torsions: lambda

#the number of equilibration iterations:
n_equilibration_iterations: 10

#The number of equilibrium steps to take between nonequilibrium switching events (only valid for nonequiibrium switching)
n_equilibrium_steps_per_iteration: 1000

#The length of the ncmc protocol (only valid for nonequilibrium switching)
n_steps_ncmc_protocol: 50

#The number of NCMC steps per move application. This controls the output frequency
#1 step/move application means writing out the work at every step. (only valid for nonequilibrium switching)
n_steps_per_move_application: 2500

#where to put the trajectories
trajectory_directory: cdk2_neq_hbonds

#how to prefix the trajectory files (project-specific name)
trajectory_prefix: cdk2

#which atoms to save (MDTraj selection syntax)
atom_selection: not water

#which phases do we want to run (solvent, complex, or both solvent and complex in the list are acceptable):
phases:
    - solvent
    - complex

#timestep in fs
timestep: 1.0

#splitting string for equilibrium integration (only valid for nonequilibrium switching):
eq_splitting: V R O R V

#splitting string for nonequilibrium switching (only valid for nonequilibrium switching):
neq_splitting: V R O H R V

#whether to measure shadow work (only valid for nonequilibrium switching):
measure_shadow_work: True

#The location of the schduler. If it's null, a localhost scheduler is created (only valid for nonequilibrium switching)
scheduler_address: null

#how many iterations to run (n_cycles*n_iterations_per_cycle) (only valid for nonequilibrium switching)
n_cycles: 10000
n_iterations_per_cycle: 1
```

jchodera:
Thanks for the heads-up. I'll do some more debugging to figure out what is going on with our 7.3.0 multi-CUDA builds.

jchodera:
@brycestx : We have a few more issues to solve before we can merge this, but I'll sync up with @pgrinaway on Mon afternoon to figure out a strategy.

brycestx:
@jchodera  I agree. I ran two mcl1 perturbations last night and both of them had trouble converging and eventually led to nan's after 3183 iterations with a 1 atom perturbation and 4062 iterations for a 6 atom perturbation in the complex phase. Would you like me to post outputs or should we wait until we figure out the offset scaling factors?

jchodera:
Weird---is the online analysis really being computed each iteration? We should probably set that to something like every 500 iterations instead, since it takes up a huge amount of time and slows the simulation considerably (over 100% slower!).

jchodera:
@lnaden is working on more robust automatic backtracking to improve the robustness of SAMS after NaNs, so that should get better. In the meantime, I'd suggest reducing the timestep to 3 fs and seeing if you can get much farther with that. We can bring the timestep back up once the new YANK release hits.

brycestx:
@jchodera I set the online analysis interval to 10, and for the purposes of debugging I find it very helpful to see how things converge as more states are sampled and the number of iterations progresses. I agree that for now running with 2fs sounds like a better idea just to make sure hmr is not the culprit.

jchodera:
> @jchodera I set the online analysis interval to 10, and for the purposes of debugging I find it very helpful to see how things converge as more states are sampled and the number of iterations progresses. I agree that for now running with 2fs sounds like a better idea just to make sure hmr is not the culprit.

You should still be able to run with HMR at 3 fs. The stability should improve enormously as you reduce the timestep---the error goes as something like O(timestep^4).

jchodera:
To discuss with @pgrinaway on Monday: We may want to finish and keep #469 as well since it looks like we won't be able to compute derivatives with respect to lambda with this strategy, while the approach in #469 could work with derivatives.

jchodera:
@brycestx : I finally figured out what is going on with your `openmm==7.3.0` issues. The old 7.3.0 package without a CUDA feature was still on anaconda cloud, and you (and travis) must have been picking up that version because you didn't specify a CUDA feature package.

Can you try this?
```bash
# force uninstall openmm without uninstalling dependencies
conda remove --yes --force openmm
# clean
conda -tipsy
# install CUDA 9.2 version (or whichever CUDA version you have
conda install --yes cuda92 openmm==7.3.0
```
Substitute whatever CUDA version you have for `cuda92`.

jchodera:
@pgrinaway : After conferring with Peter Eastman, I apparently don't need to add anything else to make this work since `addParameterOffset()` automatically registers the context variable. After reinstalling conda, I can confirm this runs!

You can go ahead and refactor/restore the deleted code and merge!

jchodera:
@pgrinaway I'll let you merge

pgrinaway:
I wanted to run it to make sure that it behaves reasonably. Oddly, with the SAMS run, it did. But now I'm seeing the "no parameter `lambda_electrostatics`" error again. What did you do to resolve that, @jchodera ? I have the commit corresponding to the last push to Peter's PR installed.

jchodera:
That error went away when I reinstalled `miniconda` and OpenMM. Perhaps I had two versions of OpenMM accidentally installed?

jchodera:
Sorry to do this @pgrinaway, but I need to demonstrate this on Monday at Merck KGaA, so I am going to merge this and cut a new release.

jchodera:
Test failures are due to timeout of this test:
```
Test that the variance of the endpoint->nonalchemical perturbation is sufficiently small for pentane->butane in vacuum ... Writing omega-00000.mol2
```

PR #473, opened 2018-06-30
Sams support for use-nonbonded-offset PR

brycestx:
Fixes state 0 lambda being set to 1, softcore method to amber, online analysis, and adding more support for sams implementation.


jchodera:
Thanks, @brycestx! Is this OK for me to merge in?

brycestx:
@jchodera I have tested it and it works for me. If it looks good to you and @pgrinaway go ahead.

PR #483, opened 2018-09-21
[DO NOT MERGE] Update GeometryEngine and ExpandedEnsembleSampler to use SamplerState

pgrinaway:
I've update the GeometryEngine and the ExpandedEnsembleSampler, but I still need to update some of the GeometryEngine tests. Opening PR now, though.

### DIFF
--- a/perses/rjmc/geometry.py
+++ b/perses/rjmc/geometry.py
@@ -8,6 +8,7 @@ import logging
 import numpy as np
 import copy
 #from perses.rjmc import coordinate_numba
+from openmmtools import states
 import simtk.openmm as openmm
 import collections
 import openeye.oechem as oechem
@@ -96,7 +97,7 @@ class FFAllAngleGeometryEngine(GeometryEngine):
         self.verbose = verbose
         self.use_sterics = use_sterics
 
-    def propose(self, top_proposal, current_positions, beta):
+    def propose(self, top_proposal, current_sampler_state, beta):
         """
         Make a geometry proposal for the appropriate atoms.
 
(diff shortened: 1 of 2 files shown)


perses/rjmc/geometry.py
```diff
@@ -96,46 +97,49 @@ def __init__(self, metadata=None, use_sterics=False, verbose=True):
...
+    def propose(self, top_proposal, current_sampler_state, beta):
         """
         Make a geometry proposal for the appropriate atoms.
 
         Arguments
         ----------
         top_proposal : TopologyProposal object
             Object containing the relevant results of a topology proposal
+        current_sampler_state : openmmtools.states.SamplerState
+            The current positions and box vectors of the proposal
```
jchodera: Why is `beta` a float? Either we should be using unitless energies (implied units of `kT`) where `beta = 1`, or we should be passing unit-bearing quantities.

perses/rjmc/geometry.py
```diff
@@ -96,46 +97,49 @@ def __init__(self, metadata=None, use_sterics=False, verbose=True):
...
         logp_proposal : float
             The log probability of the forward-only proposal
         """
+        current_positions = current_sampler_state.positions
         current_positions = current_positions.in_units_of(units.nanometers)
         if not top_proposal.unique_new_atoms:
             structure = parmed.openmm.load_topology(top_proposal.old_topology, top_proposal.old_system)
             atoms_with_positions = [structure.atoms[atom_idx] for atom_idx in top_proposal.new_to_old_atom_map.keys()]
             new_positions = self._copy_positions(atoms_with_positions, top_proposal, current_positions)
             return new_positions, 0.0
```
jchodera: This line should return a `SamplerState`, not positions.

Should `_copy_positions()` also be converted to use `SamplerState`? `_copy_positions()` seems a bit kludgey. Is there an easier way to replace its functionality with a method in`SamplerState` that either exists or we could add?

perses/rjmc/geometry.py
```diff
@@ -96,46 +97,49 @@ def __init__(self, metadata=None, use_sterics=False, verbose=True):
...
 
         Arguments
         ----------
         top_proposal : TopologyProposal object
             Object containing the relevant results of a topology proposal
-        new_coordinates : [n, 3] np.ndarray
+        new_sampler_state : openmmtools.states.SamplerState
             The coordinates of the system after the proposal
-        old_coordiantes : [n, 3] np.ndarray
+        old_sampler_state : openmmtools.states.SamplerState
```
jchodera: Again, `beta` below should be a unit-bearing quantity.

perses/rjmc/geometry.py
```diff
@@ -238,16 +242,16 @@ def _logp_propose(self, top_proposal, old_positions, beta, new_positions=None, d
 
             if direction == 'forward':
                 pdbfile = open('%s-initial.pdb' % prefix, 'w')
-                PDBFile.writeFile(top_proposal.old_topology, old_positions, file=pdbfile)
+                PDBFile.writeFile(top_proposal.old_topology, old_sampler_state.positions, file=pdbfile)
```
jchodera: Since these writing methods will omit the box vectors,  which may be important in understanding what is going on in solvated systems, perhaps we want to add a method to `SamplerState` that allows us to obtain an MDTraj `Trajectory` object to write out to PDB? Or for now, band-aid this (and related functions) by providing the box vectors too?

perses/rjmc/geometry.py
pgrinaway: Good point. Maybe there's an easier way to do this directly with `SamplerState`s.

perses/rjmc/geometry.py
pgrinaway: It is, I think I just need to fix the docstring.

perses/rjmc/geometry.py
pgrinaway: Good point. The issue with that is that the `SamplerState` lacks a `Topology` object (as does the `ThermodynamicState`. So I'd guess the method would look something like `SamplerState.to_trajectory(topology: simtk.openmm.app.Topology)`

jchodera:
@pgrinaway : Is this PR still active, or can it be closed?

jchodera:
Marking this as "DO NOT MERGE" so we can pull any needed changes into a new PR.

PR #485, opened 2018-10-10
[WIP] Use delta function in hybrid topology

pgrinaway:
This should fix the issue cited [here](https://github.com/pandegroup/openmm/issues/2185#issuecomment-428426529) by @jchodera .


jchodera:
Looks like you got them all! Might want to let @brycestx know this is a major bugfix!

brycestx:
Great find @pgrinaway @jchodera ! I never realized `step(x) = 0 if x is less than 0, 1 otherwise`, glad this was figured out!

pgrinaway:
Ok, I'm going to merge this having tested locally

PR #486, opened 2018-10-18
Separate sterics and electrostatics

pgrinaway:
In order to have greater simulation stability during NCMC, this PR allows the electrostatics and van der waals of atoms being inserted or deleted to change separately. Right now, I am still getting a NaN, but maybe there is some issue in the changes?

cc: @jchodera


pgrinaway:
Ok, trying with new lambda functions...

pgrinaway:
hm, still seems to have the problem. I'll check back later tonight to see what can be done here.

pgrinaway:
Oh, oddly, the nan came from the jump to/from the hybrid system. not the work...

perses/annihilation/ncmc_switching.py
```diff
@@ -16,8 +16,12 @@
 
 # make something hyperbolic or something to go from on to off to on
 default_hybrid_functions = {
-    'lambda_sterics' : 'lambda',
+    'lambda_sterics_core' : 'lambda',
     'lambda_electrostatics' : 'lambda',
+    'lambda_sterics_insert' : 'select(step(0.5-lambda), 1.0, 2*lambda)',
+    'lambda_sterics_delete' : 'select(step(0.5-lambda), 2.0*(lambda-0.5), 0.0)',
```
jchodera: This should be 
```
    'lambda_sterics_delete' : 'select(step(0.5-lambda), 1.0, 1 - 2.0*(lambda - 0.5))',
```

perses/annihilation/ncmc_switching.py
```diff
@@ -16,8 +16,12 @@
...
+    'lambda_electrostatics_insert' : 'select(step(0.5-lambda),2.0*(lambda-0.5),0.0)',
+    'lambda_electrostatics_delete' : 'select(step(0.5-lambda), 0.0, 2.0*lambda)',
```
jchodera: This should be
```
    'lambda_electrostatics_delete' : 'select(step(0.5-lambda), 1 - 2.0*lambda, 0.0)',

```

jchodera: Whoops, that is
```
    'lambda_electrostatics_delete' : 'select(step(0.5-lambda), 0.0, 1 - 2.0*lambda)',
```

pgrinaway:
I think this PR can now be merged (pushed some local changes that I had forgotten to commit).

jchodera:
@pgrinaway : What was the resolution of driving the lambda parameters in the integrator vs driving them in Python giving different behaviors? Was this indeed different? Or did it end up being the same?

If different, can you open an issue with everything you observed so we can debug?
If same, do we need the integrator vs python branches?

pgrinaway:
>What was the resolution of driving the lambda parameters in the integrator vs driving them in Python giving different behaviors? Was this indeed different? Or did it end up being the same?

It was just a lot easier to check that I was getting the right functions with the python functions. I don't know if there is any difference (though this weekend I can run both branches and see whether anything different happens).

 I think it's useful to keep the Python branch because one could in principle easily write protocols using just Python, but also useful to keep the integrator one, because I haven't done full benchmarking yet--it might turn out to be faster (the early result I shared was a bit anecdotal).

pgrinaway:
Let's discuss more tomorrow in person.

jchodera:
@pgrinaway : Hadn't you already removed the `install_requires` block from `setup.py`? It's causing travis failures right now.

jchodera:
```
Test analysis tools. ... LICENSE: No valid license found for oechem
LICENSE: File(s) used: /home/travis/oe_license.txt
LICENSE: Closest license failed because:
LICENSE: expiration date
```
Hm, I'm guessing the issue here is that you still have a bunch of branches in play and haven't merged everything into `master`. Please get that done ASAP this week.

jchodera:
OK, tests are running now (but timing out, as expected)!

PR #488, opened 2018-10-27
Reject nans

pgrinaway:
This PR just contains a couple minor changes:

* Reject NaNs (instead of throwing an exception)

* initialize the SAMSSampler states with guesses


pgrinaway:
These builds are timing out, so we'll need to fix that. I've added #489 for discussion.

pgrinaway:
Thanks!

PR #497, opened 2019-01-14
Update meta.yaml

pgrinaway:
This is a small PR to fix the `meta.yaml` file for building conda packages.

### DIFF
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -13,7 +13,6 @@ requirements:
   build:
     - python
     - setuptools
-    - ambertools
     - numpy
     - scipy
     - numexpr
@@ -36,7 +35,6 @@ requirements:
   run:
     - python
     - setuptools
-    - ambertools
     - numpy
     - scipy
     - numexpr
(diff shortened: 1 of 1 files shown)


devtools/conda-recipe/meta.yaml
```diff
@@ -28,8 +28,8 @@ requirements:
     - matplotlib
     - seaborn
     - mdtraj
-    - parmed # for testing/debugging
-    #- pdbfixer
+    - parmed
```
jchodera: Do you need the latest parmed, or could the older parmed that comes with `ambertools` be OK?

jchodera: Also, you have `parmed` on line 23 as well.

jchodera: You can also remove `cuda92` from line 21

pgrinaway: I need the latest parmed. How should I specify it?

jchodera: For now, I'd keep `parmed` but list it last. Not sure if order helps, but it'll remind me.

pgrinaway: Ok, done.

jchodera:
I'd leave out `ambertools` since `yank`, `openmoltools`, and `openmmtools` already require it, I think. And you don't use `ambertools` directly, right?

pgrinaway:
Yeah, you're right. Removing.

jchodera:
Go ahead and merge and release when ready

PR #498, opened 2019-01-17
Specify smiles and box vector fixes

pgrinaway:
This is a quick PR to:

* Allow the user to specify current SMILES along with a proposal molecule (this is to avoid the troubles related to round-tripping through an OpenMM Topology).

* Ensure that equilibrium samples are written out with box vectors. It seems that even though the option was enabled as true, and the system had box vectors, periodic boundary conditions, and a barostat, the reporters do not write out box vectors unless the `Topology` object also has them. This has been remedied by setting the `Topology` box vectors to the default `system` box vectors before any simulation.


jchodera:
Would be good to get this merged when ready so we can open bugfix-specific PRs to fix the issues we just identified.

hannahbrucemacdonald:
I think this is ready to merge

examples/neq-switching/run_nonequilibrium.py
```diff
@@ -9,13 +9,13 @@
 from perses.dispersed.feptasks import compute_reduced_potential
 import numpy as np
 
-def traj_frame_to_sampler_state(traj: md.Trajectory, frame_number: int):
+def traj_frame_to_sampler_state(traj: md.Trajectory, frame_number: int,box_vectors):
     xyz = traj.xyz[frame_number, :, :]
     box_vectors = traj.openmm_boxes(frame_number)
-    sampler_state = states.SamplerState(unit.Quantity(xyz, unit=unit.nanometers), box_vectors=box_vectors)
+    sampler_state = states.SamplerState(unit.Quantity(xyz, unit=unit.nanometers))
```
jchodera: This looks like a mistake---why would we accept `box_vectors` as an argument, retrieve it from `traj`, and then not include it in the `sampler_state`?

jchodera:
This actually looks really broken.

@pgrinaway : Can you schedule some time Thu to go through this with me?

examples/neq-switching/run_nonequilibrium.py
jchodera: TODO:
* Add a docstring
* remove `box_vectors` from function arguments
* keep `box_vectors` argument for `states.SamplerState`

examples/neq-switching/run_nonequilibrium.py
```diff
@@ -9,13 +9,13 @@
...
     return sampler_state
 
-def run_rj_proposals(top_prop, configuration_traj, use_sterics, ncmc_nsteps, n_replicates, temperature=300.0*unit.kelvin):
+def run_rj_proposals(top_prop, configuration_traj, use_sterics, ncmc_nsteps, n_replicates, box_vectors, temperature=300.0*unit.kelvin):
```
jchodera: Don't need `box_vectors` as a separate argument---instead, will want to take them from the mdtraj Trajectory

jchodera: Will also need to fix calls to this because positional args are changing; may want to use kwargs there

jchodera:
@pgrinaway @hannahbrucemacdonald : I'm just going to merge this and pull the changes into #504 to keep fixing from there.

PR #504, opened 2019-02-06
Improvements to relative alchemical factory and its tests

hannahbrucemacdonald:
(JDC: This PR makes significant improvements to the hybrid alchemical factory and its tests.)


hannahbrucemacdonald:
The lambda protocol has been separated out into `annihilation/lambda_protocol.py` and lambda has been redefined throughout such that
lambda 0 = off, 1 = on
not
lambda 0 = state A, 1 = state B

docs/relativecdk.rst
```diff
@@ -0,0 +1,30 @@
+.. _relativeCDK2:
```
jchodera: For the examples, it's probably better to keep the documentation in the `README.md` file in each example directory.

hannahbrucemacdonald: Sorry that wasn't supposed to be committed

examples/cdk2-example/cdk2_sams.yaml
```diff
@@ -6,8 +6,8 @@ ligand_file: CDK2_ligands.sdf
 
 #The ligand file contains multiple ligands. Choose the indices of the ligands
 #between which we should compute a relative free energy
-old_ligand_index: 14
-new_ligand_index: 15
+old_ligand_index: LIGANDA_INDEX
```
jchodera: This could be a bit problematic because the job can't be run _without_ using `sed` to replace these---so the YAML file itself isn't valid. A user may want to just run perses directly on this example without having to go through the batch script that runs all pairs.

A better approach might be to include a short python script that does this as needed:
```python
# first, calculate ligand1, ligand2, trajectory_directory, and output_yaml_filename from LSB_JOBINDEX argument

# Read the YAML file, modify it, and write new file
from yaml import load, dump
options = yaml.load('cdk2_sams.yaml')
options['old_ligand_index'] = ligand1
options['new_ligand_index'] = ligand2
options['trajectory_directory'] = trajectory_directory
with open(output_yaml_filename, 'w') as outfile:
    yaml.dump(options, outfile)
```

hannahbrucemacdonald: Sure

perses/annihilation/lambda_protocol.py
```diff
@@ -0,0 +1,79 @@
+from __future__ import print_function
+import numpy as np
+import copy
+import logging
+import traceback
+from openmmtools.alchemy import AlchemicalState
+
+# make something hyperbolic or something to go from on to off to on
+default_hybrid_functions = {
```
jchodera: We should probably document what these three sets of functions are and make sure they describe consistent protocols.

The first set (`default_hybrid_functions`) are clearly for Lepton, but why do we have both `python_hybrid_functions` and `python_reverse_functions`, what differentiates them, and where are they used?

perses/annihilation/lambda_protocol.py
```diff
@@ -0,0 +1,79 @@
...
+    class _LambdaParameter(AlchemicalState._LambdaParameter):
+        pass 
+
+    lambda_sterics_core = _LambdaParameter('lambda_sterics_core') 
+    lambda_sterics_insert = _LambdaParameter('lambda_sterics_insert')
+    lambda_sterics_delete = _LambdaParameter('lambda_sterics_delete')
+    lambda_electrostatics_insert = _LambdaParameter('lambda_electrostatics_insert')
+    lambda_electrostatics_delete = _LambdaParameter('lambda_electrostatics_delete')
+
+    def set_alchemical_parameters(self, master_lambda):
```
jchodera: Can you document what lambda=0 and lambda=1 delineate here? Is lambda=0 the old ligand and lambda=1 the new ligand?

Also, should `python_hybrid_functions` be part of this class? Otherwise, if they're at global scope and the user modifies one of these functions, it will affect the behavior of this class.

jchodera: Perhaps we want to add a test too? We could create a simple system of a ligand in vacuum, set `lambda = 0` and check that all the alchemical parameters correspond to the correct initial state, and then set `lambda = 1` and make sure all alchemical parameters correspond to the correct final state.

perses/annihilation/new_relative.py
```diff
@@ -1305,8 +1305,8 @@ def handle_nonbonded(self):
 
                 # Add particle to the regular nonbonded force, but Lennard-Jones will be handled by CustomNonbondedForce
                 particle_index = self._hybrid_system_forces['standard_nonbonded_force'].addParticle(charge, sigma, 0.0)
-                # Charge will be turned on at lambda_electrostatics_delete = 0, off at lambda_electrostatics_delete = 1
-                self._hybrid_system_forces['standard_nonbonded_force'].addParticleParameterOffset('lambda_electrostatics_delete', particle_index, -charge, 0, 0)
+                # Charge will be turned off at lambda_electrostatics_delete = 0, on at lambda_electrostatics_delete = 1
+                self._hybrid_system_forces['standard_nonbonded_force'].addParticleParameterOffset('lambda_electrostatics_delete', particle_index, +charge, 0, 0)
```
jchodera: This sign change looks like has introduced a bug. If you want to change the way you define `lambda_electrostatics_delete` so that `0` is uncharged and `1` restores the old charges, you also need to change
```
                particle_index = self._hybrid_system_forces['standard_nonbonded_force'].addParticle(charge, sigma, 0.0)
```
to
```
                particle_index = self._hybrid_system_forces['standard_nonbonded_force'].addParticle(0.0, sigma, 0.0)
```

perses/annihilation/lambda_protocol.py
pgrinaway: @jchodera We have both `python_hybrid_functions` and `python_reverse_functions` because the approach that sets the context parameters manually (with Python functions) still runs `lambda` from 0 to 1 on every protocol. I don't think there is a way to easily auto-generate "reverse" functions from the Python expressions (unlike the Lepton strings) without being super hacky/engaging in excessive metaprogramming.

jchodera:
Just looked at the travis test failures, and am collecting some notes here:
- [ ] This message appears a billion times and should be moved to DEBUG or something:
```
proposal_engine: INFO: Generating atom map...
proposal_engine: INFO: Atom map took 0.002 s
```
- [ ] We need travis to install `tqdm`:
```
ImportError: No module named 'tqdm'
```
- [ ] There are lots of errors of residues not found, like
```
ERROR: Test that the hybrid returns the correct positions for the new and old systems after construction
  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/perses/tests/test_hybrid_builder.py", line 296, in test_position_output
    topology_proposal, old_positions, new_positions = utils.generate_vacuum_topology_proposal()
  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/perses/tests/utils.py", line 895, in generate_vacuum_topology_proposal
    topology_proposal = proposal_engine.propose(solvated_system, top_old, current_mol=current_mol, proposed_mol=proposed_mol)
  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/perses/rjmc/topology_proposal.py", line 1959, in propose
    new_mol_start_index, len_new_mol = self._find_mol_start_index(new_topology)
  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/perses/rjmc/topology_proposal.py", line 2087, in _find_mol_start_index
    raise ValueError("There must be exactly one residue with a specific name in the topology. Found %d residues with name '%s'" % (len(mol_residues), resname))
ValueError: There must be exactly one residue with a specific name in the topology. Found 0 residues with name 'benzene'
```
- [ ] The test that prints `for each torsion, calculate position of atom1` outputs a long string of inscrutable numbers that should be suppressed:
```
4.470348439570653e-08
4.470348439570653e-08
4.470348439570653e-08
4.4703483489024386e-08
4.4703484469721395e-08
```
- [ ] Haven't seen this failure before:
```
ERROR: Test that the SmallMoleculeAtomMapper can generate maps that exclude hydrogens
  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/perses/tests/test_topology_proposal.py", line 114, in test_no_h_map
    kinase = KinaseInhibitorsTestSystem()
  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/perses/tests/testsystems.py", line 1841, in __init__
    super(KinaseInhibitorsTestSystem, self).__init__(**kwargs)
  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/perses/tests/testsystems.py", line 1766, in __init__
    systems[environment] = system_generators[environment].build_system(topologies[environment])
  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/perses/rjmc/topology_proposal.py", line 1832, in build_system
    msg += traceback.format_exc(e)
  File "/home/travis/miniconda/envs/test/lib/python3.5/traceback.py", line 163, in format_exc
    return "".join(format_exception(*sys.exc_info(), limit=limit, chain=chain))
  File "/home/travis/miniconda/envs/test/lib/python3.5/traceback.py", line 117, in format_exception
    type(value), value, tb, limit=limit).format(chain=chain))
  File "/home/travis/miniconda/envs/test/lib/python3.5/traceback.py", line 474, in __init__
    capture_locals=capture_locals)
  File "/home/travis/miniconda/envs/test/lib/python3.5/traceback.py", line 332, in extract
    if limit >= 0:
TypeError: unorderable types: ValueError() >= int()
```

jchodera:
We are still finding issues with the hybrid alchemical factory, but merging the improvements from this branch so we can consolidate several code changes and continue debugging.

PR #513, opened 2019-02-25
Clean up geometry engine and add analytical n-bead valence system tests

pgrinaway:
This PR implements the tractable test systems for 3, 4, and 5 bead systems. It contains code to propose transitions `3->4`, `4->5`, and `3->5` beads. This can be run by `python test_geometry_engine.py`. 

* The results of the simulation are written out to `saved_run.npy` (it can take quite a while for the default 5ns for each system + 5000 switching attempts for each pair list above). 

* The format of this file is as follows: `(sys_pos_top, configuration_rp, logp_final_positions)`, where:
    * `sys_pos_top` is a tuple of `System, positions, Topology`.
    * `configuration_rp` is a dictionary with keys of 3, 4, or 5 depending on the system size, and values are tuples `(positions, reduced_potentials)` from the equilibrium simulations.
    * `logp_final_positions` is a dictionary with keys `{original_atom_count}-{final_atom_count}` and records the results of the RJ attempts. The values are `(logp, final_positions)`, where `logp` is a numpy array of `[n_replicates, 4]`. The indices (in order) correspond to the initial logP, the logP_forward from the GeometryEngine, the logP_reverse from the GeometryEngine, and the logP_final.

I have confirmed that these run on my laptop and generate non-insane values, but I do not recall what else is desired from this (plots, if so of what? additional information being stored?). I'm putting this up as WIP so that I can add whatever output is desired. Just post here + I will add it.

BTW, it's probably not practical to run this on travis or something--it can take ~1hr on my MacBook Air.


jchodera:
Thanks so much, @pgrinaway!

The most important thing would be to make sure you are storing both the generalized work and its components in the output file. We can work with Dominic to analyze this data to see if we can figure out what is going on.

We may need to add some more information the precise atoms and valence terms drawn by the geometry engine, though this may be easier to just compute from the final positions.

We can 
In [slack](https://choderalab.slack.com/archives/C0M5LG529/p1545082461464000) and our in-person discussions, we had discussed several diagnostic plots:
* The distribution of generalized work for switches in individual directions, and of its components grouped into logical subgroups. We were hoping to compare the magnitude to [these predicted distributions](https://gist.github.com/jchodera/4dc7e7b87c7ee3e4d62690f78bb2a1e1) that account for the Jacobian contributions to decide whether we thought things were working correctly.
* The free energy difference between states estimated via BAR, since we can analytically or numerically compute this for several of these test systems. 
* The torsion distributions of the newly added beads. For the 3->5 and 3->6 bead test systems, where multiple beads are grown bonded to the same bead, the multidimensional density plot of the distribution of their torsions would illustrate whether there was some complex coupling of their positions. See [this slack post](https://choderalab.slack.com/archives/C0M5LG529/p1548607610010100).

pgrinaway:
OK, I think most of that is being stored already, but I'll check tonight when I'm home + add the plotting code.

jchodera:
Thanks!

pgrinaway:
Added a function to compute generalized work.

perses/rjmc/geometry.py
```diff
@@ -318,8 +319,8 @@ def _logp_propose(self, top_proposal, old_positions, beta, new_positions=None, d
             atom_placements.append(atom_placement_array)
 
 
-            #logp_proposal += logp_r + logp_theta + logp_phi + np.log(detJ)
-            logp_proposal+=logp_r+logp_theta+logp_phi
+            logp_proposal += logp_r + logp_theta + logp_phi - np.log(detJ)
```
pgrinaway: Why is `log(detJ)` subtracted here instead of added? Sorry if I missed some other change that makes this obvious.

perses/tests/test_geometry_engine.py
```diff
@@ -1,5 +1,6 @@
 __author__ = 'Patrick B. Grinaway'
-
+import sys
+sys.path.append('/home/dominic/github/perses/')
```
hannahbrucemacdonald: This should be removed before merging

perses/rjmc/geometry.py
dominicrufa: img_20190225_234459

dominicrufa: img_1963

dominicrufa: the first image is courtesy of jdc. i just rewrote his result so the -det(J) is explicit.  you can also look at the dimensionality of the result  (it must be dimensionless).  i defined P_r \propto (r^2)*exp(-k*beta(r-r0)^2/2) and same for P_theta, except instead of r^2 it is sin(theta).  i should mention that in the 
`_bond_logq` and `_angle_logq` functions, i computed the partition function numerically, which can be expensive...there are better ways of increasing the efficiency of these calculations, or perhaps they can just be omitted with a keyword

dominicrufa: this is great and all, but it still doesn't explain why you were sampling the wrong dihedral distributions in your simple molecule studies. that will take some more digging. but ill give it a try tomorrow

pgrinaway: hm, why is `J^-1` there
image
? Doesn't the change of variables formula involve multiplying by the absolute value of the determinant of the Jacobian matrix of the transformation (not dividing)?

pgrinaway: >i should mention that in the _bond_logq and _angle_logq functions, i computed the partition function numerically

I think this is fine at least for now.

pgrinaway:
>in the geometry engine, i had already incorporated the Z_r and Z_thetas in the _bond_logq and _angle_logq calculations, so subtracting the log(Z_r) and log(Z_theta) in the logp_propose function renders a work constant that is actually off by a factor of -log(Z_r)-log(Z_theta).
the extra partition functions were simply commented out

Sounds good.

perses/rjmc/geometry.py
dominicrufa: that's what i thought initially but that doesn't render the expression dimensionless (neglecting the extra 1/volume dimension from the ratio of the partition functions).  when @jchodera posted this picture on slack as a correction to a back-of-the-envelope derivation on the whiteboard from the day before, he specifically mentioned a correction to det(J).

but in any event, if we multiply by det(J) in the denominator (instead of dividing), then the dimensions don't cancel with the r^2 term and the sin(theta) term in P_r and P_theta, respectively

jchodera: I was surprised to find this too! But in working through the math, I found I needed to add J(r, theta, phi) = r^2 sin(theta) inside integrals when transforming dx dy dz into dr dtheta dphi, but the transformation of probability densities from p(x,y,z) to p(r,theta,phi) required dividing by J. See more on the transformation of probability densities here:

https://en.m.wikibooks.org/wiki/Probability/Transformation_of_Probability_Densities

Note that the probability density gets multiplied by J^-1.

This is also easy to see if you write 

p(x,y,z) dx dy dz = p(r,theta,phi) dr dtheta dphi

You can see that we should be replacing the p(x,y,z) in the denominator of our acceptable probability by something like p(r,theta,phi) * | dr dtheta dphi| / |dx dy dz| which is multiplying by J^-1.

jchodera: I was surprised to find this too! But in working through the math, I found I needed to add J(r, theta, phi) = r^2 sin(theta) inside integrals when transforming dx dy dz into dr dtheta dphi, but the transformation of probability densities from p(x,y,z) to p(r,theta,phi) required dividing by J. See more on the transformation of probability densities here:

https://en.m.wikibooks.org/wiki/Probability/Transformation_of_Probability_Densities

Note that the probability density gets multiplied by J^-1.

This is also easy to see if you write 

p(x,y,z) dx dy dz = p(r,theta,phi) dr dtheta dphi

You can see that we should be replacing the p(x,y,z) in the denominator of our acceptable probability by something like p(r,theta,phi) * | dr dtheta dphi| / |dx dy dz| which is multiplying by J^-1.

pgrinaway: >But in working through the math, I found I needed to add J(r, theta, phi) = r^2 sin(theta) inside integrals when transforming dx dy dz into dr dtheta dphi, but the transformation of probability densities from p(x,y,z) to p(r,theta,phi) required dividing by J.

Oh, yeah, I see what you're saying. Thanks for the clarification (and the catch!)

dominicrufa:
at present, the variance in the analytical work is rather large, but decreases with the number of division chosen for r, theta, and phi, which leads me to believe that this is a consequence of grid sampling?

jchodera:
Can you give some numerics here for N = 100, 1000, 10000?

dominicrufa:
Also, the workflow for the FFAllAngleGeometry has been converted to md_unit_system

jchodera:
Great! Let's review together and merge this PR today @dominicrufa .

perses/rjmc/geometry.py
```diff
@@ -660,29 +715,130 @@ def _angle_logq(self, theta, angle, beta):
...
+        # f=lambda x: np.exp(-beta*k_eq*0.5*(x*units.radians-theta0)**2)*np.sin(x)
+        #
+        # integral, integral_err=integrate.quad(f,-np.pi,np.pi)
+        # logZ_theta=np.log(integral)
+        #
+        #
+        # logq = np.log(np.sin(theta_target/units.radians))-beta*0.5*k_eq*(theta_target-theta0)**2-logZ_theta
+        # return logq
+
+        theta0 = angle.type.theteq.value_in_unit_system(units.md_unit_system);
```
jchodera: Add `assert`s to check dimensionality of input arguments is as expected.

perses/rjmc/geometry.py
```diff
@@ -660,29 +715,130 @@ def _angle_logq(self, theta, angle, beta):
...
+        f=lambda x: np.sin(theta)*np.exp(-(0.5/sigma_theta**2)*(x-theta0)**2)
+        division_size=np.pi/n_divisions
+        theta_array=np.linspace(0,np.pi, n_divisions)
+        theta_index=np.argmin(np.abs(theta-theta_array))
+        Z=sum(f(theta_array))
+        #integral, integral_err=integrate.quad(f,lower_bound,upper_bound)
+        #logZ_theta=np.log(integral)
+        #print('theta: ', theta)
+
+        logp = np.log(np.sin(theta_array[theta_index])*np.exp(-(0.5/sigma_theta**2)*(theta_array[theta_index]-theta0)**2))-np.log(division_size)-np.log(Z)
```
jchodera: This can be 
```python
logp = np.log(np.sin(theta_array[theta_index])) - (0.5/sigma_theta**2)*(theta_array[theta_index]-theta0)**2) - np.log(division_size) - np.log(Z)
```

perses/rjmc/geometry.py
```diff
@@ -660,29 +715,130 @@ def _angle_logq(self, theta, angle, beta):
...
-        k = bond.type.k * self._bond_softening_constant
-        sigma_r = units.sqrt(1.0/(beta*k))
-        r = sigma_r*np.random.randn() + r0
+        import scipy.integrate as integrate
+
+        r0 = bond.type.req.value_in_unit_system(units.md_unit_system)
+        k = bond.type.k.value_in_unit_system(units.md_unit_system) * self._bond_softening_constant
+        sigma_r = units.sqrt((1.0/(beta*k)).value_in_unit_system(units.md_unit_system))
+        lower_bound, upper_bound=max(0.,r0-6*sigma_r), r0+6*sigma_r
+        division_size=(upper_bound-lower_bound)/n_divisions
```
jchodera: `(n_divisions-1)`

perses/rjmc/geometry.py
```diff
@@ -660,29 +715,130 @@ def _angle_logq(self, theta, angle, beta):
...
+        import scipy.integrate as integrate
+
+        theta0 = angle.type.theteq.value_in_unit_system(units.md_unit_system);
+        #print('theta0: ', theta0)
+        k = angle.type.k.value_in_unit_system(units.md_unit_system) * self._angle_softening_constant
+        sigma_theta = units.sqrt(1.0/(beta*k).value_in_unit_system(units.md_unit_system))
+        #print('sigma_theta', sigma_theta)
+        #theta = sigma_theta*np.random.randn() + theta0
+        lower_bound, upper_bound=0., np.pi
+        division_size=np.pi/n_divisions
```
jchodera: `(n_divisions-1)`

perses/rjmc/geometry.py
```diff
@@ -711,15 +867,17 @@ def _torsion_scan(self, torsion, positions, r, theta, n_divisions=360):
         positions_copy = copy.deepcopy(positions)
         positions_copy = positions_copy.value_in_unit(units.nanometers)
```
jchodera: Again use `md_unit_system`

perses/rjmc/geometry.py
```diff
@@ -711,15 +867,17 @@ def _torsion_scan(self, torsion, positions, r, theta, n_divisions=360):
...
         positions_copy = positions_copy.astype(np.float64)
-        r = r.value_in_unit(units.nanometers)
-        theta = theta.value_in_unit(units.radians)
+
+        #r = r.value_in_unit(units.nanometers)
+        #theta = theta.value_in_unit(units.radians)
         bond_atom = torsion.atom2
         angle_atom = torsion.atom3
         torsion_atom = torsion.atom4
         phis = np.arange(-np.pi, +np.pi, (2.0*np.pi)/n_divisions) # Can't use units here.
```
jchodera: Could use this instead of `linspace` idiom.

perses/rjmc/geometry.py
```diff
@@ -940,7 +1101,7 @@ def _internal_to_cartesian(self, bond_position, angle_position, torsion_position
         angle_position = angle_position.astype(np.float64)
         torsion_position = torsion_position.astype(np.float64)
         xyz = coordinate_numba.internal_to_cartesian(bond_position, angle_position, torsion_position, np.array([r, theta, phi], dtype=np.float64))
-        return xyz, r**2*np.sin(theta)
+        return xyz, (r/units.nanometers)**2*np.sin(theta/units.radians)
```
jchodera: Be sure to use `md_unit_system` in returning these values!

perses/tests/test_geometry_engine.py
jchodera: Delete this stuff.

perses/tests/test_geometry_engine.py
```diff
@@ -115,14 +118,20 @@ class FourAtomValenceTestSystem(GeometryTestSystem):
         The periodicity, along with the phase and force constant in radians and kJ/mol respectively, atoms 0-1-2-3
     """
 
-    def __init__(self, bond=True, angle=True, torsion=True):
+    def __init__(self, bond=True, angle=True, torsion=True, n_atoms=4, add_extra_angle=False):
+
```
jchodera: Arguments for `__init__` should be documented *here*, and not at class level.

perses/tests/test_geometry_engine.py
```diff
@@ -102,6 +103,8 @@ class FourAtomValenceTestSystem(GeometryTestSystem):
         Whether to include the angle force term
     torsion : Boolean, default True
         Whether to include the torsion force term
+    four_atom: Boolean, default True
```
jchodera: Get rid of all `Arguments` here---they should be under `__init__`.

jchodera: We can also change class name from `FourAtomValenceTestSystem` to `LinearChainValenceSystem`

perses/tests/test_geometry_engine.py
```diff
@@ -212,6 +244,203 @@ def angle_parameters(self):
...
+
+        sys_pos_top[system_size] = (sys, pos, top)
+
+        configurations, reduced_potentials = simulate_equilibrium(sys, pos, 3)
+
+        configuration_rp[system_size] = (configurations, reduced_potentials)
+
+    return sys_pos_top, configuration_rp
+
+def compute_rp(system, positions):
```
jchodera: Add docstrings for all new methods!

perses/tests/test_geometry_engine_modified.py
```diff
@@ -0,0 +1,349 @@
+__author__='dominic rufa'
```
jchodera: We should delete this file.

perses/tests/test_geometry_engine.py
```diff
@@ -212,6 +244,203 @@ def angle_parameters(self):
     def torsion_parameters(self):
         return (self._default_torsion_periodicity, self._default_torsion_phase, self._default_torsion_k)
 
+def simulate_simple_systems():
```
jchodera: Delete any unused tests if they now appear in final form in `test_geometry_engine_analytical.py`

perses/tests/test_geometry_engine_analytical.py
```diff
@@ -0,0 +1,727 @@
...
+    def positions(self):
+        return self._positions
+
+    @property
+    def energy(self):
+        self._context.setPositions(self._positions)
+        return self._context.getState(getEnergy=True).getPotentialEnergy()
+
+class FourAtomValenceTestSystem(GeometryTestSystem):
+    """
```
jchodera: Make sure docstrings are complete in this file.

perses/tests/test_geometry_engine_analytical.py
```diff
@@ -0,0 +1,727 @@
...
+    @property
+    def positions(self):
+        return self._positions
+
+    @property
+    def energy(self):
+        self._context.setPositions(self._positions)
+        return self._context.getState(getEnergy=True).getPotentialEnergy()
+
+class FourAtomValenceTestSystem(GeometryTestSystem):
```
jchodera: Rename to `LinearValenceTestSystem`, etc.

perses/tests/test_geometry_engine_analytical.py
```diff
@@ -0,0 +1,727 @@
...
+        test=AnalyticalBeadSystems(i, num_iterations)
+        test.create_iid_bead_systems(printer=False)
+        test._forward_transformation(printer=False)
+        test._backward_transformation(printer=False)
+        #test.conduct_proposal_system_equilibrium_simulations(printer=False)
+        work_forward, work_reverse, work_comparison, work_forward_var, work_reverse_var=test.assertion()
+        assert all(item<1e-6 for item in work_comparison) and np.abs(work_forward_var/np.average(work_forward))<1e-3 and np.abs(work_reverse_var/np.average(work_reverse))<1e-3, "test passed"
+
+test_AnalyticalBeadSystems(transformation=[[3,4], [4,5], [3,5]], num_iterations=10)
+    # if filename!=None:
```
jchodera: Delete comments, etc.

perses/rjmc/geometry.py
```diff
@@ -616,9 +632,9 @@ def _internal_to_cartesian(self, bond_position, angle_position, torsion_position
         Calculate the cartesian coordinates given the internal, as well as abs(detJ)
```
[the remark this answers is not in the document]
dominicrufa: this requires a modification to coordinate_numba, otherwise there are compatibility errors in the unit system of choice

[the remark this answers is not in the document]
dominicrufa: i can modify it, but it may cause other tests that use coordinate numba to fail

perses/rjmc/geometry.py
dominicrufa: same issue with coordinate_numba

perses/rjmc/geometry.py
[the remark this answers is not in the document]
jchodera: Are you sure it requires modifying `coordinate_numba`? My understanding was that all the methods in there assume quantities are unitless `np.float64` types in a compatible unit system. 

The reason for using `quantity.value_in_unit_system(unit.md_unit_system)` rather than arbitrarily dividing by some units is to *ensure* everything ends up in a self-consistent unit system.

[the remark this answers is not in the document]
dominicrufa:
```python
~/github/perses/perses/rjmc/geometry.py in _internal_to_cartesian(self, bond_position, angle_position, torsion_position, r, theta, phi)
    614         """
    615         from perses.rjmc import coordinate_numba
--> 616         bond_position = bond_position.value_in_unit(units.md_unit_system).astype(np.float64)
    617         angle_position = angle_position.value_in_unit(units.md_unit_system).astype(np.float64)
    618         torsion_position = torsion_position.value_in_unit(units.md_unit_system).astype(np.float64)

~/anaconda3/lib/python3.7/site-packages/simtk/unit/quantity.py in value_in_unit(self, unit)
    619         Returns underlying value, in the specified units.
    620         """
--> 621         val = self.in_units_of(unit)
    622         if is_quantity(val):
    623             return val._value
```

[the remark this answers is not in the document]
dominicrufa: so its not ``` python coordinate numba ```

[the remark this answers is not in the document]
jchodera: Did you mean to paste a complete stack trace? There isn't an error report in this partial trace.

perses/rjmc/geometry.py
```diff
@@ -642,12 +658,51 @@ def _bond_logq(self, r, bond, beta):
...
+        # upper_bound=r0/units.nanometers+5*np.sqrt(1/(beta*k_eq*units.nanometers**2))
+        #
+        # integral, integral_err=integrate.quad(f,max([0.,lower_bound]),upper_bound)
+        # logZ_r=np.log(integral)
+        #
+        # logq = np.log(r_target**2/((units.nanometers)**2))-beta*0.5*k_eq*(r_target-r0)**2-logZ_r
+        # return logq
+
+
+        r0 = bond.type.req.value_in_unit_system(units.md_unit_system)
```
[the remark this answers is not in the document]
dominicrufa: so there is an ```is_dimensionless``` function; did you have another idea to check the exact dimensionality.  the fact that the ```r_probability_mass_function = (r_array)**2 * np.exp(-(0.5/sigma_r**2) * (r_array-r0)**2)``` returns a dimensionless quantity should be sufficient, no?

perses/rjmc/geometry.py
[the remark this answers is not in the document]
dominicrufa:
```
~/github/perses/perses/rjmc/geometry.py in _internal_to_cartesian(self, bond_position, angle_position, torsion_position, r, theta, phi)
    614         """
    615         from perses.rjmc import coordinate_numba
--> 616         bond_position = bond_position.value_in_unit(units.md_unit_system).astype(np.float64)
    617         angle_position = angle_position.value_in_unit(units.md_unit_system).astype(np.float64)
    618         torsion_position = torsion_position.value_in_unit(units.md_unit_system).astype(np.float64)

~/anaconda3/lib/python3.7/site-packages/simtk/unit/quantity.py in value_in_unit(self, unit)
    619         Returns underlying value, in the specified units.
    620         """
--> 621         val = self.in_units_of(unit)
    622         if is_quantity(val):
    623             return val._value

~/anaconda3/lib/python3.7/site-packages/simtk/unit/quantity.py in in_units_of(self, other_unit)
    655         """
    656         if not self.unit.is_compatible(other_unit):
--> 657             raise TypeError('Unit "%s" is not compatible with Unit "%s".' % (self.unit, other_unit))
    658         f = self.unit.conversion_factor_to(other_unit)
    659         return self._change_units_with_factor(other_unit, f)

TypeError: Unit "nanometer" is not compatible with Unit "UnitSystem([nanometer, dalton, picosecond, elementary charge, kelvin, mole, radian])".
```
sorry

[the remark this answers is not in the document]
dominicrufa: i'll try to figure out where this is coming from

jchodera:
> i used atom (text editor) to delete the file and pushed that...when i try to follow this large file deletion scheme, i can't get past the first step because the saved_run.npy is no longer in the directory... why isn't this resolved, exactly?

@dominicrufa : When you check in a large file and then `git rm` it, since git stores the whole revision history, that means that every time someone checks out the repo, it also pulls in those old files that were added then deleted because they're part of the `git` history. Without being careful not to check in large files (or clean them out if we accidentally do), the repo will grow enormously and checkouts will be painfully slow even if the total size of active source files is small.

It's only 5MB, so it's probably fine for now, but do be careful in future! 

I'll review this PR and get it merged, since it looks like it's ready to go!

dominicrufa:
Ok understood.  The analytical tests need cleanup and docstrings yet.  I
think I’ll just put everything into tests_geometry.py so I don’t have to
recycle the code there.

jchodera:
> The analytical tests need cleanup and docstrings yet.  I think I’ll just put everything into tests_geometry.py so I don’t have to recycle the code there.

OK, I'll let you work on that first! Tag me when it's ready for me to review, clean up, and merge!

jchodera:
This should be in sufficient working condition to merge and move on. It will still take some time before we can even expect travis tests to succeed, but this has been tested locally to some degree.

PR #514, opened 2019-03-13
Update readthedocs config so that all docs build

jchodera:
This PR attempts to fix the readthedocs builds.


hannahbrucemacdonald:
@jchodera - Just want to point out that we also seem to have `improve-docs` and `fixing-docs` branches as well. I can try tidy all of these up into one?

jchodera:
Whoops! I hadn't seen a PR open for `improve-docs`.

This PR is just attempting to update `docs/environment.yml` so that the docs actually compile. Hopefully, it will just take a bit of tinkering, and then you can get the other docs improvements in `improve-docs` merged separately.

jchodera:
Unless you've already fixed that! If so, can you open a PR?

jchodera:
I think I've got this working. New docs will appear [here](https://perses.readthedocs.io/en/fix-readthedocs/) shortly.

jchodera:
Looks like the docs build timed out. I've retriggered the build to see if I can get it to work on a second try.

In future, we can avoid the need to install so many dependencies via `docs/environment.yml` (which takes most of the build time) by localizing external package `import`s within the functions that use them. This would make it possible to still build the docs without the need to install a huge variety of external dependencies, since these would only be needed when specific methods run.

jchodera:
Docs build! See [here](https://perses.readthedocs.io/en/fix-readthedocs/annihilation.html).

I'm merging this and then @hannahbrucemacdonald can merge in a PR from the `improve-docs` branch.

PR #525, opened 2019-04-02
Replace flawed ProposalOrderTools with NetworkXProposalOrder

jchodera:
This PR replaces the flawed `ProposalOrderTools`---which frequently encountered cases where it could not identify topological torsions---with a NetworkX-based scheme from @pgrinaway (rescued from another branch).


jchodera:
The `propane <-> pentane` test now passes!

PR #517, opened 2019-03-14
Overhaul parameterization scheme used for small molecules

jchodera:
This PR focuses on overhauling the way we parameterize small molecules in `perses`.
We switch to a new scheme that uses OpenEye `OEMol`s to initialize a residue template generator for OpenMM's `ForceField`.

PR #533, opened 2019-04-07
Additional fixes for hybrid alchemical factory and tests

jchodera:
* Add `environment.yml` to make it easier to setup a development environment
* Add debug methods to `SystemGenerator` to aid in debugging
* Remove duplicate cancelling angle terms in hybrid alchemical factory
* Rework `generate_vacuum_topology_proposal` to ensure constraints are applied uniformly to old and new systems
* Instrument `run_endpoint_perturbation` with options to write System, State, and trajectories


jchodera:
For the vacuum tests, I've temporarily disabled all force terms except bonds, and turned off constraints, but I'm still seeing large DeltaFs:
```
2-phenyl ethanol -> benzene
...
lambda 0 : [-2.00420401711121, 0.2708217378999726, 0, 101.0]
lambda 1 : [-19.546856611341354, 0.6756946640128504, 0, 101.0]
CROSS VALIDATION
df: -9.652953021659261, ddf: 0.9430134643490395
df: -33.79327810845422, ddf: 0.9879172801436665
```
Hm, at least I *thought* I turned off constraints, but apparently they have reappeared in the hybrid system!
```
        <Constraints>
                <Constraint d=".10860000000000002" p1="0" p2="9"/>
                <Constraint d=".10860000000000002" p1="1" p2="10"/>
                <Constraint d=".10860000000000002" p1="2" p2="11"/>
                <Constraint d=".10860000000000002" p1="3" p2="12"/>
                <Constraint d=".10860000000000002" p1="4" p2="13"/>
                <Constraint d=".10860000000000002" p1="5" p2="19"/>
        </Constraints>
```
but not in the non-alchemical systems. Curious!

I'll investigate further in the morning.

jchodera:
I've fixed the heterogeneous constraints between old and new systems in the vacuum test, but there still may be issues with nonalchemical/hybrid constraint differences I'll look into.

Currently testing with
```bash
pip uninstall --yes perses ; python setup.py install ; nosetests -vv --nocapture perses/tests/test_hybrid_builder.py:test_explosion
```

jchodera:
That's odd---I'm also not properly turning off the nonbonded and torsion terms.

jchodera:
Some good news: I think the large DeltaF we're seeing in the `2-phenyl ethanol -> benzene` system is primarily due to totally normal fluctuations in the bonds and angles (primarily angles).
I've conducted some experiments where I turn off LJ, electrostatics, and torsions and the same magnitude of DeltaF is observed there.
When I zero out the bonds and angles for either the new or old molecule, the fluctuations (and large DeltaF) go away.
I've manually checked the `system.xml` files and they look fine.

was somehow expecting the `DeltaF` to be `~ (3/2)*Ndof*kT`, but realize that there is a difference in dimensionality as we go from the hybrid -> real systems, so the `ln (Z_real V / Z_hybrid)` implicitly has a hypervolume `V` that depends on the unit system.
A harmonic oscillator DeltaF would be something like `ln Z = ln ( sqrt(2*pi) * sigma / length )` where `sigma` is the stddev and `length` is some reference unit, which could be arbitrary.
So I think we should move on to other tests. The original goal of this test was just to make sure the `stddev` in the free energy estimate wasn't crazy, and it isn't---it was just the surprisingly large `DeltaF` that tripped us up.
It's probably OK to move ahead to other tests of the relative alchemical factory, such as actually trying to reproduce hydration free energies (without the transdimensional jump).

I'll merge this now and then focus on #517.

PR #535, opened 2019-04-09
updating branch

dominicrufa:


jchodera:
@dominicrufa : When you get a chance, it would be great to update this PR with more information about which capabilities you've added in this PR!

PR #536, opened 2019-04-14
[WIP] Overhaul parameterization scheme used for small molecules

jchodera:
This PR (originally #517) changes the way that we parameterize small molecules with GAFF. 

## `OEGAFFTemplateGenerator`

Instead of using the flawed [`gaffTemplateGenerator`](https://github.com/choderalab/openmoltools/blob/master/openmoltools/forcefield_generators.py#L477), which was buggy in perceiving chemistry from the OpenMM `Topology` (which lacks chemical information)---and didn't even run in many cases on `osx`---I have implemented a new `OEGAFFTemplateGenerator` that accepts `OEMol` versions of molecules it should match within the `Topology` object in order to determine what chemical species they are.

The `OEGAFFTemplateGenerator` also supports the ability to specify a JSON cache file, allowing parameters for small molecules to be computed only once and then cached thereafter. This should also greatly speed up testing.

### Examples:

    Create a template generator for GAFF for a single OEMol and register it with ForceField:
```python
    >>> from openmoltools.forcefield_generators import OEGAFFTemplateGenerator
    >>> template_generator = OEGAFFTemplateGenerator(oemols=oemol)
    >>> from simtk.openmm.app import ForceField
    >>> forcefield = ForceField('gaff.xml', 'amber14-all.xml', 'tip3p.xml')
    >>> forcefield.registerTemplateGenerator(template_generator.generator)
```
    Create a template generator for GAFF2 for multiple OEMols:
```python
    >>> template_generator = OEGAFFTemplateGenerator(oemols=[oemol1, oemol2], gaff_version='gaff2')
```
    You can also some OEMols later on after the generator has been registered:
```python
    >>> forcefield.add_oemols(oemol)
    >>> forcefield.add_oemols([oemol1, oemol2])
```
    You can optionally create or use a tiny database cache of pre-parameterized molecules:
```python
    >>> template_generator = OEGAFFTemplateGenerator(cache='gaff-molecules.json')
```
    Newly parameterized molecules will be written to the cache, saving time next time!

## `SystemGenerator`

The `SystemGenerator` has been moved to `perses.forcefield`. To use the `OEGAFFTemplateGenerator`, we simple specify some `oemols` and optionally a `cache` path to the JSON file that stores small molecule parameters. This will automatically configure an `OEGAFFTemplateGenerator`.

Here's an example:
```python
    # Set up SystemGenerator for vacuum or explicit solvent
    gaff_filename = get_data_filename('gaff.xml')
[...]

PR #538, opened 2019-04-19
Rjmc torsion proposal

dominicrufa:
Conduct per-atom work tests and prepare geometry engine for work-variance minimization algorithm


jchodera:
@dominicrufa: Did you ever finish this up, or should we chat a bit about what problems remain to be solved?

dominicrufa:
@jchodera : everything seems to be working well here with the exception of a small variance in work, which i think is due to numerical error.  I am going to look at it briefly tomorrow and clean it up before I mark it ready for review.

perses/rjmc/geometry.py
```diff
@@ -289,8 +304,17 @@ def _logp_propose(self, top_proposal, old_positions, beta, new_positions=None, d
             the logp of the proposal
         new_positions : simtk.unit.Quantity with shape (n_atoms, 3) with units compatible with nanometers
             The new positions (same as input if direction='reverse')
+        rjmc_info: list
```
jchodera: Would it be possible to convert this `list` into a `dict` or some other structured object? I just worry that we could easily mess up this complicated list-based data structure if we ever insert more information into the object.

A `dict` is probably the simplest way to go!

perses/rjmc/geometry.py
```diff
@@ -1289,6 +1461,7 @@ def _torsion_log_pmf(self, growth_context, torsion_atom_indices, positions, r, t
             potential_energy = state.getPotentialEnergy()
 
             # Store unnormalized log probabilities
+            #logq_i = -beta*potential_energy + u_r + u_theta
```
jchodera: You can delete things you don't use anymore instead of commenting them out!

perses/rjmc/geometry.py
```diff
@@ -1513,17 +1691,21 @@ def __init__(self, reference_system, growth_indices, global_parameter_name='grow
         # Create new System
         from simtk import openmm
         growth_system = openmm.System()
+        atoms_with_positions_system = openmm.System()
```
jchodera: It would be useful to document what the difference between `growth_system` and `atoms_with_positions_system` is here in comments!

jchodera:
@dominicrufa : It looks like you added required `bond` and `angle` arguments to a few functions without documenting what these are arguments are, e.g.:
https://github.com/choderalab/perses/pull/538/files#diff-a5d8d93e4a23e84d32747673e1244e75R1566

This is causing some tests that used to pass to now fail because the corresponding tests (e.g. in `test_geometry_engine.py`) were not updated for these corresponding parameters.

Can you give me a bit of info about what these represent? I can fix this in #536.

I'm hoping to disable any tests that still need to be overhauled so we can rely on Travis to tell us when we've accidentally broke something like this!

jchodera:
It seems that `bond, angle` were added as required arguments to:
* `_torsion_log_pmf` where they are not used
* `_propose_torsion` where they are not used
* `_torsion_logp` where they are not used

Perhaps you added these for debugging purposes only, but just forgot to remove them? Is it safe to get rid of these?

dominicrufa:
Yes, I used these for debugging and forgot to remove them.  Apologies.
These can be removed.

jchodera:
Thanks! Will do so in #536!

PR #540, opened 2019-05-13
Refactoring package structure and debugging perses-lite

hannahbrucemacdonald:


dominicrufa:
`test_elimination.py` seems to be a set of test _exclusively_ for ncmc switching dynamics.  do we want ncmc in perses-lite if repex works?

hannahbrucemacdonald:
> why is `test_new_relative.py` deleted? is the code here actually redundant at [d35893a](https://github.com/choderalab/perses/commit/d35893a5de0c7b6ec99366caa95c0959a49b0d88)

It's moved to `test_relative.py` as `new_relative.py` was moved to `relative.py` after the previous `relative.py` was deleted

hannahbrucemacdonald:
This large PR restructures the directory layout of `perses`. 

* Relative free energy methods have been reimplemented, via using a `RelativeAlchemicalState`. 
* Created relevant tests, particularly testing RJMC
* Disabled tests not currently expected to work given the current code. 
* Replaced code health badge with codecov badge
* enabled `neglect_angles` functionality in RJMC, to allow for the different flavours of hybrid systems to be used. 
* Asserting energy book keeping in the HybridTopologyFactory
* Adding analysis functionality using networkx tools to analyse free energy cycles.
* Moving functional code in the test suite to utils directory, with some functions replaced with equivalent `openmoltools` functionality.
* Superseding of `new_relative.py` with `relative.py`

perses/analysis/load_simulations.py
```diff
@@ -0,0 +1,303 @@
...
+import logging
+
+
+from pymbar import timeseries
+from pymbar import MBAR
+
+_logger = logging.getLogger("analysis")
+
+
+class molecule(object):
```
jchodera: In general, we want to follow [PEP8 naming conventions](https://www.python.org/dev/peps/pep-0008/#naming-conventions) in our code: Class names should be CamelCase (`Molecule` instead of `molecule`).

perses/analysis/load_simulations.py
```diff
@@ -0,0 +1,303 @@
...
+        self.mol = openeye.smiles_to_oemol(self.smiles)
+        self.mol = generate_conformers(self.mol,max_confs=1)
+        self.exp = kcal_to_kt(float(self.exp))
+        self.experr = kcal_to_kt(float(self.experr))
+        self.calc = kcal_to_kt(float(self.calc))
+        self.calcerr = kcal_to_kt(float(self.calcerr))
+        self.mw = self.calc_mw()
+        self.ha = self.heavy_atom_count()
+
+    def calc_mw(self):
```
jchodera: When it doesn't create huge hardship, we should prefer full words, like `compute_molecular_weight`, to `calc_mw`, which are harder to process.

We also, in general, always want to write docstrings, ideally conforming to the [numpy docstring convention](https://numpydoc.readthedocs.io/en/latest/format.html).

perses/annihilation/relative.py
```diff
@@ -1,888 +1,1968 @@
-import simtk.openmm as mm
-from simtk import unit
+import sys
+import simtk.openmm as openmm
 import simtk.openmm.app as app
+import simtk.unit as unit
```
jchodera: When possible, we want to migrate `import` statements into just the methods that need them, so that imports are handled lazily. This not only speeds up imports, it also keeps code tidier, and makes it easier to compile the sphinx docs without requiring all the dependencies just to import the file.

PR #544, opened 2019-05-17
[WIP] HybridTopologyFactory-geometry compatibility

dominicrufa:
add old-new electrostatics exceptions.  removing electrostatic and steric exceptions from old-new Nonbonded 
Force (into CustomBondForce).  addInteractionGroup between old-old and 
new-new.  removing old-old and new-new Nonbonded Force exception.  add 
angle exceptions and _logger comments


dominicrufa:
https://github.com/choderalab/perses/blob/07f4607c97a68c6221db7041f272b1bfccbbd51f/perses/annihilation/new_relative.py#L1261-L1285
as this is currently implemented in master, it looks like that for every atom pair of unique_old-unique_old, if there is a pairwise exception, the exception is being placed such that the custom sterics is being deleted and the electrostatics/steric exceptions are being placed in the 'standard_nonbonded_force' (without an exception offset).  otherwise, if it is not an exception, the electrostatics and sterics are being placed into the 'standard nonbonded force' without a parameter offset.  If I am reading this correctly, wouldn't this mean that the old-old and new-new atom pairs have nonbonded interactions for the entire lambda protocol?  if this is the case, then it doesn't correspond to the way the `geometry.py` placed/deletes unique_new/unique_old atoms.

perses/annihilation/new_relative.py
```diff
@@ -74,10 +81,22 @@ def __init__(self, topology_proposal, current_positions, new_positions, use_disp
         angle_softening_constant : float
             For bonds between unique atoms and unique-core atoms, soften the force constant at the "dummy" endpoint by this factor.
             If 1.0, do not soften
+        neglected_new_angle_terms : list
+            list of indices from the HarmonicAngleForce of the new_system for which the geometry engine neglected.
+            Hence, these angles must be alchemically grown in for the unique new atoms (forward lambda protocol)
+        neglected_old_angle_terms : list
+            list of indices from the HarmonicAngleForce of the old_system for which the geometry engine neglected.
+            Hence, these angles must be alchemically deleted for the unique old atoms (reverse lambda protocol)
+        forward_1_4_proposal_terms : list
```
hannahbrucemacdonald: `forward_1_4_proposal_terms` and `reverse_1_4_proposal_terms` don't seem to be used anywhere in this script

dominicrufa: i had them as arguments to the `HybridTopologyFactory` in a revious commit but ive since removed them and forgot to remove them from docstrings.  I'll clean it up in the morning

perses/tests/test_new_relative.py
```diff
@@ -0,0 +1,130 @@
...
+Test to assess whether the HybridTopologyFactory is bookkeeping all energies at
+lambda = 0, 1 states as compared to nonalchemical endstates given a single transformation.
+
+The following tests will be conducted...
+    1. a solvated small molecule transformation (current_mol = 'toluene', proposed_mol = '1,2-bis(trifluoromethyl) benzene') will be generated; it will be asserted that the nonalchemical zero state is equal to the alchemical zero state less the forward valence energy.
+       It will also be asserted that the alchemical one state is equal to the nonalchemical one state less  the reverse valence energy.
+"""
+
+
+def energy_bookkeeping(top_proposal, old_positions):
```
hannahbrucemacdonald: How long do these tests take? Should they be run on travis?

perses/tests/utils.py
```diff
@@ -990,6 +993,9 @@ def generate_solvated_hybrid_test_topology(current_mol_name="naphthalene", propo
     #generate topology proposal
     topology_proposal = proposal_engine.propose(solvated_system, solvated_topology)
 
+    if not propose_geometry:
+        return topology_proposal, solvated_positions
```
hannahbrucemacdonald: it's preferable for a function to return the same number of items irrespective of the function choices. 
`return topology_proposal, solvated_positions, None` or `return topology_proposal, solvated_positions, []` makes the function more robust to use

PR #547, opened 2019-06-18
Atom mapping

hannahbrucemacdonald:
fixing issues with atom mapping when not using vacuum phase simulations

### DIFF
--- a/perses/tests/utils.py
+++ b/perses/tests/utils.py
@@ -494,7 +494,7 @@ def  generate_solvated_hybrid_test_topology(current_mol_name="naphthalene", prop
         if render_atom_mapping:
             from perses.utils.smallmolecules import render_atom_mapping
             print(f"new_to_old: {proposal_engine.non_offset_new_to_old_atom_map}")
-            render_atom_mapping(f"{old_smiles}to{new_smiles}.pdf", old_oemol, new_oemol, proposal_engine.non_offset_new_to_old_atom_map)
+            render_atom_mapping(f"{old_smiles}to{new_smiles}.png", old_oemol, new_oemol, proposal_engine.non_offset_new_to_old_atom_map)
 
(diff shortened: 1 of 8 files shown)


hannahbrucemacdonald:
This PR:
* Automatically renders atom-mappings for all simulated legs (which should be identical) automatically when relative simulations are run
* Uses `oemol` object, rather than string when running the topology proposal, which is more robust
* Moves to a more robust dictionary object for storing atom mappings

perses/tests/utils.py
```diff
@@ -410,18 +419,30 @@ def  generate_solvated_hybrid_test_topology(current_mol_name="naphthalene", prop
...
     from perses.rjmc.geometry import FFAllAngleGeometryEngine
 
-    old_oemol, new_oemol = iupac_to_oemol(current_mol_name), iupac_to_oemol(proposed_mol_name)
-
-    old_smiles = oechem.OECreateSmiString(old_oemol,oechem.OESMILESFlag_DEFAULT | oechem.OESMILESFlag_Hydrogens)
-    new_smiles = oechem.OECreateSmiString(new_oemol,oechem.OESMILESFlag_DEFAULT | oechem.OESMILESFlag_Hydrogens)
+    if current_mol_name != None and proposed_mol_name != None:
+        try:
+            old_oemol, new_oemol = iupac_to_oemol(current_mol_name), iupac_to_oemol(proposed_mol_name)
+            old_smiles = oechem.OECreateSmiString(old_oemol,oechem.OESMILESFlag_DEFAULT | oechem.OESMILESFlag_Hydrogens)
```
jchodera: Perhaps you want to move the SMILES string creation options (`oechem.OESMILESFlag_DEFAULT | oechem.OESMILESFlag_Hydrogens`) to a single global definition? Otherwise, you risk getting things out of sync.

PR #551, opened 2019-07-12
Update travis tests to include allowed failures against OpenMM beta and dev versions

jchodera:
This PR updates travis tests to include python 3.6 and python 3.7 with OpenMM 7.3.1, as well as allowed failures testing OpenMM beta and dev versions.

I've also relaxed the tolerance for a stochastically failing numba test, and made it report more information when it fails.


jchodera:
This works as intended now! I'll merge this and we can figure out what broke in OpenMM 7.4.0.

PR #556, opened 2019-08-17
[WIP] improving options for lambda protocols

hannahbrucemacdonald:
This PR will:

- [ ] create a class of lambda protocols, instead of being a class attribute of `RelativeAlchemicalState`
- [ ] add common protocols from other software
- [ ] add option for user defined functions
- [ ] validate set of functions


hannahbrucemacdonald:
@dominicrufa do you think I should change the monotonicity check into a warning rather than being an assert?

dominicrufa:
there should just be a warning thrown for the monotonicity, yes

PR #568, opened 2019-08-28
fixing-rjmc

dominicrufa:
changing logger priorities, correcting md topology for hybrid system, and allowing for no unique old/unique new and no unique new/unique old valence energy reporters.


dominicrufa:
this branch has been manually merged into neq_switching because I can't do it automatically without problems

Issue #560, opened 2019-08-27
expose user-defined atom-mapping options in .yaml file


dominicrufa:
there are many atom-mapping options, but we should at least be able to toggle between `weak` and `strong` (or less- and more-restrictive)

PR #570, opened 2019-08-29
[MRG] Different mapping levels

hannahbrucemacdonald:
Addresses Issue #560

Exposes different 'strengths' of atom mapping in the yaml file, which will map atoms with different criteria (atom type/bond type/hybridization)

weak is the weakest atom differentiation criteria, so puts the most atoms in core (fewer unique atoms)
strong is the strongest, and only the most similar atoms are considered as core (more unique atoms)


hannahbrucemacdonald:
The plan in future is to have a class to handle the mapping protocol that wraps around the oemol atom and bond expressions, but leaving as is for time being

hannahbrucemacdonald:
closes #560

PR #573, opened 2019-08-29
[WIP] adding changes that were missed by branches being a mess

hannahbrucemacdonald:
passing LambdaProtocol into HybridXSampler classes, so that they can be used when setting the alchemical variables of the state.

These changes were made previously, but have been lost in the tangle of branches


hannahbrucemacdonald:
I don't know... the changes just went missing, it might be that I missed moving those changes when trying to separate everything out

hannahbrucemacdonald:
They weren't caught by the test suite, so it might be worth adding a test that checks using a different format of LambdaProtocol with the HybridXSamplers

PR #576, opened 2019-09-11
Update encrypted openeye license for travis

jchodera:
This PR updates the encrypted openeye toolkit license.


hannahbrucemacdonald:
Uhh openeye license is fixed but we are now seeing the openmoltools issue of 'C1' not in list ['C1'...]

jchodera:
@hannahbrucemacdonald : This should be ready to merge since it fixes the license issue.
Did you say you have a fix for the `openmoltools` issue, or should I dive into that?

hannahbrucemacdonald:
Yeah I fixed it previously in `openmoltools` here: https://github.com/choderalab/openmoltools/blob/1473c939e320732dcf4234d56cbbf0e465a30706/openmoltools/forcefield_generators.py#L331

We were having travis passes on master before the oe license expired

PR #578, opened 2019-09-13
pinning to openmoltools>=0.84 for openmm7.4 compatibility

hannahbrucemacdonald:


jchodera:
I've just merged https://github.com/omnia-md/conda-recipes/pull/981 so new packages should appear at https://anaconda.org/omnia/openmoltools/files in a few minutes.

jchodera:
Fixed now!

PR #584, opened 2019-10-04
Ring closure

dominicrufa:


jchodera:
@dominicrufa : Silly question: Why does this PR somehow change 309 different files?

hannahbrucemacdonald:
This is my bad - I copied some code off overflow that I should have looked closer at. We are just reverting this now

PR #585, opened 2019-10-04
[WIP] Ring closure

dominicrufa:
Adding functions to allow for ring closures in relative simulations

### DIFF
--- a/perses/annihilation/relative.py
+++ b/perses/annihilation/relative.py
@@ -107,9 +107,18 @@ class HybridTopologyFactory(object):
             list of indices from the HarmonicAngleForce of the old_system for which the geometry engine neglected.
             Hence, these angles must be alchemically deleted for the unique old atoms (reverse lambda protocol)
         special_new_terms : dict, default None
-            dict of new valence/1,4 terms used for stereochemical/ring strucure stability
+            dict of new valence/1,4 terms used for stereochemical/ring structure stability.  these are specified by the FFAllAngleGeometryEngine.forward_special_terms
+            dict takes the following form:
+                {'omitted_bonds': [], 'omitted_angles': [], 'omitted_torsions': [], 'omitted_1,4s': [], 'extra_torsions': [], 'extra_angles': []}
+                'omitted_<>' take lists of new_topology particle indices comprising a bond, angle, torison, or 1,4 interaction.
+                'extra_torsions' takes a list of lists of the form:
+                    [p1, p2, p3, p4, [periodicity, phase, k, growth_idx]]
+                'extra_angles' are not yet implemented (but necessary for stereochemistry)
+            omitted new terms are annealed in over lambda protocol
         special_old_terms : dict, default None
-            dict of new valence/1,4 terms used for stereochemical/ring strucure stability
+            dict of old valence/1,4 terms used for stereochemical/ring strucure stability.  these are specified by the FFAllAngleGeometryEngine.reverse_special_terms
+            The form is the same as `special_new_terms`
+            omitted old terms are annealed off over the lambda protocol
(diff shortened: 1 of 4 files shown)


perses/annihilation/relative.py
```diff
@@ -66,6 +66,8 @@ def __init__(self,
                  soften_only_new=False,
                  neglected_new_angle_terms = [],
                  neglected_old_angle_terms = [],
+                 special_new_terms = None,
```
jchodera: Should we use a term other than `special`? Do these represent biasing terms that are applied only during proposal, but are corrected for during the RJMC acceptance step?

dominicrufa: special terms is a dict of omitted bonds, angles, and torsions (which are annealed in through AIS), as well as extra angles and torsions that bias  RJ proposals (which are annealed off during the AIS protocol);

perses/annihilation/relative.py
```diff
@@ -104,6 +106,10 @@ def __init__(self,
         neglected_old_angle_terms : list
             list of indices from the HarmonicAngleForce of the old_system for which the geometry engine neglected.
             Hence, these angles must be alchemically deleted for the unique old atoms (reverse lambda protocol)
+        special_new_terms : dict, default None
+            dict of new valence/1,4 terms used for stereochemical/ring strucure stability
```
jchodera: `strucure` -> `structure`

Can we also add more documentation about how these terms are used? As you note above:
> special terms is a dict of omitted bonds, angles, and torsions (which are annealed in through AIS), as well as extra angles and torsions that bias RJ proposals (which are annealed off during the AIS protocol);

perses/annihilation/relative.py
```diff
@@ -104,6 +106,10 @@ def __init__(self,
...
+        special_old_terms : dict, default None
```
jchodera: Same here as above.

perses/annihilation/relative.py
dominicrufa: will do!

perses/annihilation/relative.py
dominicrufa: https://github.com/choderalab/perses/pull/585/commits/518105bb9e33e8de9403596e2d21d02322cc78a0

hannahbrucemacdonald:
If I merge conflicts can this be merged?

dominicrufa:
> If I fix conflicts can this be merged?

let me give it a read tonight and write tests (in case i forgot any); i realized i have to change a few things in the `factory`

jchodera:
Happy to chat about algorithms if needed!

dominicrufa:
> Happy to chat about algorithms if needed!

i'm concerned about the following: if we are adding a new ring (i.e. all atoms in the ring are new, would it be advisable to simply not anneal in the final bond that would close the ring over the protocol, and instead just place a lambda-independent bond there?)

my other concern is that all ring closures happening with the PR now add biasing torsions that attempt to make the ring planar (regardless of whether it is aromatic), which is not generally what we want.  however, this would involve passing an oemol to the geometry engine to check whether the ring is aromatic or not (this is also what has to happen to check the chirality of certains atoms for biasing torsions); if the ring is not aromatic, then do i simply not add biasing torsions?

hannahbrucemacdonald:
Based on some misunderstandings that I've had, I think it is important to have clarity about what 'ring closure' means -- does this constitute replacing X-membered ring (all unique_old) with Y-membered ring (all unique new) i.e. entire ring for entire ring substitution OR turning (x-ethyl,(x+1)-ethyl blah into benzene) i.e.e ring with non-ring substitutioin which I think are two distinct issues, with their own issues

Issue #588, opened 2019-10-09
Be consistent with minimizers


hannahbrucemacdonald:

hannahbrucemacdonald:
It's been suggested that simulations that are NaNing on the first step of equilibration might be because the Fire minimizer is used for relative calculations, while the LocalEnergyMinimizer is used for NEQ.

Even if this isn't responsible for the nans, we should be self-consistent anyway

jchodera:
@hannahbrucemacdonald : Did the minimizer change help your NaN issues with Mcl-1, or are you still having NaN issues?

If you're still having issues, can you create a new Perses issue with information on how to reproduce and debug?

hannahbrucemacdonald:
Yep, this fixes the NaN issues

PR #589, opened 2019-10-10
[MRG] Fix minimizers

hannahbrucemacdonald:
This PR removes any use of the FIREMinimizationIntegrator, and instead uses functions that rely on openmm.LocalEnergyMinimizer instead, which has been more robust - particularly with hybrid systems.

### DIFF
--- a/perses/annihilation/relative.py
+++ b/perses/annihilation/relative.py
@@ -135,8 +135,15 @@ class HybridTopologyFactory(object):
         self._interpolate_14s = interpolate_old_and_new_14s
 
         #new attributes from the modified geometry engine
-        self.neglected_new_angle_terms = neglected_new_angle_terms
-        self.neglected_old_angle_terms = neglected_old_angle_terms
+        if neglected_old_angle_terms:
+            self.neglected_old_angle_terms = neglected_old_angle_terms
+        else:
+            self.neglected_old_angle_terms = []
+
+        if neglected_new_angle_terms:
+            self.neglected_new_angle_terms = neglected_new_angle_terms
+        else:
+            self.neglected_new_angle_terms = []
 
         if bond_softening_constant != 1.0:
             self._bond_softening_constant = bond_softening_constant
--- a/perses/app/setup_relative_calculation.py
+++ b/perses/app/setup_relative_calculation.py
@@ -323,7 +323,7 @@ def run_setup(setup_options):
         atom_selection = setup_options['atom_selection']
         _logger.info(f"\tatom selection detected: {atom_selection}")
     else:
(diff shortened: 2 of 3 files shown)


hannahbrucemacdonald:
This addresses #588

perses/annihilation/relative.py
```diff
@@ -135,8 +135,15 @@ def __init__(self,
         self._interpolate_14s = interpolate_old_and_new_14s
 
         #new attributes from the modified geometry engine
-        self.neglected_new_angle_terms = neglected_new_angle_terms
-        self.neglected_old_angle_terms = neglected_old_angle_terms
+        if neglected_old_angle_terms:
```
jchodera: Has this PR accidentally pulled in additional changes?

perses/app/setup_relative_calculation.py
```diff
@@ -328,7 +324,7 @@ def run_setup(setup_options):
         _logger.info(f"\tatom selection detected: {atom_selection}")
     else:
         _logger.info(f"\tno atom selection detected: default to None.")
-        atom_selection = None
+        atom_selection = 'all'
```
jchodera: Looks like this has changed for some reason?

perses/app/setup_relative_calculation.py
```diff
@@ -538,10 +534,6 @@ def run_setup(setup_options):
             _logger.info(f'\tRunning {phase} phase...')
             hss_run = hss[phase]
 
-            _logger.info(f"\t\tminimizing...\n\n")
-            hss_run.minimize()
```
jchodera: Do we really want to eliminate minimization? I think we just want to switch minimizers.

jchodera:
Looks like there may be a bunch of unrelated changes here?

Is there a simple unit test that, for example, tests the setup/minimization process in vacuum so it can run on travis?

perses/annihilation/relative.py
dominicrufa: this is an additional change independent of the minimization issue; there was a discrepancy between `self.neglect_`... being an empty list versus a `NoneType`

perses/app/setup_relative_calculation.py
hannahbrucemacdonald: We are still minimizing, it's just done as part of the setup. We were minimizing twice previously

perses/app/setup_relative_calculation.py
hannahbrucemacdonald: Yes - I don't think saving no atom coordinates should be the default

jchodera:
Is this PR ready to merge?

hannahbrucemacdonald:
Yes!

PR #591, opened 2019-10-17
Add travis branch to test openmm conda-forge testing builds

jchodera:
This PR adds a travis branch (under `allowed_failures:`) that also tests the [`conda-forge` test builds](https://anaconda.org/conda-forge/openmm/files).


jchodera:
I think I've fixed this now!

Issue #592, opened 2019-10-22
High energy difference between alchemical and non-alchemical energies with furan/thiophene


hannahbrucemacdonald:
Getting huge energy differences between the alchemical end states when trying to perturb furan into thiophene. Issue doesn't arise when perturbing either of these to pyrolle (which has the extra hydrogen)

> Traceback (most recent call last):
  File "/home/brucemah/software/perses/perses/app/setup_relative_calculation.py", line 454, in <module>
    setup_dict = run_setup(setup_options)
  File "/home/brucemah/software/perses/perses/app/setup_relative_calculation.py", line 395, in run_setup
    zero_state_error, one_state_error = validate_endstate_energies(_top_prop, _htf, _forward_added_valence_energy, _reverse_subtracted_valence_energy, beta = 1.0/(kB*temperature), ENERGY_THRESHOLD = ENERGY_THRESHOLD)
  File "/home/brucemah/miniconda3/envs/omm_source/lib/python3.6/site-packages/perses-0.1-py3.6.egg/perses/tests/utils.py", line 758, in validate_endstate_energies
    assert abs(nonalch_one_rp - alch_one_rp + subtracted_energy) < ENERGY_THRESHOLD, f"The one state alchemical and nonalchemical energy absolute difference {abs(nonalch_one_rp - alch_one_rp + subtracted_energy)} is greater than the threshold of {ENERGY_THRESHOLD}."
AssertionError: The one state alchemical and nonalchemical energy absolute difference 174.23110206093253 is greater than the threshold of 0.0001.

hannahbrucemacdonald:
Error is always at lambda = 1 whether this is furan or thiophene, lamdba = 0 is fine

hannahbrucemacdonald:
It looks like this is happening because the core-core bonds are being treated as `HarmonicBondForce` when they should be `CustomHarmonicBondForce`

jchodera:
Can you also post the map for these transformations?

jchodera:
[`self._hybrid_system_forces['core_bond_force']`](https://github.com/choderalab/perses/blob/master/perses/annihilation/relative.py#L545-L580) is a `CustomBondForce`, so perhaps these bonds are being added to the wrong key?

jchodera:
[...]

PR #593, opened 2019-10-22
[WIP] fixing bug with no unique

hannahbrucemacdonald:
Addresses #592

Core atoms are being mistakenly labelled as environment when:
- there are no unique new or old atoms
- the topology proposal is being generated from a higher order phase in the setup.

The TopologyProposal is instantiated using parameters gleaned from setting up the higher order (i.e. complex > solvent > vacuum) and the new and old residue names aren't set, so default to `None`
i.e. this bug doesn't arise if you _just_ run a vacuum leg, only if a solvent leg is set up and we try 'extract' the vacuum leg from that.

The defaults in TopologyProposal could instead be changed to 'MOL'

Adding an assert that the number of cores must be at least 3

### DIFF
--- a/perses/annihilation/relative.py
+++ b/perses/annihilation/relative.py
@@ -371,7 +371,7 @@ class HybridTopologyFactory(object):
 
         #The union of the two will give the core atoms that can result from either new or old topology
         total_core_atoms = core_atoms_from_old.union(core_atoms_from_new)
-        assert sorted(core_atoms_from_old) == sorted(core_atoms_from_new), 'Core atoms must match between old and new systems'
+        assert set(core_atoms_from_old) == set(core_atoms_from_new), 'Core atoms must match between old and new systems'
 
         #as a side effect, we can now compute the environment atom indices too, by subtracting the core indices
         #from the mapped atom set (since any atom that is mapped but not core is environment)
--- a/perses/annihilation/relative.py
+++ b/perses/annihilation/relative.py
@@ -370,13 +370,13 @@ class HybridTopologyFactory(object):
                                                                      self._new_to_hybrid_map, name_of_residue)
 
         #The union of the two will give the core atoms that can result from either new or old topology
-        # TODO: Shouldn't these sets be the same?
         total_core_atoms = core_atoms_from_old.union(core_atoms_from_new)
+        assert sorted(core_atoms_from_old) == sorted(core_atoms_from_new), 'Core atoms must match between old and new systems'
 
         #as a side effect, we can now compute the environment atom indices too, by subtracting the core indices
         #from the mapped atom set (since any atom that is mapped but not core is environment)
         environment_atoms = mapped_hybrid_atoms_set.difference(total_core_atoms)
-
+        
         return total_core_atoms, environment_atoms
 
     def _determine_core_atoms_in_topology(self, topology, unique_atoms, mapped_atoms, hybrid_map, residue_to_switch):
@@ -416,6 +416,8 @@ class HybridTopologyFactory(object):
                         #we specifically want to add the hybrid atom.
                         hybrid_index = hybrid_map[atom_index]
                         core_atoms.add(hybrid_index)
+
+        assert len(core_atoms) >= 3, 'Cannot run a simulation with fewer than 3 core atoms. System has {len(core_atoms)}'        
 
         return core_atoms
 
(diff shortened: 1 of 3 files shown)


hannahbrucemacdonald:
I will add unit tests tomorrow morning!


@jchodera @dominicrufa do you think I should change the `TopologyProposal` default residue names to 'MOL'? or make them non-optional parameters?

dominicrufa:
i think defaults should be 'MOL' for all of these.

jchodera:
> @jchodera @dominicrufa do you think I should change the TopologyProposal default residue names to 'MOL'? or make them non-optional parameters?

I'm not sure I understand all the implications of this choice, but I'd keep an eye to having multiple residues that can change identity in the future. We should make choices that will allow this to work well down the line.

jchodera:
I think this also fixes #362

dominicrufa:
with multiple residues changing (i.e. with protein mutations) , i think the setup will be handled differently, so there shouldn't be a conflict in that department

hannahbrucemacdonald:
> i think defaults should be 'MOL' for all of these.

Does that mean you think I should change it in `TopologyProposal`?

dominicrufa:
https://github.com/choderalab/perses/pull/593#issuecomment-545576295
that would work

dominicrufa:
> > i think defaults should be 'MOL' for all of these.
> 
> Does that mean you think I should change it in `TopologyProposal`?

that should work; it need only change in the `SmallMoleculeSetProposalEngine`

perses/annihilation/relative.py
```diff
@@ -370,13 +370,13 @@ def _get_core_atoms(self):
                                                                      self._new_to_hybrid_map, name_of_residue)
 
         #The union of the two will give the core atoms that can result from either new or old topology
-        # TODO: Shouldn't these sets be the same?
         total_core_atoms = core_atoms_from_old.union(core_atoms_from_new)
+        assert sorted(core_atoms_from_old) == sorted(core_atoms_from_new), 'Core atoms must match between old and new systems'
```
jchodera: Should this use a `set()` comparison instead?

perses/annihilation/relative.py
```diff
@@ -417,6 +417,8 @@ def _determine_core_atoms_in_topology(self, topology, unique_atoms, mapped_atoms
                         hybrid_index = hybrid_map[atom_index]
                         core_atoms.add(hybrid_index)
 
+        assert len(core_atoms) >= 3, 'Cannot run a simulation with fewer than 3 core atoms. System has {len(core_atoms)}'
```
jchodera: Do we need to be more strict than this? The core atoms also need to be bonded, `1-2-3`.

perses/rjmc/topology_proposal.py
```diff
@@ -797,7 +797,7 @@ def __init__(self,
                  logp_proposal=None,
                  new_to_old_atom_map=None, old_alchemical_atoms=None,
                  old_chemical_state_key=None, new_chemical_state_key=None,
-                 old_residue_name=None, new_residue_name=None,
+                 old_residue_name='MOL', new_residue_name='MOL',
```
jchodera: Do we really want to change the default here, rather than just in `relative_setup.py`? 

Could this have surprising behavior for other code, or break any other tests? 

Would it be better to require these be specified or raise an exception if either one is `None`?

perses/annihilation/relative.py
hannahbrucemacdonald: Ummm - yes if there are only three core atoms then they need to be bonded, but I don't think that it's true that it has to be 1-2-3 or 1-3-2 or 1-14-26 (if there are > 3 core atoms). This is just the minimum requirement that makes sure that core_atoms doesn't return empty

jchodera: Totally correct that order doesn't matter as long as they are bonded!
If it's not necessary to check this here, this should be fine!

perses/rjmc/topology_proposal.py
jchodera: It looks like it doesn't break any tests, so it's probably fine for now!

hannahbrucemacdonald: I'm happy to make these required if that's better?

perses/annihilation/relative.py
hannahbrucemacdonald: I'll change this

perses/rjmc/topology_proposal.py
jchodera: How about we open a new issue to remind us to consider doing that later and just move on now?

hannahbrucemacdonald:
Closes #592
Closes #362

PR #601, opened 2019-12-05
[WIP] Experiments

dominicrufa:
create a networkx graph on which to perform computation


perses/app/experiments.py
```diff
@@ -0,0 +1,1301 @@
...
+        """
+        Parse the ligand input.
+        Creates the following attributes:
+        1. self.ligand_oemol_pos_top : list of tuple(oemol, position, topology)
+        2. self.ligand_ffxml : xml for ligands
+        3. self.smiles_list : list of smiles
+        4. self.ligand_md_topologies : list of mdtraj.Topology objects
+                                       corresponding to self.ligand_oemol_pos_top topologies.
+        """
+        self.ligand_oemol_pos_top = []
```
jchodera: This is a really weird data structure. Why have a tuple like this rather than storing these objects in more sensible data structures, like `self.molecule_oemols`, `self.molecule_positions`, and `self.molecule_topologies`? I think this just causes lots of confusion to group them in weird ways.

dominicrufa: for each molecule, we need to store a ligand oemol, positions, and openmm.Topology object.  after i created this, i realized we also needed to store a mdtraj.Topology, and forgot to make the it a 4-tuple, but i like the suggested lists more.  will change

jchodera:
@dominicrufa : Some quick thoughts on the refactoring/design in this PR that might be good to review with @andrrizzi :

*Docstring examples:* Please add docstring examples for how to use the class/object and its methods! This is the first thing we should do as we're contemplating the design, which should be driven by how we want someone to use it.

*Design patterns:* Classes should be named after what they do, and ideally we should use well-known design patterns. For example, [`BuildProposalNetwork`](https://github.com/choderalab/perses/blob/experiments/perses/app/experiments.py#L42-L48) should either be a factory or an object that can be configured and then initialized, but it's somewhere in between right now. If we want to keep it as a network that can be initialized, let's call it `Network`, and let it be created, configured, and then initialized like
```python
network = Network(...)
network.option = 'myoption'
# Initialize the network
network.initialize()
# Now trying to change options will result in exceptions
network.option = 'myoption2'
... returns an exception
```
If we want it to be a factory, it should be a `NetworkFactory` that has a method that returns a `Network`:
```python
network_builder = NetworkBuilder(...)
network_builder.option = 'myoption'
network = network_builder.create_network(stuff)
```

*Defaults:* Currently, all the defaults [live in one place](https://github.com/choderalab/perses/blob/experiments/perses/app/experiments.py#L49-L143). It's going to be much, much easier on us if the defaults for these classes all live within the classes themselves. In YANK, for example, the YAML file exists just to override defaults for subclasses, but the defaults are otherwise as defined in the subclasses. In the current refactor, to add one more default argument to a class, we have to also add it to this section of `experiments.py`, which makes it much less extensible or modular since we need to modify multiple files to extend the behavior.

*Modularity:* We currently hard-code [the phases of interest](https://github.com/choderalab/perses/blob/experiments/perses/app/experiments.py#L150), which is going to make it much harder to consider other kinds of phases (e.g. octanol, membrane, amorphous solid) in the future. Is there a way we can make it easier to define phases as objects that subclass some base `RelativeTransofrmationPhase` object? If so, we can easily define new phases for different applications in the future, and simply cobble together the phases we want in the YAML file. The [testsystems](https://github.com/choderalab/perses/blob/master/perses/tests/testsystems.py#L175-L187) still have some of this modular design inspiration, where we define different phases as different keys of a `dict` for a real calculation that spans multiple thermodynamic states.

*Method names:* [`setup_engines`](https://github.com/choderalab/perses/blob/experiments/perses/app/experiments.py#L176-L243) ends up not being a super-informative name, since it really configures the class and initializes all the internal data structures. If we switch to a factory idiom, this should return a `Network`; otherwise, it should configure the internal data fields and set some sort of `self._initialized = True` flag so we know the object has been initialized.

*Defaults in calls:* Instead of [explicitly specifying defaults](https://github.com/choderalab/perses/blob/experiments/perses/app/experiments.py#L266-L274), let's keep those defaults defined in the classes / methods so that we don't need to modify multiple files every time we add one more option. This is a recipe for developer mistakes otherwise.

*Simple public API:* We have multiple public API calls like [`create_network`](https://github.com/choderalab/perses/blob/experiments/perses/app/experiments.py#L278-L356) and [setup_engines`](https://github.com/choderalab/perses/blob/experiments/perses/app/experiments.py#L176-L243)  where the distinction between them and how to use them is not clear. Let's try to simplify the API.

*Edge manipulation:* Could something like [this](https://github.com/choderalab/perses/blob/experiments/perses/app/experiments.py#L359) be done with a dict-like interface? Or something simpler like `network.set_edge_weight(weight)`, or even a property like `network.edge_weight[edge] = weight`?

*Private vs public API:* Should something like [`create_network_edge`](https://github.com/choderalab/perses/blob/experiments/perses/app/experiments.py#L396-L403) be part of the public API? Do we expect someone to call it? If not, and its only used internally, we prefix with `_` to make it private.

*Code duplication:* There seems to be a lot of code duplication in how we set up phases. We may want to try to minimize that by streamlining how we define phases. One way to do this is the base class idea above, where we just have to overload some methods to change behavior for different phases.

More later!

jchodera:
@dominicrufa @hannahbrucemacdonald : Can we loop @jaimergp into the API design conversation for this new functionality?

I'd really like to see us discuss the example code for use cases before we move any further with implementation, since it is likely some other large-scale changes will be needed.

perses/app/example.py
```diff
@@ -1,3 +1,670 @@
...
+#add pre-specified topology_proposal, hybrid_factory, geometry_engine, and sampler arguments from a yaml
+yaml_file = open('setup.yaml', 'r')
+setup_options = yaml.load(yaml_file, Loader=yaml.FullLoader)
+yaml_file.close()
+
+network = FreeEnergyNetwork(system_generator,
+                            ligands, #list of ligands (to which more ligands can be appended)
+                            transformation_specifications = setup_options, #a yaml dict that has prespecified parameters for proposals and/or samplers that overwrite defaults
+                            environments = [complex_environment, aqueous_environment, octanol_environment, complex2_environment] #a list of environments
+                            )
```
dominicrufa, line 65: there needs to be a `factory` object that actually builds the `FreeEnergyNetwork`.

jchodera:
@mikemhenry : This PR contains updated/refactored code you will need to pull into your API updates, but we should not merge this PR itself since we are continuing to change the API.

mikemhenry:
@jchodera any other code/PRs that I should base my API refactoring on?

mikemhenry:
Going to close this PR as we are starting up our new API refactor, but will want to use gufe protocols and objects

PR #605, opened 2019-12-10
Changing treatment of endstates

hannahbrucemacdonald:
This PR:

* adds in analysis code
* colors residue mapping
* tidies the treatment of multistate endstates and increases consistency between SAMS and REPEX.

### DIFF
--- a/perses/app/relative_setup.py
+++ b/perses/app/relative_setup.py
@@ -82,6 +82,7 @@ class RelativeFEPSetup(object):
         self._pressure = pressure
         self._temperature = temperature
         self._barostat_period = 50
+        self._pme_tol = 1e-04
         self._padding = solvent_padding
         self._hmass = hmass
         _logger.info(f"\t\t\t_hmass: {hmass}.\n")
@@ -205,9 +206,9 @@ class RelativeFEPSetup(object):
                 barostat = None
                 _logger.info(f"omitted MonteCarloBarostat.")
             self._system_generator = SystemGenerator(forcefield_files, barostat=barostat,
-                                                     forcefield_kwargs={'removeCMMotion': False, 'nonbondedMethod': self._nonbonded_method,'constraints' : app.HBonds, 'hydrogenMass' : self._hmass})
+                                                     forcefield_kwargs={'removeCMMotion': False, 'ewaldErrorTolerance': self._pme_tol, 'nonbondedMethod': self._nonbonded_method,'constraints' : app.HBonds, 'hydrogenMass' : self._hmass})
         else:
-            self._system_generator = SystemGenerator(forcefield_files, forcefield_kwargs={'removeCMMotion': False,'nonbondedMethod': self._nonbonded_method,'constraints' : app.HBonds, 'hydrogenMass' : self._hmass})
+            self._system_generator = SystemGenerator(forcefield_files, forcefield_kwargs={'removeCMMotion': False, 'ewaldErrorTolerance': self._pme_tol, 'nonbondedMethod': self._nonbonded_method,'constraints' : app.HBonds, 'hydrogenMass' : self._hmass})
 
         _logger.info("successfully called TopologyProposal.SystemGenerator to create ligand systems.")
         self._system_generator._forcefield.loadFile(StringIO(ffxml))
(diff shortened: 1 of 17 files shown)


hannahbrucemacdonald:
It seems like I've broken something in the topology_proposal or geometry engine... The test that's failing is `test_logp_forward_check_for_vacuum_topology_proposal` which is propane -> octane. Adding the assertion that mapping atoms requires matching of `HvyDegree` will result in one CH3 group on the propane being mapped onto a CH2 of the octane, but not sure why this would break the test

hannahbrucemacdonald:
@dominicrufa do you have any idea?

dominicrufa:
> @dominicrufa do you have any idea?

i'll give it a look.  wrote this many months ago

dominicrufa:
> It seems like I've broken something in the topology_proposal or geometry engine... The test that's failing is `test_logp_forward_check_for_vacuum_topology_proposal` which is propane -> octane. Adding the assertion that mapping atoms requires matching of `HvyDegree` will result in one CH3 group on the propane being mapped onto a CH2 of the octane, but not sure why this would break the test

according to https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemConstants/OEExprOpts.html#OEChem::OEExprOpts::HvyDegree, shoudn't the CH3 _not_ be able to map to a CH2?

hannahbrucemacdonald:
Yes - we would not now expect CH2 to map onto a CH3, as they have HvyDegrees of 2 and 1 respectively. I think for now we should have them not-map, but would need to run a bunch of tests to see what's a better mapping criterion overall

jchodera:
@hannahbrucemacdonald : I don't think [this section](https://github.com/choderalab/perses/pull/605/files#diff-9e4ec57992478a60e90d8ff1c6247ab4R48) actually does anything:
```python
        context, _ = context_cache.get_context(compound_thermodynamic_state)
        platform = context.getPlatform()
        logger.info('Setting the platform precision to mixed')
        platform.setPropertyDefaultValue('Precision','mixed')
```
because you can't change the precision _after_ a `Context` has been created. You have to set the Platform property before creating the Context.

perses/app/relative_setup.py
```diff
@@ -82,6 +82,7 @@ def __init__(self, ligand_input, old_ligand_index, new_ligand_index, forcefield_
         self._pressure = pressure
         self._temperature = temperature
         self._barostat_period = 50
+        self._pme_tol = 1e-04
```
hannahbrucemacdonald: @jchodera PME tolerance is set here

jchodera:
@dominicrufa: Looks like this is ready to merge after you review!

perses/app/setup_relative_calculation.py
```diff
@@ -51,6 +52,9 @@ def getSetupOptions(filename):
     if 'protocol-type' not in setup_options:
         setup_options['protocol-type'] = 'default'
 
+    if 'run_type' not in setup_options:
```
dominicrufa: this is specific to `neq`, since often we do not want to run equilibration followed by annealing, but instead resample from a subset of decorrelated equilibrium snapshots that are saved to disk

perses/rjmc/topology_proposal.py
```diff
@@ -51,12 +51,12 @@
 # default atom expression, requires same aromaticitiy and hybridization
 # bonds need to match in bond order
 # ethane to ethene wouldn't map, CH3 to NH2 would map but CH3 to HC=O wouldn't
-DEFAULT_ATOM_EXPRESSION = oechem.OEExprOpts_Hybridization
+DEFAULT_ATOM_EXPRESSION = oechem.OEExprOpts_Hybridization | oechem.OEExprOpts_HvyDegree
```
dominicrufa: issue that this caused is resolved

perses/app/setup_relative_calculation.py
```diff
@@ -493,6 +502,18 @@ def run_setup(setup_options):
...
+
+    #### configure platform
+    platform = utils.get_fastest_platform()
+    platform_name = platform.getName()
+    platform = openmm.Platform.getPlatformByName(platform_name) 
+    if platform_name in ['CUDA','CPU']:
+        platform_precision = 'mixed'
+        platform.setPropertyDefaultValue('Precision', platform_precision)
+    else:
+        _logger.warning(f'Platform {platform_name} may not support mixed precision, therefore using default')
```
dominicrufa: this code is redundant with https://github.com/choderalab/perses/blob/4ee5f1cceb3e733bd42e743cadd248e944177ff0/perses/dispersed/utils.py#L59

perses/samplers/multistate.py
```diff
@@ -133,117 +51,87 @@ def setup(self, n_states, temperature, storage_file, minimisation_steps=100,
...
              # now generating a sampler_state for each thermodyanmic state, with relaxed positions
             context, context_integrator = context_cache.get_context(compound_thermodynamic_state_copy)
-            feptasks.minimize(compound_thermodynamic_state_copy,sampler_state) 
+            feptasks.minimize(compound_thermodynamic_state_copy,sampler_state)
             sampler_state_list.append(copy.deepcopy(sampler_state))
 
-         #nonalchemical_thermodynamic_states = [
-        #    ThermodynamicState(self._factory._old_system, temperature=temperature),
-        #    ThermodynamicState(self._factory._new_system, temperature=temperature)]
+        if endstates:
```
dominicrufa: perhaps this can go in a standalone function in `utils.py` since both neq and staged equilibrium methods need to be able to compute (m)BAR estimates w.r.t. endstates?

PR #606, opened 2019-12-10
changing map to break with different ring sizes

hannahbrucemacdonald:
Merging ring-forming branches


hannahbrucemacdonald:
I'm going to delete this PR, and instead open a PR for `ring-map` -> `master`

PR #608, opened 2019-12-20
Use LJPME for endstate corrections instead of expanded cutoff

jchodera:
Instead of expanding cutoffs, it seems simpler to just use LJPME and a tighter PME tolerance.

In this implementation, we
- move all alchemical atom LJ parameters from CustomNonbondedForce back to NonbondedForce
- delete CustomNonbondedForce (since it is no longer needed)
- set PME tolerance to 1e-5
- enable LJPME to handle long-range dispersion correction in physically reasonable manner

I'm not sure which test(s) to run, so this is implemented as best I can without running any tests, and may still need some debugging.

See the warning that the platform precision may not actually be set to `mixed` using the scheme you had been using:
```python
        # WARNING: THIS DOES NOT ACTUALLY DO ANYTHING because we can't set the precision after context creation
        # TODO: FIX THIS BY MOVING IT TO THE INITIAL SETUP
        # SEE https://github.com/choderalab/yank/blob/3ac34a66883941492825450aeddf2e5be8bc848e/Yank/experiment.py#L2218-L2304
        context, _ = context_cache.get_context(compound_thermodynamic_state)
        platform = context.getPlatform()
        logger.info('Setting the platform precision to mixed')
        platform.setPropertyDefaultValue('Precision','mixed')
```
I think this needs to be set _before_ you create any `Context`s with the `Platform`. You'll want to see [how YANK does this](https://github.com/choderalab/yank/blob/3ac34a66883941492825450aeddf2e5be8bc848e/Yank/experiment.py#L2218-L2304) and consult with @andrrizzi.

### DIFF
--- a/perses/app/setup_relative_calculation.py
+++ b/perses/app/setup_relative_calculation.py
@@ -192,23 +192,6 @@ def getSetupOptions(filename):
 
     return setup_options
 
-
-def get_max_expanded_cutoff(htf):
-    minimum_lengths_per_phase = []
-    for phase in htf.keys():
-        if phase == 'vacuum':
-            continue  # don't need expanded cutoff for vacuum phase
-        hybrid_system = htf[phase].hybrid_system
-        box_vectors = hybrid_system.getDefaultPeriodicBoxVectors()
-        dimensions = [x[i] for i,x in enumerate(box_vectors)]
-        minimum_length = min(dimensions)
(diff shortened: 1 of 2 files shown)


jchodera:
@hannahbrucemacdonald: I'm not sure where you set it, but you should be able to go back to simulating all the other states with a PME tolerance of 1e-4, which is *much* faster than 1e-5.

PR #616, opened 2020-01-06
Minimal changes to use new SystemGenerator for relative free energy calculations

jchodera:
This is a WIP PR to modify the minimal amount of code to let us use the new [`SystemGenerator` from `openmm-forcefields`](http://github.com/choderalab/openmm-forcefields).

I've duplicated the CDK2 test that @hannahbrucemacdonald pointed me to and am modifying the relative setup and YAML files to allow us to use any GAFF or SMIRNOFF force fields.


.travis.yml
```diff
@@ -69,7 +73,6 @@ script:
 env:
   global:
     - ORGNAME="omnia"
-    - OE_LICENSE="$HOME/oe_license.txt"
```
hannahbrucemacdonald: Is there a reason you've removed this?

perses/app/setup_relative_calculation.py
```diff
@@ -53,6 +53,19 @@ def getSetupOptions(filename):
...
         setup_options['protocol-type'] = 'default'
 
+    if 'small_molecule_forcefield' not in setup_options:
+        setup_options['small_molecule_forcefield'] = 'gaff-2.11'
+
+    if 'small_molecule_parameters_cache' not in setup_options:
+        setup_options['small_molecule_parameters_cache'] = None
+
+    # Not sure why these are needed
+    # TODO: Revisit these?
```
hannahbrucemacdonald: These are related to RJMC I believe, as to whether angle/dihedral terms go into the the jump, or into the alchemical protocol -- @dominicrufa will know better

perses/data/cdk2-example/cdk2_setup_neq.yaml
```diff
@@ -0,0 +1,67 @@
...
+# for biopolymers and solvents.
+# Note that small molecule force field files should NOT be included here.
+forcefield_files:
+    - amber/protein.ff14SB.xml # ff14SB protein force field
+    - amber/tip3p_standard.xml # TIP3P and recommended monovalent ion parameters
+    - amber/tip3p_HFE_multivalent.xml # for divalent ions
+
+# Small molecule force field
+# We recommnd one of ['gaff-1.81', 'gaff-2.11', 'smirnoff99Frosst-1.1.0', 'openff-1.0.0']
+small_molecule_forcefield: gaff-2.11
```
hannahbrucemacdonald: what would happen if a small molecule FF file was provided in `forcefield_files` too?

.travis.yml
jchodera: Yes! Because it also appears again on line 78, six lines below. I was just removing the duplicate.

perses/data/cdk2-example/cdk2_setup_neq.yaml
jchodera: Either the `small_molecule_forcefield` will win out or an Exception will be thrown if atom types are redefined by the selected force field. I'd advise against including a small molecule FF in the `forcefield_files`.

perses/app/setup_relative_calculation.py
jchodera: Thanks! I'll leave them for now!

In the future, I wonder if we could do something similar to YANK where each module has its own configuration options in a subsection of the YAML file, and can process its own `dict` representation of a subsection of the YAML file. That way, we don't need a monolithic method that knows about all possible options, but instead just dispatches `kwargs` to each submodule.

hannahbrucemacdonald: Wait until you hear about experiments class

perses/app/relative_setup.py
```diff
@@ -546,13 +555,20 @@ def _solvate_system(self, topology, positions, model='tip3p',vacuum=False):
         solvated_system : openmm.System
             The parameterized system, containing a barostat if one was specified.
         """
+        # DEBUG: Write PDB file being fed into Modeller to check why MOL isn't being matched
```
hannahbrucemacdonald: Useful for now, but should be taken out before merge

jchodera:
@hannahbrucemacdonald : Is our plan to merge this, or keep this branch separate and have the `experiments` PR replace it?

PR #619, opened 2020-01-12
[WIP] Update experiments branch to use new SystemGenerator

jchodera:
This WIP PR will switch the experiments classes to use the new `SystemsGenerator`.


jchodera:
Whoops, wrong base!

PR #620, opened 2020-01-12
Update experiments branch to use new SystemGenerator

jchodera:
This WIP PR will switch the experiments classes to use the new SystemsGenerator.


jchodera:
@dominicrufa : The test now passes locally!
```bash
pip uninstall --yes perses; python setup.py install
nosetests -vv --nocapture perses/tests/test_experiments.py
```

perses/app/experiments.py
```diff
@@ -52,7 +52,8 @@ class BuildProposalNetwork(object):
                          'hmass': 4 * unit.amus,
                          'map_strength': 'default',
                          'phases': ['vacuum', 'solvent', 'complex'],
-                         'forcefield_files': ['gaff.xml', 'amber14/protein.ff14SB.xml', 'amber14/tip3p.xml'],
+                         'forcefield_files': ['amber/protein.ff14SB.xml', 'amber/tip3p_standard.xml', 'amber/tip3p_HFE_multivalent.xml'],
+                         'small_molecule_forcefield' : 'openff-1.0.0',
```
dominicrufa: is this `gaff`?

perses/app/experiments.py
```diff
@@ -878,10 +880,10 @@ def _solvate(self, topology, positions, model = 'tip3p', vacuum = False):
         modeller = app.Modeller(topology, positions)
         hs = [atom for atom in modeller.topology.atoms() if atom.element.symbol in ['H'] and atom.residue.name not in ['MOL','OLD','NEW']]
         modeller.delete(hs)
```
dominicrufa: i think that stripping and readding hydrogens is problematic.  will make a note to change in `Experiments`

perses/app/experiments.py
```diff
@@ -656,6 +657,9 @@ def _parse_ligand_input(self):
                 [oemol.SetTitle("MOL") for oemol in oemols]
                 self.smiles_list = [ oechem.OECreateSmiString(oemol, oechem.OESMILESFlag_DEFAULT | oechem.OESMILESFlag_Hydrogens) for oemol in oemols]
                 self.ligand_oemol_pos_top = [[oemol, position, forcefield_generators.generateTopologyFromOEMol(oemol)] for oemol, position in zip(oemols, positions)]
+                # Create openforcefield Molecule objects
+                from openforcefield.topology import Molecule
+                self.molecules = [ Molecule.from_openeye(oemol) for oemol in oemols ]
```
dominicrufa: is this just for the purposes of caching the molecules?

dominicrufa: because i think that the vacuum, solvated, and complex systems use the existing pipeline

jchodera:
Thanks for the comments! I realized I forgot to also add caching and some illustration of how to set the small molecule force field and cache paths. I'll add these.

perses/app/experiments.py
jchodera: Currently, we only need the `openforcefield.topology.Molecule` objects for the `SystemGenerator` to know what to match inside the topologies it sees. I cached it within this object for convenience, but as long as we can get it to the `SystemGenerator`, it doesn't have to be cached.

jchodera: Eventually, I'd love to move away from all of these other objects (OEMol, positions, OpenMM Topology) and just use `openforcefield` `Molecule`s since we can generate all of those things from `Molecule`.

perses/app/experiments.py
jchodera: Correct! I think @hannahbrucemacdonald has already addressed this in https://github.com/choderalab/perses/pull/616 via https://github.com/choderalab/perses/pull/616/commits/f4cf4355e24290dbd27f6d65f714107f60a7f258 so you can replicate that here.

perses/app/experiments.py
jchodera: No, this is the `openforcefield 1.0.0` ("Parsley") force field. 

If you'd rather the default be `gaff-2.11`, just change this to `gaff-2.11`. No preference here at the moment!

jchodera:
I think I've fixed things up, but I'm still noticing test failures that arise from the code we are trying to migrate away from (e.g. `generateOEMolFromTopologyResidue`):
```
ERROR: for each torsion, calculate position of atom1
  File "/home/travis/miniconda/envs/test/lib/python3.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/miniconda/envs/test/lib/python3.6/site-packages/perses/tests/test_geometry_engine.py", line 1299, in test_existing_coordinates
    sys, pos, top = oemol_to_openmm_system(molecule2, molecule_name_2)
  File "/home/travis/miniconda/envs/test/lib/python3.6/site-packages/perses/tests/test_geometry_engine.py", line 887, in oemol_to_openmm_system
    system = system_generator.build_system(topology)
  File "/home/travis/miniconda/envs/test/lib/python3.6/site-packages/perses/rjmc/topology_proposal.py", line 2115, in build_system
    system = self._forcefield.createSystem(new_topology, **self._forcefield_kwargs)
  File "/home/travis/miniconda/envs/test/lib/python3.6/site-packages/simtk/openmm/app/forcefield.py", line 1164, in createSystem
    if generator(self, res):
  File "/home/travis/miniconda/envs/test/lib/python3.6/site-packages/openmoltools/forcefield_generators.py", line 514, in gaffTemplateGenerator
    molecule = generateOEMolFromTopologyResidue(residue)
  File "/home/travis/miniconda/envs/test/lib/python3.6/site-packages/openmoltools/forcefield_generators.py", line 148, in generateOEMolFromTopologyResidue
    infile = open(ac_output_filename)
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmp0pp__9u3/molecule-after-bond-perception.ac'
```
Should we work on eliminating those in this PR, or a separate PR? Ideally, we would _never_ try to go from a lossy format (OpenMM `Topology`) to `OEMol` or `Molecule` objects, since we don't have the information to do this correctly. But it feels the right way to do this might be to add residue perception to the `openforcefield` `Topology` object and migrate to that instead.

perses/app/experiments.py
dominicrufa: > Currently, we only need the `openforcefield.topology.Molecule` objects for the `SystemGenerator` to know what to match inside the topologies it sees. I cached it within this object for convenience, but as long as we can get it to the `SystemGenerator`, it doesn't have to be cached.

ah right.  i understand now!

dominicrufa:
> I think I've fixed things up, but I'm still noticing test failures that arise from the code we are trying to migrate away from (e.g. `generateOEMolFromTopologyResidue`):
> 
> ```
> ======================================================================
> ERROR: for each torsion, calculate position of atom1
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/home/travis/miniconda/envs/test/lib/python3.6/site-packages/nose/case.py", line 197, in runTest
>     self.test(*self.arg)
>   File "/home/travis/miniconda/envs/test/lib/python3.6/site-packages/perses/tests/test_geometry_engine.py", line 1299, in test_existing_coordinates
>     sys, pos, top = oemol_to_openmm_system(molecule2, molecule_name_2)
>   File "/home/travis/miniconda/envs/test/lib/python3.6/site-packages/perses/tests/test_geometry_engine.py", line 887, in oemol_to_openmm_system
>     system = system_generator.build_system(topology)
>   File "/home/travis/miniconda/envs/test/lib/python3.6/site-packages/perses/rjmc/topology_proposal.py", line 2115, in build_system
>     system = self._forcefield.createSystem(new_topology, **self._forcefield_kwargs)
>   File "/home/travis/miniconda/envs/test/lib/python3.6/site-packages/simtk/openmm/app/forcefield.py", line 1164, in createSystem
>     if generator(self, res):
>   File "/home/travis/miniconda/envs/test/lib/python3.6/site-packages/openmoltools/forcefield_generators.py", line 514, in gaffTemplateGenerator
>     molecule = generateOEMolFromTopologyResidue(residue)
>   File "/home/travis/miniconda/envs/test/lib/python3.6/site-packages/openmoltools/forcefield_generators.py", line 148, in generateOEMolFromTopologyResidue
>     infile = open(ac_output_filename)
> FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmp0pp__9u3/molecule-after-bond-perception.ac'
> ```
> 
> Should we work on eliminating those in this PR, or a separate PR? Ideally, we would _never_ try to go from a lossy format (OpenMM `Topology`) to `OEMol` or `Molecule` objects, since we don't have the information to do this correctly. But it feels the right way to do this might be to add residue perception to the `openforcefield` `Topology` object and migrate to that instead.

this is strange. i haven't touched `test_geometry` in a while, but i dont remember these functions being called;  i'll look again in the morning.

jchodera:
It's possible that I'm accidentally sweeping in a different AmberTools or ambermini (or some conflicting set of both). See https://github.com/openforcefield/openforcefield/issues/487

jchodera:
@dominicrufa : Do you need more changes before you merge this into the `experiments` branch? I want to make sure we don't have the two branches diverge too much so that I have to start from scratch.

I'm not quite sure what's causing the failures---they're all with parts of the code that I didn't touch, which I think involve `openmoltools`. I don't see both `ambertools` and `ambermini` installed----it looks like `ambermini` is installed in travis---but we eventually want to make sure _only_ `ambertools` is installed (and not `ambermini`). I can tweak the travis a bit to see if I can get that to work if you prefer.

jchodera:
Whoops, I've found at least one issue I introduced:
```
ERROR: test the creation of unsampled endstates
  File "/home/travis/miniconda/envs/test/lib/python3.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/miniconda/envs/test/lib/python3.6/site-packages/perses/tests/test_smc.py", line 232, in test_create_endstates
    fe_setup = RelativeFEPSetup(ligand_input = f"{os.getcwd()}/test.smi",
FileNotFoundError: [Errno 2] No such file or directory
```
I moved `test.smi` to `perses/data/` so the tests don't depend on which directory you run tests from. Will fix.

jchodera:
Ah! The issue seems to be that we were using paths relative to the current working directory, but since I started adding tests that execute in temporary directories, we no longer are in a valid directory when the tests that use `getcwd()` (which we should avoid) are run:
```
for each torsion, calculate position of atom1 ... sh: 0: getcwd() failed: No such file or directory
```
I'll fix this.

jchodera:
@dominicrufa : I think I've fixed this now.

The better idiom to use for "create a temporary directory, enter it, run something, and then exit when done" is:
```python
from perses.tests.utils import enter_temp_directory
with enter_temp_directory() as tmpdirname:
    # Do something
    ...
```

PR #622, opened 2020-01-23
Changing Hybrid{sams|repex}Sampler to accept variable number of sampler states

hannahbrucemacdonald:
This adds an optional argument called `n_samplers` which defines the number of sampler states to run with. If `n_samplers != n_states` it will pick approximately equally spaced sampler states from the list.

The default for n_samplers is to equal n_states, but possibly we would want the defaults to be:

REPEX n_samplers = n_states
SAMS n_samplers = 1

?

### DIFF
--- a/perses/app/setup_relative_calculation.py
+++ b/perses/app/setup_relative_calculation.py
@@ -85,6 +85,8 @@ def getSetupOptions(filename):
         if 'beta_factor' not in setup_options:
             setup_options['beta_factor'] = 0.8
             _logger.info(f"\t\t\tbeta_factor not specified: default to 0.8.")
+        if 'n_replicas' not in setup_options:
+            setup_options['n_replicas'] = 1
     elif setup_options['fe_type'] == 'repex':
         _logger.info(f"\t\tfe_type: repex")
         if 'offline-freq' not in setup_options:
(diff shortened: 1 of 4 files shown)


jchodera:
Can we distinguish between the number of samplers (replicas) and the number of thermodynamic states? The multistate samplers actually support multiple samplers (replicas) for any number of thermodynamic states for SAMS.

jchodera:
Oh, maybe I'm confused here and you're already distinguishing between the two?

hannahbrucemacdonald:
That’s what I’m trying to do, so n_states is the number of thermodynamic states and n_samplers is the number of sampler states.

at the moment perses would only allow number of thermodynamic == number sampler

hannahbrucemacdonald:
So I’m changing it so Ana can start doing some sams experiments

jchodera:
Awesome! Apologies for misunderstanding!

hannahbrucemacdonald:
If it’s confusing maybe there is a better choice of variable names. n_windows and n_samplers or something

jchodera:
Why not just use the sample terminology we use [in the multistate samplers themselves](https://openmmtools.readthedocs.io/en/0.18.1/api/generated/openmmtools.multistate.MultiStateSampler.html#openmmtools.multistate.MultiStateSampler.n_replicas) instead of inventing new terminology?

There, we use `n_replicas` to denote the number of replicas, and `n_states` to denote the number of thermodynamic states. The "sampler" actually refers to the whole `MutliStateSampler` object (or subclass) in this terminology, which is why I was confused about the ambiguity of using "sampler" to mean either the `MultiStateSampler` or a replica.

hannahbrucemacdonald:
ok the error at least isn't a seg fault any more, but it's now complaining that the output file from an antechamber simulation doesn't exist...

hannahbrucemacdonald:
Is it ok if we go ahead and merge the fix-sams branch for now? It does what it should do, and fixes at least part of the tests

jchodera:
Yes! I can fix the remaining tests separately in #623

PR #623, opened 2020-01-28
Update travis tests for new openmm-forcefields package to fix failing travis tests

jchodera:
@hannahbrucemacdonald: This may be why you're having trouble with #622

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,9 +58,6 @@ script:
   - conda install --yes --use-local ${PACKAGENAME}-dev
   # Install testing dependencies
   - conda install --yes --quiet nose nose-timer
-  # Install openmm-forcefields via pip (until available via omnia)
-  - conda install --yes tinydb "openforcefield>=0.6.0" "openforcefields>=1.0.0"
-  - pip install git+https://github.com/openmm/openmm-forcefields.git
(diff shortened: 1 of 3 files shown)


jchodera:
@hannahbrucemacdonald : I think all the failures are due to tests that use `generateOEMolFromTopologyResidue`, which we are trying to eliminate because it is buggy and unreliable. 

Perhaps this is a good time for me to rework all the tests that use this!

jchodera:
Hm, some package is still bringing in `parmed` separately, and I'm not sure which one it is:
```
  parmed             omnia/linux-64::parmed-3.2.0-py36_0
```
We need to get rid of that.

jchodera:
Ah, it's `openmmtools`:
```
(base) lski2838:conda-recipes choderaj$ grep parmed */meta.yaml
openmmtools/meta.yaml:    - parmed
parmed/meta.yaml:  name: parmed
parmed/meta.yaml:  home: http://parmed.github.io/ParmEd
perses/meta.yaml:    - parmed
perses/meta.yaml:    - parmed
protons/meta.yaml:    - parmed
protons/meta.yaml:    - parmed
solvationtoolkit/meta.yaml:    - parmed
solvationtoolkit/meta.yaml:    - parmed
```

PR #625, opened 2020-01-28
[MRG] Adding option for 'spectator' molecules, which are not alchemical

hannahbrucemacdonald:
Currently there is no way to input small-molecules that aren't going to be part of the alchemy.

This allows for a list of spectator molecules to be passed in, which is needed for a few of the JACS systems (i.e. Bace, which has a TLA in the system). Also will be useful for protein mutation stuff, where the ligand is vanilla.

The small molecule needs:
* to be in the same frame of reference of the protein/ligand
* it will only be added in the complex leg
* it will use the same small molecule FF as the alchemical small molecule forcefield


hannahbrucemacdonald:
I can add `Bace` as a test

hannahbrucemacdonald:
@dominicrufa can you check the test is ok? If so can you please merge

Issue #603, opened 2019-12-06
Mapping of rings not behaving as expected


hannahbrucemacdonald:
Currently have issue where rings of different sizes are being mapped onto eachother as it passes the check that 'all bonds in one are in the other'

`oechem.OEExprOpts_RingMember` is insufficient, because we also need to check that molecules aren't mapping onto rings of different sizes i.e.:

![image](https://user-images.githubusercontent.com/43652597/70334042-a2524500-1812-11ea-8f8d-8085ae29e722.png)

hannahbrucemacdonald:
Changing the breaking ring function so that it adds this check. As it is written, it will fail to map the above two molecules, and suggests the user write an atom map manually.

Although I'm not sure it's currently possible for this transformation to work, because it would require a ring system to be fully built in. This is currently an issue with some of the Bace ligands

hannahbrucemacdonald:
![image](https://user-images.githubusercontent.com/43652597/70334309-36bca780-1813-11ea-84f9-0c29e9d3088f.png)


Note atoms 12 13 14

hannahbrucemacdonald:
Attempting to sort it (so that it will fail during set up) in `ring-map` branch, but not sure how this conflicts with ring closure code

jchodera:
Fixed by #627

PR #627, opened 2020-02-03
[WIP] Ring map - checking _size_ of ring occupancy

hannahbrucemacdonald:
There is an issue with some Bace ligands, where because we are only checking if atoms are in rings or not, that we end up mapping atoms in rings of different sizes e.g.:

image


hannahbrucemacdonald:
@dominicrufa can you review please?

jchodera:
@hannahbrucemacdonald : I've retriggered the tests for you! 

I see a bunch of tests failing with:
```
perses.tests.test_topology_proposal.test_mapping_strength_levels ... sh: 0: getcwd() failed: No such file or directory
Cannot open file (ATOMTYPE.INF) with mode (w).
```
This is likely occurring because a temporary directory is being created and then deleted before some part of the antechamber process could be completed. I'm trying to figure out where this is happening, but I'm not sure how much time we want to invest in trying to make something we know is fundamentally flawed run, since it will still produce fundamentally flawed results.

jchodera:
Hm, this test runs locally. I think this might be an issue with another test that changes the current working directory, which might mess up subsequent tests if a test fails before completion.

jchodera:
I've fixed the tests that change the working directory, but haven't extensively looked for other methods that may also manipulate the working directory.

I also fixed some very broken logic where `atom_expr` and `bond_expr` were being used by `app/setup_relative_calculation.py:run_setup()`, but were not ever defined in that method. 

Hopefully this will get rid of the failures arising from `getcwd() failed: No such file or directory` and let you debug the rest!

jchodera:
Ah! We had been pinning `ambertools==19.0` in `devtools/conda-recipe`. I've fixed that, and also tried to suppress some other conda install noise.

jchodera:
I'm seeing many fewer failures now!

I'm not sure if we can suppress the `antechamber/bondtype` warnings about unknown bond types, but it would vastly clean up the test output if we could.

jchodera:
Yay! Looks like there are 9 failures now, and they're all relevant failures:
https://travis-ci.org/choderalab/perses/jobs/649939655?utm_medium=notification&utm_source=github_status

Passing this back to @hannahbrucemacdonald

hannahbrucemacdonald:
> Ah! We had been pinning ambertools==19.0 in devtools/conda-recipe. I've fixed that, and also tried to suppress some other conda install noise.


Sorry I pinned it a few days ago in an act of desperation to try fix the antechamber issues, I should have unpinned

hannahbrucemacdonald:
>I also fixed some very broken logic where atom_expr and bond_expr were being used by app/setup_relative_calculation.py:run_setup(), but were not ever defined in that method.

Sorry my bad - I was rushing to commit stuff last night, but thanks for fixing. Hopefully this should all be a bit better in the `Experiments` pr

jchodera:
@hannahbrucemacdonald : I've updated `openmoltools` to get rid of those bond warnings, so I'll rerun the tests and (hope) we can actually see the real errors now.

jchodera:
This test output is much more comprehensible!

I hadn't seen this warning before:
> Warning: maximum number of matches limit (1024) is reached

We should also really try to get the mutations PR in before we move further with the experiments PR.

hannahbrucemacdonald:
I think that warning is from openeye, when we are doing the matching criteria, sometimes if two molecules have a lot of ways to map, we see this. I'm doing my best to get this passing, but there are still a few broken tests that I'm worried about. Doing my best to get them squashed this week.

hannahbrucemacdonald:
perses.tests.test_samplers.test_hybrid_scheme
> this is failing because of a variable called `chain_id` which I think I can fix
perses.tests.test_relative.test_compare_energies
> this is broken because we've added the checking for ring memership for mapping. I need to add a atom mapping criteria to the test that lets us map napthalene to benzene
test_topology_proposal.test_mapping_strength_levels
> same as before, where I've changed what the mapping behavior is, what we consider the 'right' mapping has changed, and I need to update the tests accordingly

jchodera:
@hannahbrucemacdonald  :Can you open an issue (and assign it to me) to fix that failing protein mutation test you commented out? I can come back to it later!

hannahbrucemacdonald:
We have some passes and some fails... not completely sure what the difference between all the different CI environments are but I'll try fix it today, once and for all

hannahbrucemacdonald:
Closes #603

PR #633, opened 2020-02-12
adding openforcefield and openforcefields to GH-actions

hannahbrucemacdonald:


jaimergp:
Hm, I don't know if I am understanding correctly. Shouldn't those dependencies be part of the conda recipe?

hannahbrucemacdonald:
They aren't yet available on conda, I took the line from the travis build script - should have copied the comment over too...

  # Install openmm-forcefields via pip (until available via omnia)

https://github.com/choderalab/perses/blob/d54e6c23d5c402deb7cfd3e7417832168136ce0b/.travis.yml#L61

I'll fix the install so that it's in the test environment, rather than build!

jaimergp:
Ah, I see.

At some point we should consider upgrading to the latest `molssi/cookiecutter-cms`, maybe after getting the GHA PR merged there :)

hannahbrucemacdonald:
This fixes the tests that failed due to `openforcefield` not being present. Other failing tests should be fixed by `ring-map` and `openmoltools-0.8.5` so I'm going ahead and merging. Will open an issue that this should be fixed when the packages are available from conda

PR #640, opened 2020-02-21
🚀  exposing nonbonded_method expression

hannahbrucemacdonald:
Is there something I also need to do with the endstates if we are doing reaction-field? Just toggle them off?


jchodera:
@hannahbrucemacdonald : This ready to go?

hannahbrucemacdonald:
Yep! I'll merge now

PR #648, opened 2020-03-06
Fix atom mapper rendering

### LINKED ISSUE #647: Oemol rendering things as radicals
Need to check that the radicals are only an aesthetic error, rather than the charge being misinterpreted in the whole simulation.

The renderings can be done using: `perses/examples/atom-mapping/*ipynb`, but we are getting something that looks like this:

image

Which is Bace ligand 27 -> 29 (with zero indexing).

I will look into what net charge is associated to the oemol, but I suspect (and very much hope) that this is just a rendering problem, rather than a simulation issue

jchodera:
Fixes #647


hannahbrucemacdonald:
This looks good to me --- I'm just adding some logger lines in the script so net charge is printed in the pipeline too

PR #650, opened 2020-03-08
[WIP] Eliminate old SystemGenerator in favor of new openmmforcefields one

jchodera:
This PR intends to eliminate the old `SystemGenerator` from `perses` to instead use the [`openmmforcefields.generators.SystemGenerator`](https://github.com/openmm/openmmforcefields#automating-force-field-management-with-systemgenerator).

@dominicrufa : I've got a good deal of straightforward cleanup to do, but the one place where I'm stuck is trying to eliminate [`TopologyProposal._topology_to_smiles()`](https://github.com/choderalab/perses/blob/new_ring_closing/perses/rjmc/topology_proposal.py#L3041). Can we replace this with something that uses your new augmented topology object to get the `oemol` for the current residue under consideration? If so, any pointers on how to do that?

Once we've figured that out, I can quickly finish the rest of this PR!


jchodera:
From @dominicrufa : 
> once the topology proposal has been conducted, the TopologyProposal.new_topology and old_topology objects have attributes called residue_oemol (here: https://github.com/choderalab/perses/blob/b8c60e45b4798d099cad9624813bc57e02e0aa00/perses/rjmc/topology_proposal.py#L117), which can be used to call https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemFunctions/OECreateCanSmiString.html?highlight=smiles

jchodera:
@dominicrufa :Unfortunately, by the time we get your `augment_openmm_topology` calls, it’s too late for some use cases.

See, for example, the [`propose` method](https://github.com/choderalab/perses/blob/b8c60e45b4798d099cad9624813bc57e02e0aa00/perses/rjmc/topology_proposal.py#L3401
) in `PremappedSmallMoleculeSetProposalEngine`.

I think it might be best if we can just create a simple wrapper class that provides the corresponding  `openforcefield.topology.Molecule` object for each OpenMM Topology `Residue` object, which can be retrieved via `residue.to_molecule()` or `residue.molecule`. You'd construct it with
```python
augmented_topology = AugmentedTopology(topology, molecule=[molecule1, molecule2, molecule3])
```
where it would figure out all the amino acids automatically, but you'd need to specify as `openforcefield.topology.Molecule` objects any other small molecules that don't appear in the PDB.

It should be quick to code this up Monday. If I did this, what would you need?
* For biopolymer residues, you need just the atoms in the residue, or do you need capping groups on either end? If capping groups, which ones do you use?
* Do you need to make sure the atom ordering matches the atom ordering in the `Topology`. If there are capping groups, do they just go at the end? Do you have a map from Topology to OEMol atoms?
* You also use a NetworkX representation. We provide one for `openforcefield.topology.Molecule` objects via [`molecule.to_networkx()`](https://open-forcefield-toolkit.readthedocs.io/en/0.6.0/api/generated/openforcefield.topology.Molecule.html#openforcefield.topology.Molecule.to_networkx), which provides a [rich set of node and edge attributes](https://github.com/openforcefield/openforcefield/blob/master/openforcefield/topology/molecule.py#L2426-L2467). Would this be sufficient?

If you can quickly let me know the answers to these questions, I should be able to get you this quickly, and we can then get rid of a bunch of specialized code that exists just to juggle and pass OEMol objects, since you can then just retrieve `residue.to_molecule().to_openeye()` if you ever need to.

dominicrufa:
> @dominicrufa :Unfortunately, by the time we get your `augment_openmm_topology` calls, it’s too late for some use cases.
> 
> See, for example, the [`propose` method](https://github.com/choderalab/perses/blob/b8c60e45b4798d099cad9624813bc57e02e0aa00/perses/rjmc/topology_proposal.py#L3401) in `PremappedSmallMoleculeSetProposalEngine`.

i forgot about these _special_  proposals since they are never used outside of the parametrization/system builder pipelines (they only show up in some tests); however, I agree that your suggestion is best

> 
> I think it might be best if we can just create a simple wrapper class that provides the corresponding `openforcefield.topology.Molecule` object for each OpenMM Topology `Residue` object, which can be retrieved via `residue.to_molecule()` or `residue.molecule`. You'd construct it with
> 
> ```python
> augmented_topology = AugmentedTopology(topology, molecule=[molecule1, molecule2, molecule3])
> ```
> 
> where it would figure out all the amino acids automatically, but you'd need to specify as `openforcefield.topology.Molecule` objects any other small molecules that don't appear in the PDB.
> 
> It should be quick to code this up Monday. If I did this, what would you need?
> 
> * For biopolymer residues, you need just the atoms in the residue, or do you need capping groups on either end? If capping groups, which ones do you use?

> * Do you need to make sure the atom ordering matches the atom ordering in the `Topology`. If there are capping groups, do they just go at the end? Do you have a map from Topology to OEMol atoms?

for biopolymers, i _only_ need the atoms in the residue, no caps necessary.  as long as i have the residue in topology and an uncapped oemol where the atoms names in the topology and the oemol match, then the PointMutationEngine will automatically conduct a mapping and match the indices properly.

> * Will polymeric residues always use their residue names from the PDB (see the [Ligand Expo](http://ligand-expo.rcsb.org/) and [search](http://ligand-expo.rcsb.org/ld-search.html) or [download](http://ligand-expo.rcsb.org/ld-download.html) the master list of all components), or could they have the same name as a component but differ in some way (tautomer? protonation state?) or just have a random name?
right now, we only support a single standard amino acid for a selection of 3 letter codes.  in an ideal situation, I think I would prefer that each (i.e. protonation state) of an amino acid (or unnatural or PTMs) should have a different name, whether it's a 3 letter code or otherwise.  i don't know if it would be best that we have a cache in perses of all the polymeric residues that are implementable or otherwise

> * You also use a NetworkX representation. We provide one for `openforcefield.topology.Molecule` objects via [`molecule.to_networkx()`](https://open-forcefield-toolkit.readthedocs.io/en/0.6.0/api/generated/openforcefield.topology.Molecule.html#openforcefield.topology.Molecule.to_networkx), which provides a [rich set of node and edge attributes](https://github.com/openforcefield/openforcefield/blob/master/openforcefield/topology/molecule.py#L2426-L2467). Would this be sufficient?

i think this networkx representation would be sufficient.  I just have to change a few lines in the `geometry.py` script

> 
> If you can quickly let me know the answers to these questions, I should be able to get you this quickly, and we can then get rid of a bunch of specialized code that exists just to juggle and pass OEMol objects, since you can then just retrieve `residue.to_molecule().to_openeye()` if you ever need to.

jchodera:
This PR has been superceded by #618

PR #652, opened 2020-03-10
Code cleanup

hannahbrucemacdonald:
Adding docstrings and making everything a bit nicer.

### DIFF
--- a/perses/analysis/load_simulations.py
+++ b/perses/analysis/load_simulations.py
@@ -165,7 +165,7 @@ class Simulation(object):
         if self.bindingdg is not None:
             print(f'Relative binding free energy is {self.bindingdg} ({self.bindingddg})')
         if self.hydrationdg is not None:
-            print(f'Relative binding free energy is {self.hydrationdg} ({self.hydrationddg})')
+            print(f'Relative hydration free energy is {self.hydrationdg} ({self.hydrationddg})')
 
     # TODO fix this whole thing to be consistent with above
     def historic_fes(self, stepsize=100):
--- a/perses/analysis/load_simulations.py
+++ b/perses/analysis/load_simulations.py
@@ -2,6 +2,7 @@ import numpy as np
 import matplotlib.pyplot as plt
 import sys
 from openeye import oechem, oegraphsim
+from openmmtools.constants import kB
 
 import logging
 _logger = logging.getLogger()
@@ -16,7 +17,7 @@ class Simulation(object):
 
     This automatically loads the data and performs the data. Results can be accessed through attributes.
 
-    >>> # Load and analyze the simulation for `lig0to1/out.*`
+    >>> # Load and analyze the simulation for `lig0to1/*.nc`
     >>> simulation_results = Simulation(0,1)
     >>> # Report the binding free energy
     >>> print(f'Relative binding free energy: {simulation_results.bindingdg}')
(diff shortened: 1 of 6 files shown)


perses/analysis/load_simulations.py
```diff
@@ -2,55 +2,69 @@
...
+    count : int
+        Number of times the 'histories' have been 'sampled' from
+    hydrationdg : float
+        relative hydration free energy (kcal/mol)
+    hydrationddg : float
+        variance in relative hydration free energy (kcal/mol)
+    bindingdg : float
+        relative hydration free energy (kcal/mol)
+    bindingddg : float
+        variance in relative hydration free energy (kcal/mol)
```
jchodera: According to your computation below:
```
            self.bindingddg = (self.comddg + self.solddg)**0.5
```
this is the standard error, and *not* the variance.

It might cause less confusion if you always stored the standard error instead?

perses/analysis/load_simulations.py
```diff
@@ -2,55 +2,69 @@
...
+    solddg_history : type
+        variance in solvent free energy at equally spaced intervals of simulation (kcal/mol)
+    comddg_history : type
+        variance in complex free energy at equally spaced intervals of simulation (kcal/mol)
+    count : int
+        Number of times the 'histories' have been 'sampled' from
+    hydrationdg : float
+        relative hydration free energy (kcal/mol)
+    hydrationddg : float
+        variance in relative hydration free energy (kcal/mol)
```
jchodera: According to your computation below:
```
            self.hydrationddg = (self.vacddg + self.solddg)**0.5
```
this is the standard error, and *not* the variance.

It might cause less confusion if you always stored the standard error instead?

perses/analysis/load_simulations.py
```diff
@@ -138,7 +152,20 @@ def load_data(self):
...
 
 
-    def historic_fes(self,stepsize=100):
+    def historic_fes(self, stepsize=100):
+        """ Function that performs mbar at intervals of the simulation
+        by postprocessing. Can be slow if stepsize is small
+
+        Parameters
+        ----------
+        stepsize : int
```
jchodera: Should be
```
stepsize : int, optional, default=100
```

perses/analysis/load_simulations.py
```diff
@@ -2,55 +2,69 @@
...
+_logger.setLevel(logging.INFO)
+_logger = logging.getLogger("load_simulations")
 
-    def heavy_atom_count(self):
-        """ Counts the number of heavy atoms in an oemol
 
-        Parameters
-        ----------
+class Simulation(object):
+    """Object that holds the results of a free energy simulation.
```
jchodera: Can you note that this automatically loads the data and performs the analysis, making the results available through attributes below?

A quick docstring example would be great:
```python
>>> # Load and analyze the simulation for `lig0to1/out.*`
>>> sim = Simulation(0, 1)
>>> # Report the binding free energy
>>> print(f'Binding free energy = {sim.bindingdg} +- {sim.bindingddg}')
```

perses/analysis/load_simulations.py
```diff
@@ -86,8 +100,8 @@ def __init__(self,A,B):
 
     @staticmethod
     def _kt_to_kcal(x):
-        q = unit.quantity.Quantity(x, unit = unit.kilojoules_per_mole)
-        return q.in_units_of(unit.kilocalories_per_mole)._value
+        q = unit.quantity.Quantity(x, unit=unit.kilojoules_per_mole)
+        return q / unit.kilocalories_per_mole
 
     def load_data(self):
```
jchodera: Should this be private (`_load_data`)?

perses/analysis/load_simulations.py
```diff
@@ -199,101 +238,91 @@ def sample_history(self,method='binding'):
...
-            sim = simulation(b, a)
-            all_simulations.append(sim)
-        else:
-            print(f'Output directory lig{b}to{a} doesnt exist')
-
-    return all_simulations
-
-if __name__ == '__main__':
-    run(sys.argv[1])
+#def get_experimental(molecules, i,j):
```
jchodera: Feel free to delete this code if we don't need it. We can always come back and rescue it if we need it.

perses/analysis/load_simulations.py
```diff
@@ -2,55 +2,77 @@
...
+    ----------
+    ligA : int
+        integer of first ligand
+    ligB : int
+        integer of first ligand
+    directory : string
+        output directory location : lig{A}to{B}
+    vacdg : float
+        relative free energy in vacuum phase (kcal/mol)
+    vacddg : float
```
jchodera: @hannahbrucemacdonald : These docstrings are now incorrect. According to the code, you are storing variances for  `vacddg`, `solddg`, and `comdd`, but standard errors for `hydrationddg` and `complexddg`.

What if you always stored standard errors instead, and remembered that you have to propagate the error in `A+B` as `sqrt(dA**2 + dB**2)`?

jchodera:
Is this ready to merge @hannahbrucemacdonald  or will it cause conflicts with #618?

PR #655, opened 2020-03-24
Updating small molecule pipeline to fit into the new protein-mutation stuff

hannahbrucemacdonald:
This PR

- [x] passes mapping criteria into `propose` function, rather than init statement of `SmallMoleculeSetProposalEngine `
- [x] turns off `validate_energy_bookkeeping` in the pipeline, so that passes for the JACS set, where sometimes entire rings are deleted**
- [x] removing bug where solvent phase was ignoring user defined nonbonded method
- [x] calling `augment_openmm_topology` for 'secondary' phases - i.e. if topology_proposal has already been called for a different phase
- [x] updating yaml loader line to get rid of the depreciated warning
- [x] replacing LangevinSplittingDynamicsMove with LangevinDynamicsMove***
- [x] running `compute_potential_components` and `validate_endstate_energies` on Reference, as it's more precise than CUDA, even though it's slower****
- [x] calling `_assign_ring_ids` irrespective of ring breaking or not* . If IntType isn't in the atom expression then this has no effect anyway
- [x] avoiding calling `SmallMoleculeSetProposalEngine._calculate_probability_matrix` if only two molecules are there
- [x] fixing small indexing bug

** I am opening Issue #656 to try understand what types of things count as ring breaking, and how we can juggle all different sorts of things simultaneously
*** This means that `constraint_tolerance=1e-06` has been taken out
**** I can just try run this on one of complex/solvent to get a speed up

### DIFF
--- a/perses/app/setup_relative_calculation.py
+++ b/perses/app/setup_relative_calculation.py
@@ -520,7 +520,7 @@ def run_setup(setup_options):
             #TODO expose more of these options in input
             if setup_options['fe_type'] == 'sams':
                 hss[phase] = HybridSAMSSampler(mcmc_moves=mcmc.LangevinDynamicsMove(timestep=timestep,
-                                                                                    collision_rate=5.0 / unit.picosecond,
+                                                                                    collision_rate=1.0 / unit.picosecond,
                                                                                     n_steps=n_steps_per_move_application,
                                                                                     reassign_velocities=False,
                                                                                     n_restart_attempts=20),
@@ -530,7 +530,7 @@ def run_setup(setup_options):
                 hss[phase].setup(n_states=n_states, n_replicas=n_replicas, temperature=temperature,storage_file=reporter,lambda_protocol=lambda_protocol,endstates=endstates)
             elif setup_options['fe_type'] == 'repex':
                 hss[phase] = HybridRepexSampler(mcmc_moves=mcmc.LangevinDynamicsMove(timestep=timestep,
-                                                                                     collision_rate=5.0 / unit.picosecond,
+                                                                                     collision_rate=1.0 / unit.picosecond,
                                                                                      n_steps=n_steps_per_move_application,
                                                                                      reassign_velocities=False,
                                                                                      n_restart_attempts=20),
--- a/perses/app/setup_relative_calculation.py
+++ b/perses/app/setup_relative_calculation.py
@@ -521,9 +521,7 @@ def run_setup(setup_options):
                                                                                     collision_rate=5.0 / unit.picosecond,
                                                                                     n_steps=n_steps_per_move_application,
                                                                                     reassign_velocities=False,
-                                                                                    n_restart_attempts=20,
-                                                                                    splitting="V R O R V",
-                                                                                    constraint_tolerance=1e-06),
+                                                                                    n_restart_attempts=20),
                                                hybrid_factory=htf[phase], online_analysis_interval=setup_options['offline-freq'],
                                                online_analysis_minimum_iterations=10,flatness_criteria=setup_options['flatness-criteria'],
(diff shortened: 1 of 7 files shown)


perses/app/setup_relative_calculation.py
```diff
@@ -517,26 +519,22 @@ def run_setup(setup_options):
...
             if setup_options['fe_type'] == 'sams':
-                hss[phase] = HybridSAMSSampler(mcmc_moves=mcmc.LangevinSplittingDynamicsMove(timestep=timestep,
-                                                                                             collision_rate=5.0 / unit.picosecond,
-                                                                                             n_steps=n_steps_per_move_application,
-                                                                                             reassign_velocities=False,
-                                                                                             n_restart_attempts=20,
-                                                                                             splitting="V R O R V",
-                                                                                             constraint_tolerance=1e-06),
+                hss[phase] = HybridSAMSSampler(mcmc_moves=mcmc.LangevinDynamicsMove(timestep=timestep,
+                                                                                    collision_rate=5.0 / unit.picosecond,
```
jchodera: I think we want to use 1/ps for the collision rate unless we have a compelling reason to use a higher rate.

perses/app/setup_relative_calculation.py
```diff
@@ -517,26 +519,22 @@ def run_setup(setup_options):
...
-                                                                                             collision_rate=5.0 / unit.picosecond,
-                                                                                             n_steps=n_steps_per_move_application,
-                                                                                             reassign_velocities=False,
-                                                                                             n_restart_attempts=20,
-                                                                                             splitting="V R O R V",
-                                                                                             constraint_tolerance=1e-06),
-                                                                                             hybrid_factory=htf[phase],online_analysis_interval=setup_options['offline-freq'])
+                hss[phase] = HybridRepexSampler(mcmc_moves=mcmc.LangevinDynamicsMove(timestep=timestep,
+                                                                                     collision_rate=5.0 / unit.picosecond,
```
jchodera: I think we want to use 1/ps for the collision rate unless we have a compelling reason to use a higher rate.

perses/rjmc/geometry.py
```diff
@@ -644,15 +648,15 @@ def _logp_propose(self, top_proposal, old_positions, beta, new_positions=None, d
...
+        final_context_components = [(force, energy*beta) for force, energy in compute_potential_components(final_context, platform=openmm.Platform.getPlatformByName("Reference"))]
+        atoms_with_positions_reduced_potential_components = [(force, energy*beta) for force, energy in compute_potential_components(atoms_with_positions_context, platform=openmm.Platform.getPlatformByName("Reference"))]
         _logger.debug(f"reduced potential components before atom placement:")
         for item in atoms_with_positions_reduced_potential_components:
             _logger.debug(f"\t\t{item[0]}: {item[1]}")
         _logger.info(f"total reduced potential before atom placement: {atoms_with_positions_reduced_potential}")
 
         _logger.debug(f"potential components added from growth system:")
-        added_energy_components = [(force, energy*beta) for force, energy in compute_potential_components(context)]
+        added_energy_components = [(force, energy*beta) for force, energy in compute_potential_components(context, platform=openmm.Platform.getPlatformByName("Reference"))]
```
jchodera: Do we want to switch to the fastest double-precision platform here?

hannahbrucemacdonald: So switching to 'CUDA' double-precision for this test, the test still fails which seems to be due to a difference in the `NonbondedForce` when calculated on CUDA or Reference. @jchodera do you know why this might be?

For a _real_ system:
CUDA, double: NonbondedForce: -156986.35768798555
Reference: NonbondedForce: -156986.41156104617


Our energy threshold is small 1E-4 so this is enough to tip it into being "unvalid"

Issue #657, opened 2020-03-24
Change endstate validation test to use GPU in double precision mode


jchodera:
We should update the endstate validation tests to use the GPU platform in double precision mode.

Let's see if we can fall back gracefully to Reference if the GPU is not available.

jchodera:
What we really want is the fastest double precision platform.

hannahbrucemacdonald:
This is technically fixed, as we now run the test on `CUDA` with double, if it's available, otherwise fall back to the slower but still correct `Reference`. Fixed in #659

I have opened an openmm issue https://github.com/openmm/openmm/issues/2635
to see if there's a nicer way we could do this (i.e. potentially capturing OpenCL platform if double is possible.)

I am going to close this, and open another issue #663

PR #659, opened 2020-04-02
[READY! 💯] cuda double and energy ratio for check

hannahbrucemacdonald:
… rather than abs. diffenergy validation done on CUDA-double, check changed to energy ratio,…

### DIFF
--- a/perses/rjmc/geometry.py
+++ b/perses/rjmc/geometry.py
@@ -452,7 +452,6 @@ class FFAllAngleGeometryEngine(GeometryEngine):
         if self._storage:
             self._storage.write_object("{}_proposal_order".format(direction), proposal_order_tool, iteration=self.nproposed)
         
-        from openmmtool.utils import get_fastest_platform
         platform_name = 'CUDA'
 
         # Create an OpenMM context
--- a/perses/rjmc/geometry.py
+++ b/perses/rjmc/geometry.py
@@ -455,13 +455,14 @@ class FFAllAngleGeometryEngine(GeometryEngine):
         if self.use_sterics:
             platform_name = 'CPU' # faster when sterics are in use
         else:
-            platform_name = 'Reference' # faster when only valence terms are in use
+            platform_name = 'CUDA' # faster when only valence terms are in use
 
 
         # Create an OpenMM context
         from simtk import openmm
+        from perses.dispersed.utils import configure_platform
         _logger.info("creating platform, integrators, and contexts; setting growth parameter")
-        platform = openmm.Platform.getPlatformByName(platform_name)
+        platform = configure_platform(platform_name, fallback_platform_name='Reference', precision='double')
         integrator = openmm.VerletIntegrator(1*unit.femtoseconds)
         atoms_with_positions_system_integrator = openmm.VerletIntegrator(1*unit.femtoseconds)
         final_system_integrator = openmm.VerletIntegrator(1*unit.femtoseconds)
@@ -489,7 +490,7 @@ class FFAllAngleGeometryEngine(GeometryEngine):
         #Print the energy of the system before unique_new/old atoms are placed...
         state = atoms_with_positions_context.getState(getEnergy=True)
         atoms_with_positions_reduced_potential = beta*state.getPotentialEnergy()
-        atoms_with_positions_reduced_potential_components = [(force, energy) for force, energy in compute_potential_components(atoms_with_positions_context, platform=openmm.Platform.getPlatformByName("Reference"))]
+        atoms_with_positions_reduced_potential_components = [(force, energy) for force, energy in compute_potential_components(atoms_with_positions_context)]
         _logger.debug(f'atoms_with_positions_reduced_potential_components:')
         for f, e in atoms_with_positions_reduced_potential_components:
             _logger.debug(f'\t{f} : {e}')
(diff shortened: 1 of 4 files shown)


hannahbrucemacdonald:
closes #657

perses/rjmc/geometry.py
```diff
@@ -455,13 +455,14 @@ def _logp_propose(self, top_proposal, old_positions, beta, new_positions=None, d
         if self.use_sterics:
             platform_name = 'CPU' # faster when sterics are in use
```
dominicrufa, line 456: we don't use this in the _canonical_ system setup regime, but perhaps platform_name if self.use_sterics should also be set to 'CUDA'

perses/rjmc/geometry.py
```diff
@@ -719,13 +720,15 @@ def _corrected_reduced_potential(self, growth_system_generator, positions, platf
             #then the first one is the normal growth torsion force object and the second is the added torsion force object used to handle chirality and ring-closing constraints
             growth_system.removeForce(max(custom_torsion_forces))
 
-        mod_context = openmm.Context(growth_system, _integrator, openmm.Platform.getPlatformByName(platform_name))
+        from perses.dispersed.utils import configure_platform
+        platform = configure_platform(platform_name, fallback_platform_name='Reference', precision='double')
+        mod_context = openmm.Context(growth_system, _integrator, platform)
```
dominicrufa, line 725: maybe we should be using a context cache in `geometry.py`?  when i did my original pass through to fix rjmc, i wasn't familiar with `openmmtools.cache`, so i omitted, but maybe it's time to fix for good?  just for the sake of not having to specify a platform every time a context is created for bookkeeping purposes

perses/rjmc/geometry.py
```diff
@@ -648,15 +649,15 @@ def _logp_propose(self, top_proposal, old_positions, beta, new_positions=None, d
 
         state = final_context.getState(getEnergy=True)
         final_context_reduced_potential = beta*state.getPotentialEnergy()
-        final_context_components = [(force, energy*beta) for force, energy in compute_potential_components(final_context, platform=openmm.Platform.getPlatformByName("Reference"))]
-        atoms_with_positions_reduced_potential_components = [(force, energy*beta) for force, energy in compute_potential_components(atoms_with_positions_context, platform=openmm.Platform.getPlatformByName("Reference"))]
```
dominicrufa, line 652: the  default platform for `compute_energy_components` is `utils.get_fastest_platform()` : https://github.com/choderalab/perses/blob/6e5c1932e37e875080528776e554b4fb42efcc85/perses/tests/utils.py#L41
which can return a 'CPU`

perses/rjmc/geometry.py
dominicrufa: may have to create a `DEFAULT_PLATFORM` variable at the top of this file?

perses/tests/utils.py
```diff
@@ -777,14 +787,16 @@ def validate_endstate_energies(topology_proposal, htf, added_energy, subtracted_
...
+        _logger.debug(f'added forces:{sum([energy for name, energy in energy_comps])}')
+        _logger.debug(f'rp: {rp}')
         del context, integrator
 
-    #print(f"added_energy: {added_energy}; subtracted_energy: {subtracted_energy}")
     nonalch_zero_rp, alch_zero_rp, alch_one_rp, nonalch_one_rp = rp_list[0], rp_list[1], rp_list[2], rp_list[3]
-    assert abs(nonalch_zero_rp - alch_zero_rp + added_energy) < ENERGY_THRESHOLD, f"The zero state alchemical and nonalchemical energy absolute difference {abs(nonalch_zero_rp - alch_zero_rp + added_energy)} is greater than the threshold of {ENERGY_THRESHOLD}."
-    assert abs(nonalch_one_rp - alch_one_rp + subtracted_energy) < ENERGY_THRESHOLD, f"The one state alchemical and nonalchemical energy absolute difference {abs(nonalch_one_rp - alch_one_rp + subtracted_energy)} is greater than the threshold of {ENERGY_THRESHOLD}."
+
+    ratio = (nonalch_zero_rp - alch_zero_rp + added_energy) / (0.5*(nonalch_zero_rp + alch_zero_rp + added_energy))
```
dominicrufa, line 796: i feel dumb, but can you explain the motivation for using `0.5`?

perses/rjmc/geometry.py
```diff
@@ -455,13 +455,14 @@ def _logp_propose(self, top_proposal, old_positions, beta, new_positions=None, d
         if self.use_sterics:
             platform_name = 'CPU' # faster when sterics are in use
         else:
-            platform_name = 'Reference' # faster when only valence terms are in use
+            platform_name = 'CUDA' # faster when only valence terms are in use
```
jchodera, line 458: Can we avoid specifying CUDA? This means this won't work on Travis tests or on local laptops if CUDA is not available.

Is there a way to just discover the fastest platform instead?

perses/rjmc/geometry.py
jchodera: Should we modify `get_fastest_platform` to specify any requirements (e.g. like double precision is required) so we can return the fastest platform that satisfies those requirements?

perses/rjmc/geometry.py
dominicrufa: i guess just call `utils.get_fastest_platform` and then modify the platform if it gets a CPU with single precision.

perses/tests/utils.py
jchodera: I think it's to make the relative error computation symmetric?
Shouldn't we also take the absolute value to compute a relative error?

perses/rjmc/geometry.py
dominicrufa: ^ that makes sense

perses/rjmc/geometry.py
jchodera: The problem is that returning `OpenCL` means that we don't know if it supports double precision (e.g. on `osx` or Travis). I think we actually have to instantiate a `Context` with a simple `System` (without specifying a platform) to see what the fastest platform that supports those kernels is, and then figure out what platform was assigned.

hannahbrucemacdonald: This is fed into `configure_platform(fallback='Reference',precision='double')`

so if cuda is not there like on travis, it falls back to the reference platform incase the fastest platform doesn't support double

perses/tests/utils.py
hannahbrucemacdonald: It does use abs on line 797 and 799 but I can move it up if that's more legible. Also the 0.5 isn't necessary as we can just change the threshold by a factor of two it doesn't really matter

hannahbrucemacdonald:
Ok I've changed `openmmtools` here: https://github.com/choderalab/openmmtools/pull/463

will make the corresponding changes here now

jchodera:
> i think the openmmtools change hasn't been pushed yet; hence the GH fails

@dominicrufa : What change? Does a PR need to be opened, or has it already been merged into `master` and a new release needs to be cut?

hannahbrucemacdonald:
> Ok I've changed openmmtools here: choderalab/openmmtools#463

I put in the commit message `tests will fail until openmmtools is merged and released`


Let me know if there is more functionality that would be useful to add to `get_fastest_platform`

PR #660, opened 2020-04-06
define n_replicas in getSetupOptions

ajsilveira:


perses/app/setup_relative_calculation.py
```diff
@@ -468,6 +468,8 @@ def run_setup(setup_options):
         _logger.info(f"\tno nonequilibrium detected.")
         n_states = setup_options['n_states']
         _logger.info(f"\tn_states: {n_states}")
+        n_replicas = setup_options['n_states']
```
hannahbrucemacdonald, line 471: Should this be `n_replicas = setup_options['n_replicas']` if you want to have different numbers of states and replicas? Or are you trying to change it so that the default behaviour is n_replicas = n_states?

hannahbrucemacdonald:
I think the tests are failing because of a missing openeye license file, which might be because this is a forked PR?

perses/app/setup_relative_calculation.py
ajsilveira: my bad! my intention was exactly `n_replicas = setup_options['n_replicas']`. I'll fix it now. Thanks!

hannahbrucemacdonald:
Thanks for doing this! I am going to merge now because I suspect that the tests are failing because it's a fork, and I doubt this is code-breaking

PR #662, opened 2020-04-06
Checks the maximum number of atoms mapped possible after hydrogen exception step

### LINKED ISSUE #651: Maps not symmetrical WRT to ligand 0 and 1
This is related to issue #649 

Basically the map should be the same whichever direction you go in, but we seem to have an issue for the same two ligands that I used as an example with the previous issue.

for ligand 10 -> 24 we get:
image

which is correct considering the mapping criteria used
atom_expression:
    - IntType

but doing the opposite 24 -> 10 we get:
image

Where it hasn't maximised the number of possible matches found for the given atom_expression. I realise this map is fine to run, but it isn't the one you'd expect from the atom_expression, and it should be symmetrical.

Maybe it's something to do with the openeye MCS function? But we could get around it by doing mappings of i -> j and j -> i and picking the best map from both sets for the simulation

hannahbrucemacdonald:
Closes #651 by checking atom count after hydrogen exceptions

### DIFF
--- a/perses/rjmc/topology_proposal.py
+++ b/perses/rjmc/topology_proposal.py
@@ -372,6 +372,7 @@ class AtomMapper(object):
         mcs.SetMCSFunc(oechem.OEMCSMaxBondsCompleteCycles())
         unique = False
         matches = [m for m in mcs.Match(oegraphmol_proposed, unique)]
+        _logger.info([m.NumAtoms() for m in matches])
 
         if not allow_ring_breaking:
             # Filter the matches to remove any that allow ring breaking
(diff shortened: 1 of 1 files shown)


perses/rjmc/topology_proposal.py
```diff
@@ -397,11 +398,22 @@ def _get_mol_atom_map(current_oemol,
...
+        count_after_hydrogen_mapping = []
+        for match in top_matches:
+            map_dict = AtomMapper.hydrogen_mapping_exceptions(current_oemol, proposed_oemol, match, matching_criterion) 
+            count_after_hydrogen_mapping.append(len(map_dict))
+            all_new_to_old_atom_maps.append(map_dict)
+
+        max_num_atoms = max(count_after_hydrogen_mapping)
+        _logger.info(f'Maximum atom matched after hydrogen exceptions: {max_num_atoms}')
+        new_to_old_atom_maps = [map for count, map in zip(count_after_hydrogen_mapping,all_new_to_old_atom_maps) if count == max_num_atoms]
+
```
dominicrufa, line 416: looks good; this is certainly more robust than what i had written. makes more sense to retrieve the `new_to_old_atom_maps` _after_ conducting hydrogen mapping exceptions.  would it also make sense to check this _after_ the hydrogen constraint mapping filter is applied? to ensure that the number of atoms mapped is maximized?

hannahbrucemacdonald:
Sorry I'm not sure I follow what you're saying. Is it that `hydrogen_mapping_exceptions` is expensive? 

If so I could leave the first check of `[m for m in top_matches if m.NumAtoms() == max_num_atoms]` in, so that it's being called fewer times? Like a cheap filter then the expensive filter? Should be the same speed that it was previously?

dominicrufa:
> Sorry I'm not sure I follow what you're saying. Is it that `hydrogen_mapping_exceptions` is expensive?
> 
> If so I could leave the first check of `[m for m in top_matches if m.NumAtoms() == max_num_atoms]` in, so that it's being called fewer times? Like a cheap filter then the expensive filter? Should be the same speed that it was previously?

`hydrogen_mapping_exceptions` is inexpensive.  we will omit the `repair_constraints` as this can be cumbersome

Issue #661, opened 2020-04-06
Reaction Field not working


hannahbrucemacdonald:
I added the `nonbonded_method` keyword to allow for reaction field to be run, but this doesn't seem to work, as the simulation still ends up running with PME. Not sure what's going wrong, but maybe something is getting overwritten during the hybrid system generation?

I will take some time to debug in the next few days

hannahbrucemacdonald:
@jchodera 


I'm not sure what the behaviour of this _should_ be

https://github.com/openmm/openmmforcefields/blob/11a9ed2edf0745e77d3e362f27a6c1d5b77abfc5/openmmforcefields/generators/system_generators.py#L80


but we were passing in forcefield_kwargs = {'nonbondedmethod' : app.CutoffPeriodic}

but then it's overwritten by the default for `periodic_forcefield_kwargs` which is PME.

Is it that we are doing it wrong? Or should the default behaviour of SystemGenerator be to first check if nonbondedmethod is defined in one of the other dictionaries?

jchodera:
Good catch!

In order to figure out how to apply parameters to both periodic and non-periodic systems, I wanted to make sure that these were specified in `nonperiodic_forcefield_kwargs` and `periodic_forcefield_kwargs`. 

How about this: We could throw an exception if someone specifies a `nonbondedMethod` in `forcefield_kwargs` instead of `periodic_forcefield_args` or `nonperiodic_forcefield_args`? We can make it informative to instruct the user that this should be put into one of those two.

The fix for perses should be easy then---just change
```python
 forcefield_kwargs = {'removeCMMotion': False, 'ewaldErrorTolerance': self._pme_tol, 'nonbondedMethod': self._nonbonded_method,'constraints' : app.HBonds, 'hydrogenMass' : self._hmass} 
```
to
```python
forcefield_kwargs = {'removeCMMotion': False, 'ewaldErrorTolerance': self._pme_tol, 'constraints' : app.HBonds, 'hydrogenMass' : self._hmass} 
periodic_forcefield_kwargs = {'nonbondedMethod': self._nonbonded_method} 
```
and feed both to `SystemGenerator`.

hannahbrucemacdonald:
Ok sounds good, that's the change I've made in perses. Should I open an openmmforcefields issue? Would you like me to make the fix, or do you want to?

jchodera:
I've already opened an issue here and am just about to push a fix:
https://github.com/openmm/openmmforcefields/issues/108

PR #664, opened 2020-04-07
Replica improvements

hannahbrucemacdonald:


hannahbrucemacdonald:
Closes #661

perses/annihilation/lambda_protocol.py
```diff
@@ -102,6 +102,12 @@ def __init__(self, functions='default'):
                                   lambda x: x,
                                   'lambda_torsions':
                                   lambda x: x}
+            elif self.type == 'ele-scaled':
```
jchodera, line 105: Maybe add a comment here that this is "linear scaling of electrostatic energy for insertion and deletion"?

perses/app/relative_setup.py
```diff
@@ -263,9 +262,9 @@ def __init__(self, ligand_input, old_ligand_index, new_ligand_index, forcefield_
 
         # Create SystemGenerator
         from openmmforcefields.generators import SystemGenerator
-        forcefield_kwargs = {'removeCMMotion': False, 'ewaldErrorTolerance': self._pme_tol, 'nonbondedMethod': self._nonbonded_method,'constraints' : app.HBonds, 'hydrogenMass' : self._hmass}
```
dominicrufa, line 266: i missed something; why is te nonbonded method not being specified anymore?

perses/annihilation/lambda_protocol.py
hannahbrucemacdonald: Done in docstrings

perses/app/relative_setup.py
hannahbrucemacdonald: Good spot, this is to do with openmm-forcefields, where whatever nonbonded method we chose was being overwritten by the default for `periodic_forcefield_kwargs` (which is PME anyway, but the reason that RF wasn't working before)

openmm-forcefield fix: https://github.com/openmm/openmmforcefields/pull/109
perses issue: https://github.com/choderalab/perses/issues/661

hannahbrucemacdonald:
closes #661

perses/app/relative_setup.py
dominicrufa: so if i wanted `NoCutoff` for vacuum, what would i do?

PR #665, opened 2020-04-09
Neq customintegrator

dominicrufa:
adding a nonequilibrium `CustomIntegrator` that will conduct (attempted) i.i.d sampling at endstates and a forward/backward annealing


dominicrufa:
image

dominicrufa:
perhaps this should actually go into `openmmtools` since it is just a subclass of : https://github.com/choderalab/openmmtools/blob/c2b61c410b255c4e08927acf8cfcb1cf46f64b70/openmmtools/integrators.py#L1649

maxentile:
Nice work! Taking a look shortly.

Can we copy the test from cells 11 and 12 of notebook into `tests/`?

maxentile:
I had a clarification question about the intended logic for end-state sampling:
1. Do we have two running trajectories for the end-state sampling, that are not dependent on what happens during noneq switching?
2. [currently implemented] Do we want to restart lambda=1 end-state MD from whatever geometry came out of the last lambda=0->1 switching trajectory? (and vice-versa)

I worry that with (2), the resulting estimates will be biased, since each switching trajectory will be initialized with an off-equilibrium configuration that depends on what happened in the previous, opposite-direction switching trajectory. This violates a few assumptions, and doesn't become consistent in limit of increasing number of trajectories (only becoming consistent in limit of long switching or long equilibration).

Implementation of (1) entirely inside a CustomIntegrator should be doable, although clunky (e.g. requiring to store the most recent lambda=0 and lambda=1 equilibrium configurations in per-dof variables).

@jchodera , could you comment on this?

jchodera:
@maxentile : The design I discussed with @dominicrufa is that the `CustomIntegrator` should execute the following cycle: `equil A > neq switch to B > equil B > neq switch to A > ...` repeats, accumulating protocol work the whole time.

This integrator should take two different parameters:
* `nequil` is the number of steps that the integrator should reside in A or B for the `equil` stages
* `nswitch` is the number of steps the integrator should use to switch from A to B or B to A

The goal here is that a single long simulation run on Folding@home will continue to switch back and forth between A and B, pausing at A or B long enough to generate a new iid equilibrium sample. We will have full control over `nequil` and `nswitch` so we can (1) explore these parameters, and (2) choose `nequil` sufficiently long to ensure all switches _start_ from equilibrium.

For use on Folding@home, we can select `nequil` and `nswitch` to be multiples of some interval for which the cumulative protocol work is retrieved and stored (e.g. 100 ps = 25000 steps at 4 fs) so that we can later extract the work for each nonequilibrium realization.

jchodera:
And yes! This might be better in `openmmtools` since it has utility on its own!

dominicrufa:
@jchodera , the `CustomIntegrator` is currently implemented as we had discussed the other day.  it allows for both of the parameters that you mentioned.

maxentile:
Thanks!

>@maxentile : The design I discussed with @dominicrufa is that the CustomIntegrator should execute the following cycle: equil A > neq switch to B > equil B > neq switch to A > ... repeats, accumulating protocol work the whole time.

On the same page about that, I was just seeking to clarify whether we should use different logic to initialize `equil A`, `equil B` within that cycle. (Should `equil A` be initialized with the previous output of `equil A` (option 1), or should `equil A` be initialized with the previous output of `neq switch to A` (option 2, implemented). Implementing option 1 in a CustomIntegrator would require storing per-dof variables `x`, `x_A`, `x_B`, where option 2 required only storing per-dof variable `x`.)

It sounds like the decision to use option 2 is intentional, and the bias introduced can be made small enough by picking large enough `nequil` and `nswitch`.

jchodera:
> It sounds like the decision to use option 2 is intentional, and the bias introduced can be made small enough by picking large enough nequil and nswitch.

Yes, exactly.

@dominicrufa : Do you want to move this and its test to `openmmtools` and we can cut a new release?

jchodera:
> I was just seeking to clarify whether we should use different logic to initialize equil A, equil B within that cycle. (Should equil A be initialized with the previous output of equil A (option 1)

This is a clever idea I hadn't thought of, and would let us program a FAH run to just generate an equilibrium trajectory in A and shoot off nonequilibrium trajectories from A -> B periodically.

@maxentile : Do you want to take a stab at a variant integrator that does this once @dominicrufa merges his into `openmmtools`? We might want to experiment with both on FAH.

maxentile:
>This is a clever idea I hadn't thought of, and would let us program a FAH run to just generate an equilibrium trajectory in A and shoot off nonequilibrium trajectories from A -> B periodically.
>@maxentile : Do you want to take a stab at a variant integrator that does this once @dominicrufa merges his into openmmtools? We might want to experiment with both on FAH.

I think this was a design option that @dominicrufa had considered as well!

And yes, happy to after we get the current implementation merged into openmmtools.

perses/dispersed/loopy_annealing_integrator.py
```diff
@@ -0,0 +1,153 @@
...
+        splitting = 'V R O R V'
+        super().__init__(alchemical_functions, splitting, nsteps_neq = nsteps_neq, **kwargs)
+    
+    def _add_global_variables(self):
+        """
+        modify the super (_add_global_variables) to add a collector for forward and reverse annealing, as well as to specify the number of equilibration
+        steps at endstates.
+        """
+        super()._add_global_variables()
+        self.addGlobalVariable('forward_work', 0)
```
jchodera, line 55: @dominicrufa : We don't want to separately accumulate forward and backward work, since it will get very complicated to deconvolute. We just want to accumulate the total `protocol_work` and then extract differences in the total work at different times to get the work for each switch.

perses/dispersed/loopy_annealing_integrator.py
```diff
@@ -0,0 +1,153 @@
...
+        self.addGlobalVariable('backward_work', 0)
+        self.addGlobalVariable('n_steps_eq', self._n_steps_eq)
```
jchodera, line 57: Should we be consistent between the class arguments (`nsteps_eq`, `nsteps_neq`) and the global parameter names (`n_steps_eq`, `n_steps_neq`)?

jchodera: I guess this may not matter much since we call the nonequilibrium switching time something complicated like `n_lambda_steps`

perses/dispersed/loopy_annealing_integrator.py
jchodera: Or do we just _also_ separate the work into forward and backward, in _addition_ to accumulating `protocol_work`? If so, I don't see any problem with doing both!

dominicrufa: I'm confused about what you mean here; can you explain further?

dominicrufa: you don't want to accumulate forward and backward works for each step and then use a `BAR` estimator?

perses/dispersed/loopy_annealing_integrator.py
dominicrufa: this is how `AlchemicalNonequilibrumLangevinIntegrator` treats `nsteps_neq`; i figured i'd preserve the internal consistency defined in the super class

perses/dispersed/loopy_annealing_integrator.py
dominicrufa: i _think_ i understand what you are talking about, but let me be clear and you can raise any concerns.  at the start of an integration step, the `forward_work`, `backward_work` and `protocol_work` are all zeroed.  the `protocol_work` is accumulated in the _forward_ annealing, passed to `forward_work`, and zeroed.  `protocol_work` is re-accumulated (from zero) in the backward direction, and then passed to `backward_work`

dominicrufa: strictly speaking, there is no reason to create a `backward_work` variable, since it is contained in the `protocol_work` at the end of the entire integration step, but since i created a `forward_work` variable to log the accumulated work in the forward pass, it made sense to me to also create a `backward_work` variable

PR #667, opened 2020-04-22
[WIP] F@H generator script

dominicrufa:
create a script to automate the project pipeline for folding at home


dominicrufa:
`core.xml` is not yet written

perses/app/fah_generator.py
```diff
@@ -0,0 +1,203 @@
...
+conda config --add channels omnia --add channels conda-forge
+conda create -n perses python3.7 perses tqdm dicttoxml
+pip uninstall --yes openmmtools
+pip install git+https://github.com/choderalab/openmmtools.git
+
+argv[1]: setup.yaml (argument for perses.app.setup_relative_calculation.getSetupOptions)
+argv[2]: neq_setup.yaml (contains keywords for openmmtools.integrators.PeriodicNonequilibriumIntegrator arguments)
+argv[3]: run_number (project run number; defined by f"setup_options['trajectory_directory']_phase/RUN_{run_number}")
+"""
+import yaml
```
jchodera, line 18: As a general practice, let's always keep `import`s local to right before they're used.
This is generally useful as it keeps the code tidier and more modular, and prevents long `import` times.

perses/app/fah_generator.py
```diff
@@ -0,0 +1,203 @@
...
+from perses.dispersed.smc import SequentialMonteCarlo
+from simtk.openmm import XmlSerializer
+from perses.utils import data
+from openmmtools.integrators import PeriodicNonequilibriumIntegrator
+from copy import deepcopy
+import tqdm
+from openmmtools.integrators import LangevinIntegrator
+
+import datetime
+class TimeFilter(logging.Filter):
```
jchodera, line 44: As a practice, let's make sure to add docstrings to all methods and classes, even for special-purpose code.

perses/app/fah_generator.py
```diff
@@ -0,0 +1,203 @@
...
+        try:
+          last = self.last
+        except AttributeError:
+          last = record.relativeCreated
+        delta = datetime.datetime.fromtimestamp(record.relativeCreated/1000.0) - datetime.datetime.fromtimestamp(last/1000.0)
+        record.relative = '{0:.2f}'.format(delta.seconds + delta.microseconds/1000000.0)
+        self.last = record.relativeCreated
+        return True
+
+fmt = logging.Formatter(fmt="%(asctime)s:(%(relative)ss):%(name)s:%(message)s")
```
jchodera, line 55: Add a comment for why we need this formatter code?

perses/app/fah_generator.py
```diff
@@ -0,0 +1,203 @@
...
+    level=logging.INFO,
+    datefmt='%Y-%m-%d %H:%M:%S')
+_logger = logging.getLogger()
+_logger.setLevel(logging.INFO)
+[hndl.addFilter(TimeFilter()) for hndl in _logger.handlers]
+[hndl.setFormatter(fmt) for hndl in _logger.handlers]
+
+setup_yaml = sys.argv[1]
+neq_yaml = sys.argv[2]
+ENERGY_THRESHOLD=1e-4
```
jchodera, line 68: Add a comment for what ENERGY_THRESHOLD is and what units it is in?

perses/app/fah_generator.py
```diff
@@ -0,0 +1,203 @@
...
+logging.basicConfig(
+    format='%(asctime)s %(levelname)-8s %(message)s',
+    level=logging.INFO,
+    datefmt='%Y-%m-%d %H:%M:%S')
+_logger = logging.getLogger()
+_logger.setLevel(logging.INFO)
+[hndl.addFilter(TimeFilter()) for hndl in _logger.handlers]
+[hndl.setFormatter(fmt) for hndl in _logger.handlers]
+
+setup_yaml = sys.argv[1]
```
jchodera, line 66: As a general practice, let's be sure to at least use [`argparse`](https://docs.python.org/3/library/argparse.html), and never use `sys.argv`. 

This should also always occur inside
```python
if __name__ == "__main__":
    ...
```
so that we can import methods from this module if we need to reuse them.

perses/app/fah_generator.py
```diff
@@ -0,0 +1,203 @@
...
+                             'lambda_sterics_delete': f"select(step({x} - 0.5), 2.0 * ({x} - 0.5), 0.0)",
+                             'lambda_electrostatics_insert': f"select(step({x} - 0.5), 2.0 * ({x} - 0.5), 0.0)",
+                             'lambda_electrostatics_delete': f"select(step({x} - 0.5), 1.0, 2.0 * {x})",
+                             'lambda_bonds': x,
+                             'lambda_angles': x,
+                             'lambda_torsions': x}
+
+from openmmtools.constants import kB
+
+def make_neq_integrator(setup_options_dict, neq_options_dict, alchemical_functions = DEFAULT_ALCHEMICAL_FUNCTIONS):
```
jchodera, line 83: Always include docstrings for all methods in the [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) format that fully document what the method does, its parameters, and its return values:
```python
def myfunc(a, b, c=None):
    """
    Perform a thing on the whatsits.

    Parameters
    ----------
    a : int
        The first argument
    b : str
        The second argument. One of ['one', 'two', 'three']
    c : float, optional, default=None
        The third argument

    Returns
    -------
    result : str
        The result
    """
```
We *always* include these, no matter what. Trust me, it's important if you ever want code to be used by anyone, including Future You.

perses/app/fah_generator.py
```diff
@@ -0,0 +1,203 @@
...
+    integrator_kwargs = deepcopy(setup_options_dict)
+    integrator_kwargs.update(neq_options_dict)
+    integrator_kwargs['alchemical_functions'] = alchemical_functions
+    integrator = PeriodicNonequilibriumIntegrator(**integrator_kwargs)
+    return integrator
+
+def relax_structure(temperature, system, positions, nminimize=100, nequil = 4, n_steps_per_iteration=250):
+    integrator = LangevinIntegrator(temperature = temperature)
+    context = openmm.Context(system, integrator)
```
jchodera, line 92: Are we guaranteed that we don't need to explicitly set the context parameters to anything but their default values to have hybrid state A? I was concerned this might be necessary.

perses/app/fah_generator.py
```diff
@@ -0,0 +1,203 @@
...
+    state = context.getState(getEnergy=True, getForces=True, getPositions=True, getVelocities=True, getParameters=True)
+
+    del context, integrator
+​
+    return state
+
+
+def run_neq_fah_setup(setup_yaml, neq_yaml, run_number, **kwargs):
+    """
+    given a run_number, read a perses setup yaml, a neq-specific yaml, and conduct the following:
```
jchodera, line 115: Fully document parameters, return values.

perses/app/fah_generator.py
```diff
@@ -0,0 +1,203 @@
...
+        references = {'start_ligand': setup_dict['old_ligand_index'],
+                      'end_ligand': setup_dict['new_ligand_index'],
+                      'protein_filename': setup_dict['protein_pdb'],
+                      'passed_strucutre_relax': passed}
+        with open(f"{dir}/reference.pkl", 'wb') as f:
+            pickle.dump(references, f)
+
+
+if __name__ == "__main__":
+    setup_yaml, neq_setup_yaml, run_number = sys.argv[1], sys.argv[2], sys.argv[3] #define args
```
jchodera, line 187: Always at least use `argparse`

perses/app/fah_generator.py
```diff
@@ -0,0 +1,203 @@
...
+
+    #open the setup yaml to pull trajectory_directory
+    yaml_file = open(setup_yaml, 'r')
```
jchodera, line 190: Can you provide an example of the setup_yaml YAML file? I'm not sure what it should contain.

perses/app/fah_generator.py
```diff
@@ -0,0 +1,203 @@
...
+    setup_options = yaml.load(yaml_file, Loader=yaml.FullLoader)
+    yaml_file.close()
+    traj_dir = setup_options['trajectory_directory']
+
+    gather all
```
jchodera, line 195: Is this valid Python?

perses/app/fah_generator.py
```diff
@@ -0,0 +1,203 @@
...
+    phases = setup_options['phases']
+    for phase in phases:
+        now_path = os.path.join(os.getcwd(), f"{traj_dir}_{phase}")
+        if not os.path.exists(now_path):
+            os.mkdir(now_path)
+        else:
+            raise Exception(f"{now_path} already exists.  Aborting.")
+    run_neq_fah_setup(setup_yaml, neq_yaml, int(run_number), **kwargs)
```
jchodera, line 203: Where do the `kwargs` come from? Should we pull them from the YAML so that we can also configure the switching time, etc?

perses/app/fah_generator.py
dominicrufa: yes

jchodera:
From CLI, arguments ideally will be:
* RUN index to set up
* YAML file for everything else

YAML file would ideally contain:
* Project numbers for complex and solvent legs (e.g. 13400, 13401)
* PDB, SDF for reference
* SMILES list for targets

Directory structure must be (e.g. for ligand `index`):
* complex: `13400/RUN{index}/{system,state,integrator,core}.xml`
* solvent: `13401/RUN{index}/{system,state,integrator,core}.xml`

perses/app/fah_generator.py
```diff
@@ -0,0 +1,315 @@
...
+
+
+def run_neq_fah_setup(ligand_file,
+                      old_ligand_index,
+                      new_ligand_index,
+                      forcefield_files,
+                      trajectory_directory,
+                      index=0,
+                      box_dimensions=(8.5,8.5,8.5),
+                      timestep=1.0 * unit.femtosecond,
```
jchodera, line 122: The timestep should be 4 fs, not 1 fs.

perses/app/fah_generator.py
```diff
@@ -0,0 +1,255 @@
...
+    setup_options['trajectory_prefix']=None
+
+    #run the run_setup to generate topology proposals and htfs
+    setup_dict = run_setup(setup_options, serialize_systems=False, build_samplers=False)
+    topology_proposals = setup_dict['topology_proposals']
+    htfs = setup_dict['hybrid_topology_factories']
+
+    #create solvent and complex directories
+    for phase in htfs.keys():
+        dir = os.path.join(os.getcwd(), f"{trajectory_directory}_{phase}", f"RUN_{format(old_ligand_index, '03')}_to_{format(new_ligand_index, '03')}")
```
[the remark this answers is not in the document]
hannahbrucemacdonald: This is outdated

perses/app/fah_generator.py
```diff
@@ -0,0 +1,255 @@
...
+            data.serialize(state, f"{dir}/state.xml")
+        except Exception as e:
+            print(e)
+            passed=False
+        else:
+            passed=True
+
+
+        #lastly, make a core.xml
+        #TODO: make core.xml; i don't get the format. help
```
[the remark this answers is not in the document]
hannahbrucemacdonald: thanks

perses/app/fah_generator.py
```diff
@@ -40,18 +40,20 @@
                              'lambda_torsions': x}
 
 
-def make_neq_integrator(nsteps_eq, nsteps_neq, neq_splitting, alchemical_functions = DEFAULT_ALCHEMICAL_FUNCTIONS, **kwargs):
+def make_neq_integrator(nsteps_eq=1000, nsteps_neq=1000, neq_splitting='V R H O R V', timestep=4.0 * unit.femtosecond, alchemical_functions = DEFAULT_ALCHEMICAL_FUNCTIONS, **kwargs):
```
jchodera, line 43: If you're using 4 fs timesteps and we want 1 ns equilibrium and nonequilibrium dwell periods, both `nsteps_eq` and `nsteps_neq` should be `250000`.

jchodera:
@hannahbrucemacdonald : These `System` objects are also still missing a `MonteCarloBarostat`!

dominicrufa:
> @hannahbrucemacdonald : These System objects are also still missing a MonteCarloBarostat!

system generator is created here: https://github.com/choderalab/perses/blob/53bf910b65c17a21fc941850888c81f9f9928b36/perses/app/relative_setup.py#L285

pressure is defaulted as not None, and nonbonded method is defaulted as PME:https://github.com/choderalab/perses/blob/53bf910b65c17a21fc941850888c81f9f9928b36/perses/app/relative_setup.py#L246

pressure and barostat is only deleted from the system generator here (if vacuum phase, but only after complex/solvent are made): https://github.com/choderalab/perses/blob/53bf910b65c17a21fc941850888c81f9f9928b36/perses/app/relative_setup.py#L390

pressure and nonbonded method is set here (outside the class ): https://github.com/choderalab/perses/blob/53bf910b65c17a21fc941850888c81f9f9928b36/perses/app/setup_relative_calculation.py#L385

system for the HTF is pulled here: https://github.com/choderalab/perses/blob/53bf910b65c17a21fc941850888c81f9f9928b36/perses/app/setup_relative_calculation.py#L495

and the HTF copies the old system's barostat here: https://github.com/choderalab/perses/blob/53bf910b65c17a21fc941850888c81f9f9928b36/perses/annihilation/relative.py#L243

there _might_ be a non-default set somewhere in the setup yaml, but i don't know where that is...perhaps it would be best if a default yaml could be put in `app`

perses/app/fah_generator.py
```diff
@@ -0,0 +1,347 @@
...
+    _logger.info(f'Starting to relax')
+    integrator = LangevinIntegrator(temperature = temperature)
+    context = openmm.Context(system, integrator)
+    context.setPeriodicBoxVectors(*system.getDefaultPeriodicBoxVectors())
+    context.setPositions(positions)
+
+    _logger.info(f'Starting to minimise')
+    # Minimize
+    for iteration in tqdm.tqdm(range(nminimize)):
+        openmm.LocalEnergyMinimizer.minimize(context, 0.0, 1)
```
jchodera, line 103: It may be better to just do one minimization that doesn’t set a step limit or force norm threshold. I realize that LBFGS doesn’t do its thing if you don’t give it enough steps---it may not work right if we run one step at a time.

jchodera: I'd change this from 
```python
for iteration in tqdm.tqdm(range(nminimize)):
        openmm.LocalEnergyMinimizer.minimize(context, 0.0, 1)
```
to
```python
openmm.LocalEnergyMinimizer.minimize(context)
```

perses/app/fah_generator.py
```diff
@@ -0,0 +1,348 @@
...
+        state : openmm.State
+            state of simulation (getEnergy=True, getForces=True, getPositions=True, getVelocities=True, getParameters=True)
+    """
+
+    from perses.dispersed.feptasks import minimize
+    from openmmtools.integrators import LangevinIntegrator
+    import tqdm
+    _logger.info(f'Starting to relax')
+    integrator = LangevinIntegrator(temperature = temperature)
+    context = openmm.Context(system, integrator)
```
jchodera, line 94: @hannahbrucemacdonald : You're using the CUDA platform here without setting `Precision` to `mixed` and the `DeeterministicForce` to `true`! This may be responsible for the discrepancies we're seeing.

Suggest you change this to use an explicit `platform_name`:
```
platform = openmm.Platform.getPlatformByName(platform_name)
if platform_name in ['CUDA', 'OpenCL']:
    platform.setPropertyDefaultValue('Precision', 'mixed')
if platform_name in ['CUDA']:
    platform.setPropertyDefaultValue('DeterministicForces', 'true')
context = openmm.Context(system, integrator, platform)
```

perses/app/fah_generator.py
```diff
@@ -315,29 +309,26 @@ def run(yaml_filename=None,index=None):
 
     import os
     # make master directories
-    if not os.path.exists('13406'):
-        os.makedirs('1340r/RUNS/')
-    if not os.path.exists('13407'):
-        os.makedirs('13407/RUNS/')
+    if not os.path.exists('13408'):
```
jchodera, line 312: @hannahbrucemacdonald Shouldn't these RUN indices be specified in the YAML file?

hannahbrucemacdonald:
@dominicrufa

jchodera:
Could we clean up the failures too?
```
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/runner/work/perses/perses/perses/tests/test_coordinate_numba.py", line 15, in <module>
    from perses.tests.test_geometry_engine import _get_internal_from_omm
  File "/home/runner/work/perses/perses/perses/tests/test_geometry_engine.py", line 46, in <module>
    small_molecule_forcefield = small_molecule_forcefield)
  File "/usr/share/miniconda/envs/test/lib/python3.7/site-packages/openmmforcefields/generators/system_generators.py", line 183, in __init__
    or nonperiodic_forcefield_kwargs (if it should be applied to non-periodic systems)""")
ValueError: nonbondedMethod cannot be specified in forcefield_kwargs;
                 must be specified in either periodic_forcefield_kwargs (if it should be applied to periodic systems)
                 or nonperiodic_forcefield_kwargs (if it should be applied to non-periodic systems)
```
Let me know if you want me to take a stab at these!

PR #674, opened 2020-05-21
Remove explicit outdated parmed dependency and add rc channel

jchodera:


jchodera:
The current failures are due to
```
Failure: RuntimeError (ParmEd is too old, please upgrade to 2.0.4 or later) ... ERROR
```
which is appearing because the latest `ambertools=20` on `conda-forge` returns an erroneous `parmed.version`:
https://github.com/conda-forge/ambertools-feedstock/issues/35

jchodera:
Once we fix `openmoltools`, this will hopefully stop failing.

Issue #680, opened 2020-06-01
Docs


hannahbrucemacdonald:
Lots of the links are broken in the docs --- move to makethedocs or some sort of markdown doc thing instead of sphinx like we do now

jchodera:
Fixing the API docs simply requires ensuring that readthedocs can correctly install `perses` and enough dependencies to import and inspect the `perses` modules. I've added @jaimergp as a maintainer to http://perses.readthedocs.io/ to help, but I need readthedocs usernames from @hannahbrucemacdonald and @pgrinaway as well.

We'll likely need to mess around with [the docs environment](https://github.com/choderalab/perses/blob/master/docs/environment.yml) to bring it up to date.

Even better would also be for @dominicrufa @hannahbrucemacdonald to migrate `import`s into the methods they are used in so they are just-in-time imported. Otherwise, docs building requires literally every dependency to be installed in order to import a package and inspect the docstrings, which is painful and slow.

jaimergp:
I am going to try the mocking features to address this in a new PR:
https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_mock_imports

PR #682, opened 2020-06-02
Fix docs

jaimergp:
After #680 

# To do

- [ ] Add CI to make sure docs build without warnings.
- [ ] Ensure mocking works instead of installing all dependencies. Some are too complex to mock, but at least we will get rid of most.
- [ ] Make sure these changes apply to readthedocs as well


jaimergp:
Check [logs for docs](https://github.com/choderalab/perses/pull/682/checks?check_run_id=730851873#step:5:10277) to see all warnings. It's not as broken as before, but there's plenty of pointers to dead / moved code and I am not very familiar with those. It might be also helpful to disable `numba`'s ultra verbose DEBUG logging.

jaimergp:
Also, imports and the logging machinery needs some work. Not in this PR, but just writing it down for future reference.

environment.yml
```diff
@@ -32,3 +32,5 @@ dependencies:
   - tqdm
   - nose
   - nose-timer
+  - openforcefield
+  - sphinx
```
dominicrufa, line 36: this enable documentation?

docs/rjmc.rst
```diff
@@ -15,16 +15,17 @@ Topology proposal engines
...
     PeptideLibraryEngine
-    NullProposalEngine
-    NaphthaleneProposalEngine
-    ButaneProposalEngine
-    PropaneProposalEngine
+    .. JRG: This class does not exist anymore?
+    .. ButaneProposalEngine
+    .. NaphthaleneProposalEngine
+    .. PropaneProposalEngine
+    .. NullProposalEngine
```
dominicrufa, line 28: this allows for documentation of classes? this should actually be for `topology_proposal.py`, no?

docs/rjmc.rst
```diff
@@ -34,7 +35,8 @@ OpenMM System generation utilities
     :nosignatures:
     :toctree: api/generated/
 
-    SystemGenerator
+    .. .. This has moved to openforcefields?
```
dominicrufa, line 38: yes

docs/rjmc.rst
```diff
@@ -46,7 +48,8 @@ Geometry proposal engines
 
     GeometryEngine
     FFAllAngleGeometryEngine
-    OmegaGeometryEngine
+    .. JRG: This class does not exist anymore?
+    .. OmegaGeometryEngine
```
dominicrufa, line 52: this was removed

docs/rjmc.rst
```diff
@@ -56,9 +59,11 @@ Geometry utility classes
...
-    GeometrySystemGeneratorFast
-    PredHBond
-    ProposalOrderTools
+    .. JRG: These classes does not exist anymore?
+    .. BootstrapParticleFilter
+    .. GeometrySystemGeneratorFast
+    .. PredAtomTopologyIndex
+    .. PredHBond
+    .. .. This one is also dead code in some modules!
+    .. ProposalOrderTools
```
dominicrufa, line 69: i mostly rewrote `geometry.py` and never updated documentation. should there be a .rst for every API-containing `.py` script in the code?

jaimergp: That would be one way. The current approach is a semi-automated way to approach this, but it should be rearranged into a different directory tree. I like the `api` stuff to be separate from the user-oriented docs. I can propose a new scheme, but first we need to trim out all the dead leaves :)

If you can add comments to those classes that are gone in the current version and identify if the missing references (check the logs on `CI / Build documentation`) are a bug or just the byproduct of dead code, then I can work on the rest! Thanks!

hannahbrucemacdonald:
@jaimergp @dominicrufa @jchodera could this be merged soon? It is only changes to the docs, and while the changes aren't complete, they are still an improvement and the longer it's left the more conflicts there will be to merge

PR #689, opened 2020-06-10
[WIP] using geometry to fix mapping issues

hannahbrucemacdonald:


hannahbrucemacdonald:
`use_geometry` will use the geometry of ligand B to pick which map to use. The default is True, but if this is set to `False` it will return to the previous code. Ultimately, this should be depreciated and if `use_geometry` is False, then multiple maps are returned.

hannahbrucemacdonald:
So my initial fix where the cdist score for the map was used, was fixing the problem with this Bace example, when the two map options are:

image

image


but this doesn't fix the issue where there are two groups that are both meta, and want to point the other way to eachother. Because the mapping of the hydrogen will always be favourable, we don't get the optimal map out of the MCS search as it's worse. While the O and the C can be unmapped using `oechem.OEExprOpts_AtomicNumber` it still puts both the unique atoms onto the same carbon. 

image

This also wouldn't help if we had a methyl group on one side and an ethyl on the other side. 

Instead, for this case, the geometry needs to be captured _before_ it goes into the MCS search. So `geometry` can be set to `strong`, and then `_assign_distance_ids` is called to find the geometrically close atoms.

Using this gives us the _correct_ map where the two R groups are on carbons 10 and 12 respectively.

image

perses/rjmc/topology_proposal.py
```diff
@@ -401,46 +423,82 @@ def _get_mol_atom_map(current_oemol,
...
-        
+        _logger.info(f'Number of maps is {len(all_new_to_old_atom_maps)}')
+
+        if geometry is not None:
+            # now want to pick the map with smallest distance for B geometry
+            current_coords = np.zeros(shape=(current_oemol.NumAtoms(),3))
+            for i in current_oemol.GetCoords():
+                current_coords[i] = current_oemol.GetCoords()[i]
+            proposed_coords = np.zeros(shape=(proposed_oemol.NumAtoms(),3))
+            for i in proposed_oemol.GetCoords():
```
dominicrufa, line 452: my understanding is that getting the coords is not necessarily deterministic (unless an sdf or mol2 files makes the proposed ligand and has a given geometry)

perses/rjmc/topology_proposal.py
```diff
@@ -401,46 +423,82 @@ def _get_mol_atom_map(current_oemol,
...
+                    if not proposed_H[atom]:  # skip H's - only look at heavy atoms
+                        map_score += all_to_all[M[atom], atom]
+                all_scores.append(map_score/len(M))
+            _logger.debug(f'Mapping scores: {all_scores}')
+            # TODO: return one map from any group of scores
+
+            # returning lowest score
+            best_map_index = np.argmin(all_scores)
+            _logger.debug(f'Returning map index: {best_map_index}')
+            return all_new_to_old_atom_maps[best_map_index]
```
dominicrufa, line 471: so here, we are returning a single dict

perses/rjmc/topology_proposal.py
```diff
@@ -401,46 +423,82 @@ def _get_mol_atom_map(current_oemol,
...
         # check the most mapped is the same once hydrogen exceptions are handled
         all_new_to_old_atom_maps = []
         count_after_hydrogen_mapping = []
-        for match in top_matches:
-            map_dict = AtomMapper.hydrogen_mapping_exceptions(current_oemol, proposed_oemol, match, matching_criterion) 
+        if return_all_maps:
+            list_of_dicts = []
+            for match in top_matches:
+                list_of_dicts.append(AtomMapper.hydrogen_mapping_exceptions(current_oemol, proposed_oemol, match, matching_criterion))
+            return list_of_dicts
```
dominicrufa, line 434: here we are returning a list of dicts

perses/rjmc/topology_proposal.py
```diff
@@ -338,35 +341,55 @@ def _get_mol_atom_map(current_oemol,
...
+            will return a list of viable maps. This is for debugging purposes only
+            as a list being returned in the overall pipeline would break everything
+        geometry : str, default 'weak'
+            one of ['strong','weak' or None]
+            if strong -- atoms that are close in the geometries of ligand A and B are forced to map together
+            if weak -- if there are multiple mappings, the mapping that is closer to the geometry of ligand B is chosen
+            if None, all geometry information is ignored
         Returns
         -------
         matches : list of match
```
dominicrufa, line 357: the object that is returned is now inconsistent. perhaps if geometry is used and returns the best match, it should simply return a list of length 1?

perses/rjmc/topology_proposal.py
```diff
@@ -401,46 +423,82 @@ def _get_mol_atom_map(current_oemol,
...
             all_new_to_old_atom_maps.append(map_dict)
+            _logger.debug(f'index {i} = {map_dict}')
 
         max_num_atoms = max(count_after_hydrogen_mapping)
         _logger.info(f'Maximum atom matched after hydrogen exceptions: {max_num_atoms}')
-        new_to_old_atom_maps = [map for count, map in zip(count_after_hydrogen_mapping,all_new_to_old_atom_maps) if count == max_num_atoms]
-        
+        _logger.info(f'Number of maps is {len(all_new_to_old_atom_maps)}')
+
+        if geometry is not None:
```
dominicrufa, line 446: a bit nitpicky, but could this be a staticmethod that lives in the AtomMapper class?

perses/rjmc/topology_proposal.py
```diff
@@ -401,46 +423,82 @@ def _get_mol_atom_map(current_oemol,
...
+                for map in new_to_old_atom_maps:
+                    hit_number = 0
+                    map_tuples = list(map.items())
+                    atom_map = {atom_new: atom_old for atom_new, atom_old in zip(list(proposed_oemol.GetAtoms()), list(current_oemol.GetAtoms())) if (atom_new.GetIdx(), atom_old.GetIdx()) in map_tuples}
+                    for key, value in atom_map.items():
+                        if key.GetName() == value.GetName():
+                            hit_number += 1
+                    index_overlap_numbers.append(hit_number)
+            else:
+                raise Exception(f"the ranking criteria {matching_criterion} is not supported.")
```
dominicrufa, line 495: so the geometry filter is happening before before mapping of atomic indices filter...would this run faster if we filtered by geometry _after_ everything else?

perses/rjmc/topology_proposal.py
```diff
@@ -534,6 +592,41 @@ def _assign_ring_ids(molecule, max_ring_size=10):
             atom.SetIntType(ring_as_base_ten)
         return id_molecule
 
+    @staticmethod
+    def _assign_distance_ids(old_mol, new_mol, distance=0.3):
```
dominicrufa, line 596: is 0.3A just the one that happened to work?

perses/rjmc/topology_proposal.py
```diff
@@ -534,6 +592,41 @@ def _assign_ring_ids(molecule, max_ring_size=10):
...
+        of both molecules are known, and they are aligned to the same frame of reference.
+
+        this function is invariant to which is passed in as {old|new}_mol
+        Parameters
+        ----------
+        old_mol : oechem.OEMol
+            first molecule to compare
+        new_mol : oechem.OEMol
+            second molecule to compare
+        distance : float, default = 0.2
```
dominicrufa, line 608: inconsistent with default arg

perses/rjmc/topology_proposal.py
hannahbrucemacdonald: The coords come from the input file, so should be the same every time --- I will make sure that this isn't called if we are using a smiles as input though

perses/rjmc/topology_proposal.py
hannahbrucemacdonald: Possibly --- but that is still going to potentially break other things further down the line. I've added _logger.warning that returning all matches shouldn't be done in the normal pipeline

perses/rjmc/topology_proposal.py
hannahbrucemacdonald: yes --- it can be changed, but seemed reasonable to me. I'm happy to change it though

perses/rjmc/topology_proposal.py
hannahbrucemacdonald: Yes - unless geometry='weak'. Doing mapping _then_ geometry doesn't work for the CDK2 example I explain above, because of the hydrogen mapping making the top matches wrong.

perses/rjmc/topology_proposal.py
hannahbrucemacdonald: Done

perses/rjmc/topology_proposal.py
hannahbrucemacdonald: In fact it's not going to work for Jnk1 as the ligands are not that closely aligned...

hannahbrucemacdonald:
Ugh this is a pain --- the distance thing _helps_ but doesn't fix it for Jnk1/Thrombin where the rings aren't perfectly aligned... I think I need to change the MCS scoring function so that hydrogens aren't counted in it but I don't know how

hannahbrucemacdonald:
Update:

1) Using geometry='strong' works if the molecules are _very_ well aligned, and doesn't work if they're not, like Jnk1.

2) Using geometry='weak' doesn't work for Jnk1, because the MCS score is always going to be better when the other meta hydrogen maps onto the other meta hydrogen and we won't even get the map with one less atom mapped to measure using cdist, even though it would get the best score.
 - this suggests that we don't want to use hydrogens in the MCS score, _but_ as far as I can see, there's no way that I can personally do that, it would be as simple as changing this to number-heavy-atoms and number-heavy-bonds here: https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemClasses/OEMCSMaxBondsCompleteCycles.html
- we can strip all the hydrogens off the molecule, and do the MCS on that, which will give us both matches. We can then find which of these are unique using cdist, then add restraints so that each of these are then spat out again when the hydrogens are added back in (this is super ugly but the only way I can think)

perses/app/relative_setup.py
```diff
@@ -120,6 +121,8 @@ def __init__(self,
             box dimensions for the complex phase
         solvent_box_dimensions: Vec(3), optional, default=None
             box dimensions for the solvent phase
+        geometry : 'str', default 'strong'
+            wether to use geometry of the second ligand during atom mapping
```
jchodera, line 125: What are the available options here?

PR #696, opened 2020-07-20
fixing the torsion test

hannahbrucemacdonald:
Test fails if torsions are -pi and +pi, which are the same


hannahbrucemacdonald:
https://github.com/omnia-md/conda-recipes/pull/1061

jchodera:
@hannahbrucemacdonald: Thanks for taking a stab at fixing this!

It's been a while, but from what I recall, the `_cartesian_to_internal` method and OpenMM are supposed to operate on the exact same domain and range, and if there's a discrepancy, that indicates a bug with `_cartesian_to_internal`.

@dominicrufa: Do you have enough experience with this method that you think you could fix the method instead of making the test more lax?

jchodera:
I think @dominicrufa needs to dive into the geometry engine and just make sure that `_cartesian_to_internal` has the same domain and range as the OpenMM equivalent functions in order to fix this. We shouldn't change the unit tests to accept disagreement between OpenMM and the geometry engine, since that isn't the right solution.

The OpenMM docs should document the domain and range of each function---if they are not clear, we can open an OpenMM issue and have Peter update them.

hannahbrucemacdonald:
Ok thanks - I've made this a draft so that it doesn't accidentally get wrapped in

jchodera:
I'm unclear if this is still an issue, but we just need to make sure we use the a domain (e.g. `[-pi, +pi)`) for measuring and generating torsion angles that is consistent with OpenMM.

mikemhenry:
Since the tests are passing it is unclear to me what this fixes.

Looking at http://docs.openmm.org/latest/api-c++/generated/OpenMM.CustomTorsionForce.html#customtorsionforce
It looks like the range is `[-pi,+pi]`

>Note that theta is guaranteed to be in the range [-pi,+pi], which may cause issues with force discontinuities if the energy function does not respect this domain.

And we are testing that range:
https://github.com/choderalab/perses/blob/master/perses/tests/test_coordinate_numba.py#L94
`phis = np.arange(-np.pi, np.pi, (2.0*np.pi)/n_divisions)`

jchodera:
@mikemhenry : Here's the situation:

The original test checks that OpenMM and the tools in [`perses/rjmc`](https://github.com/choderalab/perses/tree/master/perses/rjmc) compute torsions the same way, and return results on the same domain [-pi,+pi).

The test failed, suggesting that our implementation in [`perses/rjmc`](https://github.com/choderalab/perses/tree/master/perses/rjmc) is computing torsions on the wrong domain. This is bad.

Instead of fixing the failing method, Hannah changed the test to no longer fail if the domains don't match up. This is even worse, because now we won't know if it's broken.

We need to actually fix the method that is failing to return torsions on the same domain that OpenMM uses.

mikemhenry:
Okay, I thought _this_ PR changed the tests to pass but since the tests are passing on master, I was confused. But it sounds like the tests running on master were changed so they pass, without fixing the underlying issue. 

Can you point me to where the tests were changed? I tried using the history/gitblame to figure it out but didn't have any luck.

I need a failing test in order to check to see if the changes I make to get the domains to match up works. But right now the tests in this PR are passing (which I now understand were changed in a way to pass without fixing the actually issue) but since the tests on master are passing I'm not sure if changes I make will actually fix the problem.

jchodera:
I see the changes in [`test_coordinate_numba.py`](https://github.com/choderalab/perses/pull/696/files) which add an additional branch to not fail if the domains are different. 

I think we need to revert that change. If the tests still pass, that's great: Either someone accidentally fixed the issue in the meantime, or it's a test that only stochastically catches failures.

mikemhenry:
Someone must have fixed it then since `pytest -k test_openmm_dihedral` passes on master, which doesn't have the changes proposed in this PR.

PR #707, opened 2020-08-13
Add library functions for bootstrapping

mcwitt:
This PR pulls out some of the common bootstrapping code into a new submodule, `perses.anaylsis.resample`.

Also includes various fixes:

- Typo in standard error of sum formula: 0860cd5

### DIFF
--- a/perses/analysis/fah_analysis.py
+++ b/perses/analysis/fah_analysis.py
@@ -15,6 +15,7 @@ import joblib
 import logging
 import os
 from typing import Optional
+from perses.analysis.resample import bootstrap_uncorrelated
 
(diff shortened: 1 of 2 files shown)


hannahbrucemacdonald:
This looks great to me! I think we will mostly want to use the correlated version with the folding at home analysis, although the stripping of outliers is an issue as it can result in f/r works of different lengths. Possibly we should be using _strip_outliers in a way that both the f/r works are thrown out if one of them is an outlier, and then use correlated bootstrapping throughout?

Bootstrapping over NEQ switches makes better sense than what I originally wrote!

PR #708, opened 2020-08-14
[READY] Create visualization module for generating PyMOL movies

zhang-ivy:
Visualization module for perses protein mutations and small molecule transformations

TODO
----

- [x] Add test for protein mutations
- [x] Test on xtc ~~and PDB~~ trajectories
    - Note: PDB trajectories are too slow in their read/write times
- [x] Add test for small molecule transformation


hannahbrucemacdonald:
This looks great, and will be very helpful! I think it might be quite helpful to have a few files in an `examples` directory, so that someone could test creating a video without having files already. Also, I wrote a few functions in `perses/analysis/extract_trajectory.py` which are terrible, but it would be useful to have these checked over and possibly integrated, so that it's easier to go from the `.nc` output of a perses calculation to the `.pdb` or `.xtc` files that are the inputs of these functions.

Looks great though and let me know when you want a review!

perses/analysis/visualization.py
```diff
@@ -0,0 +1,404 @@
...
+        cmd.set_name(new_name, "new")
+
+        # Check that the trajectories are the same length
+        old_states = cmd.count_states("old")
+        new_states = cmd.count_states("new")
+        if old_states != new_states:
+            raise Exception(f"The old and new trajectories are not the same length! old: {old_states}, new: {new_states}")
+
+        # Remove solvent
+        _logger.info("Removing solvent...")
```
dominicrufa, line 201: maybe keep solvent near binding site? idk

perses/analysis/visualization.py
```diff
@@ -0,0 +1,404 @@
...
+    frames = ["output/frame-%05d.png" % (i) for i in range(200)] # Create list of file paths for the pngs
+    v.save_mp4(frames)
+
+    # Small molecule transformation example:
+    v = Visualization("old.pdb", "new.pdb", "dcd", "C")
+    v.load()
+    v.format(zoom_distance=3, rotate_x_angle=90)
+    v.save_frames(outfile_prefix="output/frame")
+    frames = ["output/frame-%05d.png" % (i) for i in range(200)] # Create list of file paths for the pngs
+    v.save_mp4(frames)
```
dominicrufa, line 87: i think this is clear enough for me to understand.

dominicrufa (dismissed):
you arent going to add the .pdb, .dcd, .mp4 files, right? it is probbaly best to just make a ~100 frame simulation of something very small (like a capped amino) and run the tests on that (and delete the outputs). looks good otherwise

jchodera:
Can we get this finished up and merged?

perses/analysis/visualization.py
zhang-ivy: Keeping solvent near the binding site isn't super useful as it changes across frames, so watching solvent in a movie just look like a lot of bouncing around

perses/analysis/visualization.py
```diff
@@ -0,0 +1,404 @@
...
+        old_states = cmd.count_states("old")
+        new_states = cmd.count_states("new")
+        if old_states != new_states:
+            raise Exception(f"The old and new trajectories are not the same length! old: {old_states}, new: {new_states}")
+
+        # Remove solvent
+        _logger.info("Removing solvent...")
+        cmd.remove("resn hoh")
+        cmd.remove("resn na")
+        cmd.remove("resn cl")
```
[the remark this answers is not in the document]
zhang-ivy: Solvent is not usually kept when making movies, but if someone wanted to do this they could add it themselves (+ write a test) and make a PR

perses/analysis/visualization.py
```diff
@@ -0,0 +1,404 @@
...
+################################################################################
+
+class Visualization(object):
+    """
+    Visualization tools for perses non-equilibrium switching.
+
+    from perses.analysis.visualization import Visualization
+
+    # Protein mutation example:
+    v = Visualization("old.pdb", "new.pdb", "dcd", "A", mutated_residue="667", ligand_chain="C")
```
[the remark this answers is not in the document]
zhang-ivy: There is a small molecule example right after the protein mutation example. Small molecule transformations have not been tested though -- the reason for this is that I had asked Hannah for the old and new pdbs/dcds from a neq switching free energy calc involving a small molecule transformation and she didn't have it handy. If you can send me one (traj = 10 frames long), I can add it to the test

perses/analysis/visualization.py
```diff
@@ -0,0 +1,404 @@
+"""
+Visualization tools for perses (protein mutations or small molecule transformations).
+Requires PDB and trajectory (.dcd or .xtc) files for old and new atoms.
+Note: Not tested on small molecules yet
```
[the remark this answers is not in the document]
zhang-ivy: I already added the pdb, dcd (10 frames), and mp4 files to `data/visualization`. The test I wrote uses the pdb and dcd to try to generate a protein mutation movie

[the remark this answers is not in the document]
zhang-ivy: Yes there is a difference in the code if the user wants to generate a movie for a small molecule transformation. The code is all there, it just hasn't been tested (see next comment)

perses/analysis/visualization.py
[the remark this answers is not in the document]
zhang-ivy: Added the test for this!

mikemhenry (approved):

mikemhenry (changes requested):
@zhang-ivy really quick question, what is the `old.dcd` file for? I understand we a dcd file for the test but why old and new? Also its is small so not a big deal but what is the `movie.mp4` file for?

perses/analysis/visualization.py
```diff
@@ -0,0 +1,416 @@
...
+    else:
+        raise ImportError("PyMOL is required for this module. Please `conda install -c schrodinger pymol`.")
+
+def _check_mpy():
+    """Check whether moviepy was imported, if not raise exception. """
+    if _FOUND_MPY:
+        return
+    else:
+        raise ImportError(
+            "moviepy is required for this module. Please `conda install -c conda-forge moviepy` or `pip install moviepy`.")
```
ijpulidos, line 45: I think it could make this clearer and less prone to errors if we deal the except directly with a `ModuleNotFoundError`.
```suggestion
try:
    import pymol
    from pymol import cmd
except ModuleNotFoundError as error:
   raise error("PyMOL is required for this module. Please `conda install -c schrodinger pymol`.")

try:
    import moviepy.editor as mpy
except ModuleNotFoundError as error:
    raise error("moviepy is required for this module. Please `conda install -c conda-forge moviepy` or `pip install moviepy`.")
```

ijpulidos (approved):
Nice work!

perses/analysis/visualization.py
zhang-ivy: The reason why I didn't write it with a `ModuleNotFoundError` directly in the `except` is because we don't want to force `pymol` and `moviepy` to be dependencies for perses. Therefore, I only want there to be a `ModuleNotFoundError` or `ImportError` when someone actually tries to use this module (aka when they instantiate a `Visualization` object). The current implementation for throwing the `ImportError` was suggested by Levi Naden (software scientist at molssi), but if you can think of a cleaner way to do it, please let me know!

ijpulidos: Oh okay, I understand. There is probably a cleaner way, but I guess this works for now, that's why it was a non-blocking suggestion. No worries.

mikemhenry: What if instead we just imported `pymol` and `moviepy` instead the `Visualization` class? That would prevent it from being a dependency and simplify this logic.

zhang-ivy: I don't think that will work, because then pymol and moviepy will only be available in the init, and not for the rest of the functions.

mikemhenry (approved):

PR #711, opened 2020-08-17
Moonshot alchemy hbm

hannahbrucemacdonald:


jchodera:
These utility functions have been migrated to [`fah-xchem`](https://github.com/choderalab/fah-xchem).

PR #716, opened 2020-08-24
Adding ionic strength to perses pipeline

hannahbrucemacdonald:


perses/app/relative_setup.py
```diff
@@ -689,7 +693,7 @@ def _generate_vacuum_topologies(self, topology_proposal, old_positions):
 
         return ligand_topology_proposal, old_ligand_positions
 
-    def _solvate_system(self, topology, positions, model='tip3p',phase='complex', box_dimensions=None):
+    def _solvate_system(self, topology, positions, model='tip3p',phase='complex', box_dimensions=None,ionic_strength=0.15*unit.molar):
```
dominicrufa, line 696: probably best not to hardcode the conc here, but instead reference to `self._ionic_strength`?

perses/app/setup_relative_calculation.py
```diff
@@ -73,6 +73,16 @@ def getSetupOptions(filename):
         setup_options['protocol-type'] = 'default'
 
 
+    if 'temperature' not in setup_options:
```
dominicrufa, line 76: im being nitpicky, but maybe it would be best to put some of the defaults into the documentation?

hannahbrucemacdonald: That's a good point. I have been working on the documentation in a different branch, so I will update it there

Issue #722, opened 2020-08-25
Turn off DEBUG logging


jchodera:
When running the latest `master`, `stdout` is filled with ~8000 lines of debug output. Is there a way to turn this off?

hannahbrucemacdonald:
Closed by #725

PR #726, opened 2020-08-26
Trying to get codecov to work

hannahbrucemacdonald:


hannahbrucemacdonald:
@jaimergp could you help with this please? I'm not sure why it's not working.

I made some changes to the GHactions, as I think it might be that nosetests needs to generate a file that can be uploaded to codecov, but it also looked like codecov didn't have repo access to perses (which I have now added). If this doesn't work, I'm not sure what else to try!!

hannahbrucemacdonald:
Closes #722

hannahbrucemacdonald:
Ok! I think that because codecov is complaining that there is nothing in master to compare against, this isn't going to work, _until_ it's merged into master...?

hannahbrucemacdonald:
Merging my own changes as this is non-code breaking

PR #739, opened 2020-09-10
fix CI.yaml

zhang-ivy:
#682 introduced parsing errors in `.github/workflows/CI.yaml`: `matrix.cfg.os` in this [line](https://github.com/choderalab/perses/blob/master/.github/workflows/CI.yaml#L17) is no longer defined in the [matrix](https://github.com/choderalab/perses/blob/master/.github/workflows/CI.yaml#L20).

Thus, tests are failing for subsequent PRs.

I fixed the `matrix.cfg.os` lines based on [this commit](https://github.com/choderalab/perses/commit/877b17db943b8852a279fc0aa57420ff27386682), however, the current CI.yaml in master has other differences from that commit, so I'm not sure if there are more undesired changes introduced by #682.


jaimergp:
Ok, that did it. @zhang-ivy there's a single error due to some stereochemistry issues. I'll let you take over from here :)

PR #761, opened 2021-02-01
fix coverage report + remove old CI

mikemhenry:
Right now there is a lot of CI spam from coverage reports on packages we don't care about, I think this is a vestigial organ left over from when CI was on travis, I don't believe this env variable gets set on github actions like it did on travis.

I'm going to use this PR to cleanup some old stuff laying around from when the CI was on travis.

PR #763, opened 2021-02-12
Switch to pytest runner

mikemhenry:
As a first pass in improving our testing, we are going to switch to using [`pytest`](https://docs.pytest.org/en/stable/) to run tests. Because of the [compatibility between `pytest` and `nose`](https://docs.pytest.org/en/stable/nose.html) we shouldn't need to do anything to the old tests, but we should start writing new tests using `pytest` features.

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -21,8 +21,8 @@ jobs:
         cfg:
           - { os: ubuntu-latest, python-version: 3.7, openmm: latest }
           - { os: ubuntu-latest, python-version: 3.8, openmm: latest }
-          - { os: ubuntu-latest, python-version: 3.7, openmm: nightly }
-          - { os: ubuntu-latest, python-version: 3.8, openmm: nightly }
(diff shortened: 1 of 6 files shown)


mikemhenry:
Note: I biased this PR off of #761 since it will be merged soon and fixes some issues with the testing environment.

jaimergp:
Awesome, works right away!

Minor things:
* [This kind of warning](https://github.com/choderalab/perses/pull/763/checks?check_run_id=1889827825#step:9:149) is probably due to a docstring containing a literal `\P` (as in `C:\Program Files`), or maybe some kind of latex formula. They will go away as soon as you turn those docstrings into raw-docstrings (`r"""some text"""`)

By the way, this might be a good moment to enable tests for the `examples/` directory in a separate GHA step, if that's not planned (also in a different PR if you wish).

mikemhenry:
Thanks! I will go through and fix those warnings on this PR since they really clog up the CI results. I've made a plan here to add the examples to the CI here https://github.com/choderalab/perses/issues/766

Issue #760, opened 2021-02-01
Fix CI


mikemhenry:
Working on getting CI to pass, looks like it is do to an import for `openforcefield` failing, which may be due to this PR

https://github.com/openforcefield/openff-toolkit/pull/803

jchodera:
@mikemhenry : We should upgrade the CI to GitHub Actions after updating to the new [MolSSI Cookiecutter](https://github.com/MolSSI/cookiecutter-cms).

I'd reach out to @jaimergp to have him give you pointers on this process the first time if you haven't done either a Cookiecutter upgrade or a GitHub Actions update.

Let's also start tagging issues with their assignees, labels, and eventually milestones/projects. Perhaps you can come up with a labeling scheme that would make some sense?

Let me know if you need any repo or org permissions!

jaimergp:
This repo was already migrated to GH Actions (see [here](https://github.com/choderalab/perses/actions)). There might be some PRs opened that predate that contribution, so they'd a rebase in that case. The openforcefield migration is surely a cause of failures. Check this for instructions: https://github.com/openforcefield/openff-toolkit/issues/819

mikemhenry:
Something I didn't quite understand (but I have figured it now when switching over to pytest) is that on our CI we skip tets marked with `@skipIf(running_on_github_actions)` and `@attr('advanced')` so while things are passing, we have tests that fail but were not caught by CI. I found these when running the tests locally since I was not on GitHub actions and I just used `pytest -v` to run the tests. I believe that many of the tests that fail locally likely just need to be updated and have fallen behind as the API has changed.

This also explains why CI takes 20 minutes but when running all tests it took 3 hours!

jaimergp:
I think those were too resource intensive to be run on Travis (where the limit was 1h), and I just [converted](https://github.com/choderalab/perses/pull/617) those `skipIf` directly without reconsidering a "promotion".

Or maybe some might need a GPU too. 🤷

PR #765, opened 2021-02-12
run nightly nightly

mikemhenry:
Instead of including checks against nightly openmm builds on ever PR, we will run checks every night and if they fail make an issue on github

This PR will be smaller when PR #760  and PR #763 get merged in.


mikemhenry:
@jaimergp Not exactly what is wrong with the syntax in the new Nightly.yml file, but my idea was to move the nightly testing as part of a cron job and not test PRs against nightly. This _should_ make a github issue when nightly fails, which should help keep those issues visible.

.github/workflows/Nightly.yml
```diff
@@ -0,0 +1,100 @@
...
+          nosetests perses --with-coverage --cover-package=perses --verbosity=3 --with-timer -a '!advanced' --cover-xml --cover-xml-file=./coverage.xml
+          popd
+
+      - name: Make an issue
+      - uses: JasonEtco/create-an-issue@v2
+        if: ${{ failure() }}
+        env:
+            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+             assignees: mikemhnery
```
jaimergp, line 99:
```suggestion
             assignees: mikemhenry
```

:)

jaimergp:
What if instead of having different workflows we use conditionals based on [`github.event_name`](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context)?

.github/workflows/Nightly.yml
mikemhenry: oops :name_badge:

mikemhenry:
That could work and cuts down on the duplication. At one point I split the checks into a new workflow so they wouldn't block auto merging or make it look like our build was failing when it wasn't, but I think marking the checks as not required should fix that, I'll give it a try.

.github/workflows/Nightly.yml
```diff
@@ -0,0 +1,100 @@
...
+on:
+  schedule:
+    # nightly tests
+    - cron: "0 0 * * *"
+  # Be able to run manually
+  workflow_dispatch:
+
+jobs:
+  test:
+    name: Test on ${{ matrix.cfg.os }}, Python ${{ matrix.cfg.python-version }}, OpenMM ${{ matrix.cfg.openmm }}
```
jaimergp, line 12: What about this, but on `.github/workflows/CI.yaml`?

```suggestion
    if: (github.event_name == 'schedule'  && matrix.cfg.nightly) || github.event_name != 'schedule' 
    name: Test on ${{ matrix.cfg.os }}, Python ${{ matrix.cfg.python-version }}, OpenMM ${{ matrix.cfg.openmm }}
```

You'll need a `nightly` key on the relevant matrix entries.

jaimergp: The individual steps will need to check against `if: matrix.cfg.nightly` only.

mikemhenry: I cut those when at first I was dropping nightly from running on PRs

mikemhenry:
Part of what I'm trying to do is not have our build badge be red, but I think I've fixed that by chancing which checks are required

mikemhenry:
Going to work on making nightly work

PR #767, opened 2021-02-12
remove unused imoports and expand wildcards

mikemhenry:
This is part of my refactoring project.

All unused imports have been removed and wild cards have been expanded.
I've also removed unnecessary `pass` statements 

It is possible something will be broken if there are some run time effects on import, but we should fix those!

### DIFF
--- a/perses/analysis/analysis.py
+++ b/perses/analysis/analysis.py
@@ -139,7 +139,6 @@ class Analysis(object):
         :param environment:
         :return:
         """
-        pass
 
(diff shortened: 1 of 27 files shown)


mikemhenry:
Ah I forgot that because of the way the license works, I can work on a fork of this repo but need to work directly.

jchodera:
+1 for the general practice of working on a branch of the main repo instead of a fork, due to license secrets for CI.

PR #768, opened 2021-02-12
Feat/remove unused imports and expand wildcards

mikemhenry:
This is part of my refactoring project.

All unused imports have been removed and wild cards have been expanded.
I've also removed unnecessary pass statements

It is possible something will be broken if there are some run time effects on import, but we should fix those!

### DIFF
--- a/examples/SamplerExample/run.py
+++ b/examples/SamplerExample/run.py
@@ -1,6 +1,5 @@
 import yaml 
 import sys
-import itertools
 import os
 
 def run_relative_perturbation(ligA, ligB,tidy=True):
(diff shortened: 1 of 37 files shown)


mikemhenry:
Just an FYI, to simulate running on CI locally:

`GITHUB_ACTIONS=True pytest -v --durations=0 -a "not advanced"`

jchodera (approved):
Bravo! This is a great step in the right direction.

In future, it would be useful to move all imports into the methods that need them ("lazy" imports), which serves multiple functions:
* Simply running anything that imports any part of perses takes a long time because of all the file-level imports
* The file-level imports also make it wickedly complicated to get the readthedocs env right, and certainly aren't needed for introspection to generate API docs (until we start using type annotation)
* We want to keep `openeye` imports local to functions that need it in preparation for eventual replacement with `openff-toolkit`

These don't need to be done now, but are things to think about as you continue refactoring!

PR #770, opened 2021-02-22
Add support for inserting/deleting counterions when changing charge to maintain net neutrality

dominicrufa:
- [x] add counterion decoupling in `topology_proposal` for charge-changing mutations
- [x] write test to identify appropriate counterions to decouple
- [ ] update to master


jchodera:
Awesome! Let's have @mikemhenry also help with reviewing this PR.

Note that the branch is out of date, and you'll need to rebase or merge `master`.

perses/app/relative_point_mutation_setup.py
```diff
@@ -250,6 +250,23 @@ def __init__(self,
             validate_bool = False if old_res.name in ring_amino_acids or proposed_residue in ring_amino_acids else True
             new_positions, logp_proposal = geometry_engine.propose(topology_proposal, pos, beta,
                                                                    validate_energy_bookkeeping=validate_bool)
+
+            #check for charge change...
+            charge_diff = point_mutation_engine._get_charge_difference(current_resname = topology_proposal._old_topology.residue_topology.name,
+                                                                       new_resname = topology_proposal._new_topology.residue_topology.name)
+            if charge_diff != 0:
```
jchodera, line 257: Are fractional (but near zero) charges a concern? Or is `charge_diff` an integer?

perses/app/relative_point_mutation_setup.py
```diff
@@ -250,6 +250,23 @@ def __init__(self,
...
+                new_ion_indices_to_neutralize = point_mutation_engine.get_counterion_indices(charge_diff,
+                                                                       old_res = topology_proposal._old_topology.residue_topology,
+                                                                       new_res = topology_proposal._new_topology.residue_topology,
+                                                                       new_positions = new_positions,
+                                                                       new_topology = topology_proposal._new_topology,
+                                                                       positive_ion_name='NA',
+                                                                       negative_ion_name='CL',
+                                                                       radius=0.3)
```
jchodera, line 265: Should this be something that uses units? 0.3 nm?

perses/app/relative_point_mutation_setup.py
```diff
@@ -313,6 +330,32 @@ def get_complex_htf(self):
     def get_apo_htf(self):
         return self.apo_htf
 
+    @staticmethod
+    def _modify_new_system(counterions_to_neutralize, system):
```
jchodera, line 334: Could we use a more informative name, such as `_neutralize_system`?

perses/rjmc/topology_proposal.py
```diff
@@ -1401,6 +1401,34 @@ class PolymerProposalEngine(ProposalEngine):
     This base class is not meant to be invoked directly.
     """
 
+    _aminos = ['ALA', 'ARG', 'ASN', 'ASP', 'CYS', 'GLN', 'GLU', 'GLY', 'HIS', 'ILE', 'LEU', 'LYS', 'MET', 'PHE',
```
jchodera, line 1404: Was this accidentally swept in from another PR?

perses/rjmc/topology_proposal.py
```diff
@@ -1470,6 +1474,98 @@ def generate_oemol_from_pdb_template(pdbfile):
...
+        chargediff : int
+            charge(new_res) - charge(old_res)
+        """
+        assert new_resname in PolymerProposalEngine._aminos
+        assert current_resname in PolymerProposalEngine._aminos
+
+        new_rescharge, current_rescharge = 0,0
+        resname_to_charge = {current_resname: 0, new_resname: 0}
+        for resname in [new_resname, current_resname]:
+            if resname in PolymerProposalEngine._negative_aminos:
```
jchodera, line 1500: How do we treat residues in different protonation states? Do we require that we use `HIS`, `HID`, `HIE` names to access different protonation states of histidine, for example? I know OpenMM uses PDB standard names, not AMBER protonation state variant names.

Is there a more robust way to determine the net charge on the system, such as using the total partial charges from the force field from the unmodified system? That would presumably be more robust, even if you just have to round the fractional charges.

perses/rjmc/topology_proposal.py
```diff
@@ -1470,6 +1474,98 @@ def generate_oemol_from_pdb_template(pdbfile):
...
+            new mutating residue
+        new_positions : np.ndarray(N, 3)
+            positions of atoms corresponding to new_topology
+        new_topology : openmm.Topology
+            topology of new system (with new_res)
+        positive_ion_name : str, default 'NA'
+            name of positive ions to identify as candidates for neutralization (mdtraj naming convention)
+        negative_ion_name : str, default 'CL'
+            name of negative ions to identify as candidates for neutralization (mdtraj naming convention)
+        radius : float, default 0.3
```
jchodera, line 1535: Can we make this a unit-bearing `Quantity`?

perses/rjmc/topology_proposal.py
```diff
@@ -1470,6 +1474,98 @@ def generate_oemol_from_pdb_template(pdbfile):
...
+
+        Parameters
+        ----------
+        charge_diff : int
+            the charge difference between the new_system - old_system
+        old_res : openmm.Topology.residue
+            old mutating residue
+        new_res : openmm.Topology.residue
+            new mutating residue
+        new_positions : np.ndarray(N, 3)
```
jchodera, line 1527: Are these dimensionless (and in which unit convention) or unit-bearing? Perhaps we should make them unit-bearing `Quantity` objects?

perses/rjmc/topology_proposal.py
```diff
@@ -1401,6 +1401,34 @@ class PolymerProposalEngine(ProposalEngine):
     This base class is not meant to be invoked directly.
     """
 
+    _aminos = ['ALA', 'ARG', 'ASN', 'ASP', 'CYS', 'GLN', 'GLU', 'GLY', 'HIS', 'ILE', 'LEU', 'LYS', 'MET', 'PHE',
+                    'SER', 'THR', 'TRP', 'TYR', 'VAL'] # common naturally-occurring amino acid names
+                    # Note this does not include PRO since there's a problem with OpenMM's template DEBUG
+    _positive_aminos = ['ARG', 'HIS', 'LYS']
+    _negative_aminos = ['ASP', 'GLU']
+    _neutral_aminos =  [amino for amino in _aminos if amino not in ['ARG', 'HIS', 'LYS', 'ASP', 'GLU']]
```
mikemhenry, line 1409:
```suggestion
    _neutral_aminos =  [amino for amino in _aminos if amino not in (_positive_aminos + _negative_aminos)]
```

This will be easier to maintain and prevent mistakes of updating aminos in one spot but not propagating them in other spots

perses/rjmc/topology_proposal.py
```diff
@@ -1470,6 +1474,98 @@ def generate_oemol_from_pdb_template(pdbfile):
...
+        for resname in [new_resname, current_resname]:
+            if resname in PolymerProposalEngine._negative_aminos:
+                resname_to_charge[resname] -= 1
+            elif resname in PolymerProposalEngine._positive_aminos:
+                resname_to_charge[resname] += 1
+
+        return resname_to_charge[new_resname] - resname_to_charge[current_resname]
+
+    @staticmethod
+    def get_counterion_indices(charge_diff,
```
mikemhenry, line 1508: Can we call `_get_charge_difference` inside this method instead of providing the `charge_diff`? Like is `old_res` and `new_res` here different from the ones that will be specified in the `current_resname` and `new_resname`?

perses/rjmc/topology_proposal.py
```diff
@@ -1470,6 +1474,98 @@ def generate_oemol_from_pdb_template(pdbfile):
...
+        atoms = list(traj.topology.select(f"resname {ion_name}"))
+
+        query_indices = traj.top.select('protein')
+
+        # Get ion indices within radius of mutated residue
+        neighboring_ions = list(md.compute_neighbors(traj, radius, query_indices, haystack_indices=atoms))
+
+        # Get ion indices outside of radius of mutated residue
+        nonneighboring_ions = [atom for atom in atoms if atom not in neighboring_ions]
+        assert len(nonneighboring_ions) > 0, f"there are no available nonneighboring ions"
```
mikemhenry, line 1563: Did you mean to have some more debugging information here? If not, I think you should drop the f-string formatting

perses/tests/test_topology_proposal.py
```diff
@@ -905,9 +905,60 @@ def test_simple_heterocycle_mapping(iupac_pairs = [('benzene', 'pyridine')]):
 
         assert num_hetero_maps > 0, f"there are no differences in atomic number mappings in {iupac_pair}"
 
+def test_protein_counterion_topology_fix():
```
mikemhenry, line 908: I didn't benchmark this test's run time but if it is "short" then can we add test cases where the charge +1, and zero?

mikemhenry (changes requested):
Mostly questions, I agree with @jchodera that we should add unit quantities to everything we can.

perses/rjmc/topology_proposal.py
zhang-ivy: I think this line (and the lines below it) were moved here so that we could identify which amino acids are positive vs negative. This distinction is necessary because we determine the `charge_diff` by subtracting `charge_of_old_aa` - `charge_of_new_aa`

perses/rjmc/topology_proposal.py
zhang-ivy: >How do we treat residues in different protonation states? Do we require that we use HIS, HID, HIE names to access different protonation states of histidine, for example?

It's on my TODO list to fix how we handle protonation states in Perses. Currently, I think we always choose `HIS`. 

>Is there a more robust way to determine the net charge on the system

This would be more robust, which would be useful for non-protein systems. I'll open an issue for us to address that later.

perses/rjmc/topology_proposal.py
zhang-ivy: These do not need to be `Quantity` objects, as they are just being fed into mdtraj.

zhang-ivy (approved):
I've updated the PR with John's suggested water->ion approach.
I've responded to Mike and John's comments.
This is ready for a second review!

perses/rjmc/topology_proposal.py
zhang-ivy: Fixed this!

perses/app/relative_point_mutation_setup.py
```diff
@@ -313,6 +330,32 @@ def get_complex_htf(self):
...
+        -------
+        modify system in place
+
+        """
+        force_dict = {i.__class__.__name__: i for i in system.getForces()}
+        if 'NonbondedForce' in [i for i in force_dict.keys()]:
+            nbf = force_dict['NonbondedForce']
+            for idx in counterions_to_neutralize:
+                charge, sigma, eps = nbf.getParticleParameters(idx)
+                nbf.setParticleParameters(idx, charge*0.0, sigma, eps*0.0)
```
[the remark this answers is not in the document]
zhang-ivy: The new implementation no longer zeroes the epsilon

perses/rjmc/topology_proposal.py
```diff
@@ -1470,6 +1473,83 @@ def generate_oemol_from_pdb_template(pdbfile):
...
+        query_atoms = traj.top.select('protein')
+
+        # Get water atoms within radius of protein
+        neighboring_atoms = md.compute_neighbors(traj, radius, query_atoms, haystack_indices=water_atoms)[0]
+
+        # Get water atoms outside of radius of protein
+        nonneighboring_residues = set([atom.residue.index for atom in traj.topology.atoms if atom.index not in  neighboring_atoms])
+        assert len(nonneighboring_residues) > 0, "there are no available nonneighboring waters"
+        # Choose N random nonneighboring waters, where N is determined based on the charge_diff
+        choice_residues = np.random.choice(list(nonneighboring_residues), size=abs(charge_diff), replace=False)
```
mikemhenry, line 1544: Do we ever set a `perses` wide random seed? It doesn't look like we do and while it is outside the scope of this PR we should let the user specify the seed to enable reproducibility.

mikemhenry (changes requested):

perses/app/relative_point_mutation_setup.py
```diff
@@ -152,9 +161,10 @@ def __init__(self,
                 in the htf, flatten torsions involving unique new atoms at lambda = 0 and unique old atoms are lambda = 1
             flatten_exceptions : bool, default False
                 in the htf, flatten exceptions involving unique new atoms at lambda = 0 and unique old atoms at lambda = 1
-            repartitioned_endstate : int, default None
-                the endstate (0 or 1) at which to build the RepartitionedHybridTopologyFactory. By default, this is None,
-                meaning a vanilla HybridTopologyFactory will be built.
+            vanilla : bool, default True
+                whether to generate a vanilla HybridTopologyFactory
+            repartitioned : bool, default False
+                whether to generate a RepartitionedHybridTopologyFactory
```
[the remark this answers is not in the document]
zhang-ivy: When I run REST, I need `repartitioned =True`. When I run neq switching or repex only, then I need `vanilla=True`. If I want to run both REST and neq switching, then I need both to be `True`. Just wanted to provide flexibility here.

perses/app/relative_point_mutation_setup.py
```diff
@@ -300,20 +302,125 @@ def __init__(self,
...
+        return self.apo_rhtf_0
+    
+    def get_complex_rhtf_1(self):
+        return self.complex_rhtf_1
+
+    def get_apo_rhtf_1(self):
+        return self.apo_rhtf_1
+
+    @staticmethod
+    def _modify_new_system(water_atoms, system, charge_diff):
```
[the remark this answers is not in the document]
zhang-ivy: Sure either of those sounds fine -- could you make the suggested change?

perses/app/relative_point_mutation_setup.py
```diff
@@ -300,20 +302,125 @@ def __init__(self,
...
+        force_dict = {i.__class__.__name__: i for i in system.getForces()}
+        if 'NonbondedForce' in [i for i in force_dict.keys()]:
+            nbf = force_dict['NonbondedForce']
+            for idx in water_atoms:
+                idx = int(idx)
+                charge, sigma, epsilon = nbf.getParticleParameters(idx)
+                if charge == O_CHARGE:
+                    nbf.setParticleParameters(idx, ion_charge, ion_sigma, ion_epsilon)
+                elif charge == H_CHARGE:
+                    nbf.setParticleParameters(idx, charge*0.0, sigma, epsilon)
```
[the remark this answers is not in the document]
zhang-ivy: I don't think we ever need to turn an ion into a water.. We are only ever changing a water into a positive or negative ion.

perses/rjmc/topology_proposal.py
zhang-ivy: I don't think so, but feel free to create an issue about this if you think we should make a separate PR that addresses this!

perses/rjmc/topology_proposal.py
```diff
@@ -1470,6 +1473,83 @@ def generate_oemol_from_pdb_template(pdbfile):
...
+        Returns
+        -------
+        ion_indices : np.array(abs(charge_diff)*3)
+            indices of water atoms to be turned into ions
+        """
+
+        import mdtraj as md
+        # Create trajectory
+        traj = md.Trajectory(new_positions[np.newaxis, ...], md.Topology.from_openmm(new_topology))
+        water_atoms = list(traj.topology.select(f"water"))
```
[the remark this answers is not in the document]
zhang-ivy: Good catch. It doesn't. Can you make the suggested change?

perses/rjmc/topology_proposal.py
zhang-ivy: Hmm, not sure..

zhang-ivy: If it doesn't accept `Quantity` objects, then we will see an mdtraj error, so I'm not sure that we need to change anything here

perses/rjmc/topology_proposal.py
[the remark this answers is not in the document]
mikemhenry:
```suggestion
        water_atoms = list(traj.topology.select("water"))
```

perses/app/relative_point_mutation_setup.py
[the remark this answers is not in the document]
mikemhenry: :+1:

perses/app/relative_point_mutation_setup.py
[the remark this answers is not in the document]
mikemhenry: made change in 323472926b41941ca8c8d618f65034324080293c

perses/rjmc/topology_proposal.py
zhang-ivy: It should not be unit bearing, as it will be fed into mdtraj.compute_neighbors, which takes in an int.

perses/rjmc/topology_proposal.py
zhang-ivy: `old_res` is a residue object, whereas `current_resname` is a name.
I don't see a significant reason why its better to call `_get_charge_difference` inside `get_water_indices` over the way it is currently. If you disagree, feel free to modify and I'll approve it.

perses/tests/test_topology_proposal.py
zhang-ivy: Sure, feel free to add those.

perses/rjmc/topology_proposal.py
mikemhenry: That makes sense!

perses/rjmc/topology_proposal.py
jchodera: Our APIs should strive to standardize how unit-bearing quantities are passed. We should use unit-bearing quantities, check to make sure the dimensionality is correct internally, and convert internally into whatever the tools we use under the hood (e.g. MDTraj, which uses nm).

@mikemhenry Maybe you can help us clean up some of these API issues?

mikemhenry (approved):

perses/tests/test_topology_proposal.py
```diff
@@ -905,20 +905,144 @@ def test_simple_heterocycle_mapping(iupac_pairs = [('benzene', 'pyridine')]):
 
         assert num_hetero_maps > 0, f"there are no differences in atomic number mappings in {iupac_pair}"
 
+def test_protein_counterion_topology_fix_positive():
```
mikemhenry, line 908: @zhang-ivy can you check this test and the other two that I added? The bits I changed:

* changed `charge_diffs` to `charge_diff_test` 
* removed `counterion_name` since it was unused
*  added a negative and zero test for charges

zhang-ivy: Looks good to me!

perses/rjmc/topology_proposal.py
mikemhenry: Tracking that here:
https://github.com/choderalab/perses/issues/790

perses/rjmc/topology_proposal.py
```diff
@@ -1466,6 +1474,83 @@ def generate_oemol_from_pdb_template(pdbfile):
...
+    @staticmethod
+    def _get_charge_difference(current_resname, new_resname):
+        """
+        return the charge of the old res - charge new res
+
+        Parameters
+        ----------
+        current_resname : str
+            three letter identifier for original residue
+        new_resnme : str
```
zhang-ivy, line 1486:
```suggestion
        new_resname : str
```

perses/rjmc/topology_proposal.py
```diff
@@ -1466,6 +1474,83 @@ def generate_oemol_from_pdb_template(pdbfile):
...
+                               radius=0.8):
+        """
+        Choose random water(s) (at least `radius` nm away from the protein) to turn into ion(s). Returns the atom indices of the water(s) (index w.r.t. new_topology)
+
+        Parameters
+        ----------
+        charge_diff : int
+            the charge difference between the old_system - new_system
+        new_positions : np.ndarray(N, 3)
+            positions of atoms corresponding to new_topology
```
zhang-ivy, line 1520:
```suggestion
            positions (nm) of atoms corresponding to new_topology
```

mikemhenry (approved):

perses/app/relative_point_mutation_setup.py
```diff
@@ -21,10 +21,18 @@
 kT = kB * temperature
 beta = 1.0/kT
 ring_amino_acids = ['TYR', 'PHE', 'TRP', 'PRO', 'HIS']
+CL_CHARGE = unit.Quantity(value=-1.0, unit=unit.elementary_charge)
```
jchodera, line 24: Do these need to be hard-coded? If you have the `Topology` object and an unmodified system, can't you just pull these from the `System` object directly?

perses/app/relative_point_mutation_setup.py
```diff
@@ -103,7 +111,8 @@ def __init__(self,
                  apo_box_dimensions=None,
                  flatten_torsions=False,
                  flatten_exceptions=False,
-                 repartitioned_endstate=None,
+                 vanilla=True,
```
jchodera, line 114: Can we use a more informative name than `vanilla`? And instead of selecting "generate the unmodified form" if `True`, perhaps we could instead use a flag for generating the modified form, giving it a more informative name?

zhang-ivy: The flag below this line is `repartitioned=True`, which allows generating the repartitioned htf. I want to allow the user to only generate the repartitioned htf or the vanilla htf or both, but I could change the flags to be `generate_vanilla_htf` and `generate_repartitioned_htf` if you prefer.

perses/app/relative_point_mutation_setup.py
zhang-ivy: I'm not sure we want to pull the charge from the system -- what if the system doesn't have any Na+ or Cl- ions (and instead has a different type of ion)? Making it system dependent might also propagate any errors in parametrization (even if the chance for this to occur is very small).

jchodera: I think the more important thing is that we want to use parameters that are compatible with the force field. This essentially hard-codes the force field, but when the water or ion model is updated, you'll be introducing weird mixtures of parameters that may not work together.

We can always fall back to defaults in there are no monovalent counterions in the system, but it is quite likely we will always want to be running with physiological or experimental salt concentration, rather than minimal neutralizing counterion concentrations.

perses/app/relative_point_mutation_setup.py
jchodera: Ah, I see! Yes, that would be much clearer, though perhaps we could use clearer names?

Perhaps `generate_unmodified_hybrid_topology_factory` and `generate_rest_capable_hybrid_topology_factory`.
They're long, but they're unambiguous.

zhang-ivy: Fixed

perses/app/relative_point_mutation_setup.py
zhang-ivy: Fixed

zhang-ivy:
@mikemhenry @jchodera -- I've added fixes per John's suggestions: 1) change the flag names for `PointMutationExecutor` to be more descriptive and 2) removed the hard coded counterion/water nonbonded parameters and 3) set the temperature back to 300 K (because using 298 K broke things).
Please review again -- hopefully this is ready to be merged.

Issue #673, opened 2020-05-19
Undefined stereochemistry for CDK2 ligand 11


hannahbrucemacdonald:

hannahbrucemacdonald:
<img width="564" alt="Screen Shot 2020-05-19 at 2 56 53 PM" src="https://user-images.githubusercontent.com/43652597/82335998-f11b2200-99e1-11ea-8731-7b492d708bbc.png">

The thione here is recognised by openeye as being chiral.

In `perses.utils.openeye.createOEMolFromSDF` we do two checks:
`oechem.OE3DToInternalStereo(molecule)`  --- which uses the coordinates to work out the stereochemistry of the atoms based on their coordinates _and_
`has_undefined_stereocenters`

Firstly, it says:
`Warning: OE3DToAtomStereo is unable to perceive atom stereo from a flat geometry on atom 44 of molecule '31'`, but if you call `oechem.OE3DToInternalStereo(molecule)` again then it returns True so I'm not sure what's going on.

Secondly, `has_undefined_stereocenters` because S is chiral, but the stereochemistry is undefined. The major thing is that I'm not completely sure what sort of stereochemistry this _should_ be, I'm guessing there is conjugation with the aromatic ring, so the S-C bond has double bond characteristics, so it's sort of an E/Z isomerism? Who knows though.

I could force this to run by just making checking stereochemistry an option or a warning, but I'm not sure that's the right thing to do

jchodera:
@hannahbrucemacdonald : A few things here:
* We should deprecate `perses.utils.openeye.createOEMolFromSDF` in favor of always going through the `openforcefield` toolkit. The equivalent call is
```python
from openforcefield.topology import Molecule
molecule = Molecule.from_file('filename.sdf')
molecule.to_openeye()
```
* Whenever you submit an issue, please provide sufficient information to reproduce the issue---ideally with a minimal test case. If you don't, we either have to spend a lot of time digging to try to figure out what issue you're actually running into, or there is simply no way for us to help. 

I am going to presume you are working with the files in `perses/data/cdk2-example/`. Attempting to read the `CDK2_ligands.sdf` file does indeed result in a stereochemistry error for the ligand with the title `31`:
```
>>> from openforcefield.topology import Molecule
>>> molecules = Molecule.from_file('CDK2_ligands.sdf')
Warning: OE3DToAtomStereo is unable to perceive atom stereo from a flat geometry on atom 44 of molecule '31'
Traceback (most recent call last):
[...]

PR #778, opened 2021-03-16
add allow_undefined_stereo option for SDF files with undefined stereocenters

glass-w:
This is directly related to #673. 

In this PR I've added an option to set `allow_undefined_stereo` when creating an `OEMol` object using `createOEMolFromSDF` in `relative_point_mutation_setup.py`.

@schallerdavid and I are working with sets of ligands that contain undefined stereocenters after preparation with OpenEye (this has been reported to them and they are working on a fix). Until this is solved (or we move to `openforcefield`, as discussed in #657) we need an option to relax the requirement that all stereocenters need to be defined.

### DIFF
--- a/perses/app/relative_point_mutation_setup.py
+++ b/perses/app/relative_point_mutation_setup.py
@@ -91,6 +91,7 @@ class PointMutationExecutor(object):
                  conduct_endstate_validation=True,
                  ligand_input=None,
                  ligand_index=0,
+                 allow_undefined_stereo_sdf=False,
                  water_model='tip3p',
                  ionic_strength=0.15 * unit.molar,
                  forcefield_files=['amber14/protein.ff14SB.xml', 'amber14/tip3p.xml'],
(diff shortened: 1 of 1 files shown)


jchodera:
@glass-w @schallerdavid : Can you clarify your specific issue here? Which one of these is it:
* Using SMILES strings to specify ligands that have undefined stereocenters
* Using 2D SDF files that do not specify stereochemistry
* Using 3D SDF files with flat nitrogens (or other atom OE believes are chiral)

schallerdavid:
@jchodera ,

this issue was about larotrectinib and selitrectinib, which has an almost planar sp3 nitrogen. When reading the sdf of the docking pose into perses, we received an undefined stereo center error.

jchodera:
Thanks for the clarification! We should deprecate everything in `perses.utils.openeye` as part of continued refactoring to switch to `openff-toolkit`.

@glass-w: You'll have to rebase or update to master, but once you do, we can merge!

glass-w:
@mikemhenry this should be ready to merge

PR #783, opened 2021-04-09
Change default temp in protein mutation setup pipeline

zhang-ivy:
Currently, the `temperature` is defaulted at 300 K, but it should be 298 K (as the default temperature for integrators in `openmmtools` is set to 298 K and 298 K is closer to 25 celsius).


mikemhenry (approved):
Looks good! I was going to add a note to put this in the change log but doesn't look like we are maintaining one :scream: 
https://github.com/choderalab/perses/blob/master/docs/changelog.rst

I mention that since this will be a change that will "break" reproducibility potentially and wanted to make sure we highlight it. Right now we kinda keep track of changes on the releases page.

I'll make an issue out of this!

PR #785, opened 2021-04-20
[WIP] reaction field

dominicrufa:
implementing reaction field hybrid factory

https://pubs.rsc.org/am/content/articlehtml/2020/cp/d0cp03835k


perses/annihilation/relative.py
```diff
@@ -2541,3 +2537,269 @@ def _handle_exceptions(self):
...
+    def render_bool_string(bool_variable_prefixes, bool_variable_names):
+        selections = [f"{bool_variable_prefix} * {bool_variable_name}" for bool_variable_prefix, bool_variable_name in zip(bool_variable_prefixes, bool_variable_names)]
+        return '( ' + '+'.join(selections) + ' )'
+
+
+    def _transcribe_bonds(self):
+        """
+        handle the harmonic bonds...
+        """
+        scale_bool_string = RxnHybridTopologyFactory.render_bool_string(['1', 'scale_lambda', 'interscale_lambda'],
```
zhang-ivy, line 2717: We should allow for multiple, disjoint scale regions. 
The disjoint-ness should be checked in the init()

perses/annihilation/relative.py
```diff
@@ -2541,3 +2537,269 @@ def _handle_exceptions(self):
...
+                                                                    ['nonscale_region', 'scale_region', 'interscale_region'])
+        core_bond_expression = f"{scale_bool_string} * (K/2)*(r-length)^2;"
+        bool_string = RxnHybridTopologyFactory.render_bool_string([''] + [f"lambda_{i}_bonds" for i in range(self._num_alchemical_regions)],
+                                                                    ['environment_region'] + [f"alchemical_region_{i}" for i in range(self._num_alchemical_regions)])
```
zhang-ivy, line 2721: Add comment defining the environment region

perses/annihilation/relative.py
```diff
@@ -2541,3 +2537,269 @@ def _handle_exceptions(self):
...
+        passed_old_terms, passed_new_terms = [], []
+        old_term_collector = {}
+        new_term_collector = {}
+
+        #gather the old system bond force terms into a dict
+        for term_idx in old_system_bond_force.getNumBonds():
+            p1, p2, r0, k = old_system_bond_force.getBondParameters(term_idx) #grab the parameters
+            hybrid_p1, hybrid_p2 = self._old_to_hybrid_map[p1], self._old_to_hybrid_map[p2] #make hybrid indices
+            sorted_list = sorted([hybrid_p1, hybrid_p2]) #sort the indices
+            assert not sorted_list in passed_old_terms, f"this bond already exists"
```
zhang-ivy, line 2763: Nothing is being added to `passed_old_terms`, so this assert statement isn't meaningful

perses/annihilation/relative.py
```diff
@@ -2541,3 +2537,269 @@ def _handle_exceptions(self):
...
+            old_term_collector[sorted_list] = [r0, k]
+
+        # repeat for the new system bond force
+        for term_idx in new_system_bond_force.getNumBonds():
+            p1, p2, r0, k = new_system_bond_force.getBondParameters(term_idx)
+            hybrid_p1, hybrid_p2 = self._new_to_hybrid_map[p1], self._new_to_hybrid_map[p2]
+            sorted_list = sorted([hybrid_p1, hybrid_p2])
+            assert not sorted_list in passed_new_terms, f"this bond already exists"
```
zhang-ivy, line 2771: Nothing is being added to `passed_new_terms`, so this assert statement isn't meaningful

jchodera:
@dominicrufa : Instead of simply subclassing and overriding tons of behavior, does it make sense to refactor this to have a shared base class and two subclasses, one for reaction field and one for PME? Or could there be one class that supports both?

Are the simplifying ideas that arise from _only_ supporting reaction field such that these should be two distinct classes because the approaches are vastly different? Or does the common base class make sense here?

perses/annihilation/relative.py
```diff
@@ -2692,39 +2654,197 @@ def __init__(self,
...
+            template[0] = 1
+            for idx, _set in enumerate(self._scale_regions):
+                if particles.intersection(_set) != {}: # at least one of the particles is in the idx_th scale region
+                    #assert that it does not intersect with any other scale region...
+                    further_indices = range(idx+1, self._num_scale_regions)
+                    intersections = [particles.intersection(self._scale_regions[q]) for q in further_indices]
+                    for region_label, val in zip(further_indices, intersections):
+                        assert val == {}, f"term containing particles {particles} intersects with scale regions {idx} and {region_label}"
+                    if particles.issubset(_set): #then this term is wholly in the scale region
+                        template[idx+1] = 1
```
zhang-ivy, line 2830: this should be `template[2*idx + 1]

perses/annihilation/relative.py
```diff
@@ -2692,39 +2654,197 @@ def __init__(self,
...
+                        template[0] = 0
+                    else: #it is interscale region
+                        template[idx+2] = 1
```
zhang-ivy, line 2833: this should be `template[2*idx + 2]

perses/annihilation/relative.py
```diff
@@ -2692,39 +2654,197 @@ def __init__(self,
         # Generate the topology representation
         self._hybrid_topology = self._create_topology()
 
-        # define the scale and nonscale region
-        self._make_scale_regions(scale_region)
+    def _handle_scale_regions(self, scale_regions):
+        num_scale_regions = None
```
zhang-ivy, line 2658: `num_scale_regions` does not need to be initialized here. its not used in the first if statement and in the else statement its redefined

perses/annihilation/relative.py
```diff
@@ -2692,39 +2654,197 @@ def __init__(self,
...
+            for entry in scale_regions:
+                assert type(entry) == list
+
+            #the last bit is to ensure that the scale regions are disjoint
+            num_scaled_particles = sum([len(i) for i in scale_regions])
+            import itertools
+            num_combined_scaled_particles = len(list(itertools.chain.from_iterable(scale_regions)))
+            assert num_scaled_particles == num_combined_scaled_particles, f"each scale region must be disjoint"
+            self._scale_regions = [set(i) for i in scale_regions]
+
```
zhang-ivy, line 2674: note that you never set `self._num_scale_regions` in this else statement. 
I would recommend not having both properties `self._num_scale_regions` and `self._scale_regions`. You can just use `len(self._scale_regions)` to retrieve the `num_scale_regions`. Or, you could define a helper function `get_num_scale_regions()` that retrieves the number of scale regions based on `self._scale_regions`. It's not ideal to have to update two different properties related to the same variable every time the variable changes.

perses/annihilation/relative.py
```diff
@@ -2692,39 +2654,197 @@ def __init__(self,
...
+        is_single_alch_region = True if type(self._topology_proposal._core_new_to_old_atom_map) == dict else False
+        if is_single_alch_region:
+            assert type(self._topology_proposal._unique_old_atoms) == list
+            assert type(self._topology_proposal._unique_new_atoms) == list
+            self._num_alchemical_regions = 1
+        else:
+            len_list =len(self._topology_proposal._core_new_to_old_atom_map)
+            for entry in [self._topology_proposal._core_new_to_old_atom_map, self._topology_proposal._unique_old_atoms, self._topology_proposal._unique_new_atoms]:
+                assert entry == list
+                assert len(entry) == len_list
```
zhang-ivy, line 2704: are you sure that you want every element in the loop to have the same length as the `core_new_to_old_atom_map`?

perses/annihilation/relative.py
```diff
@@ -2692,39 +2654,197 @@ def __init__(self,
...
+            self._num_alchemical_regions = len_list
+
+        atom_classes = {
+                        'unique_old_atoms' : {i: set() for i in range(len_list)},
```
zhang-ivy, line 2708: `len_list` is not defined if `is_single_alch_region` == True. Change `len_list` to `self._num_alchemical_regions`

perses/annihilation/relative.py
```diff
@@ -2692,39 +2654,197 @@ def __init__(self,
...
+                    hybrid_idx = self._new_to_hybrid_map[atom_idx]
+                    atom_classes['unique_new_atoms'][main_idx].add(hybrid_idx)
+
+        # The core atoms:
+        if is_single_alch_region:
+            core_atoms = set()
+            for new_idx, old_idx in self._topology_proposal._core_new_to_old_atom_map.items():
+                new_to_hybrid_idx, old_to_hybrid_index = self._new_to_hybrid_map[new_idx], self._old_to_hybrid_map[old_idx]
+                assert new_to_hybrid_idx == old_to_hybrid_index, f"there is a -to_hybrid naming collision in topology proposal core atom map: {self._topology_proposal._core_new_to_old_atom_map}"
+                core_atoms.append(new_to_hybrid_idx)
```
zhang-ivy, line 2742: you cannot append to a set()

dominicrufa:
https://gist.github.com/dominicrufa/fbc4e0e7473dc676af179427d91967a6; it looks like my first attempt at rxn field is not differentiable...i am using the SW scheme given by [this paper](https://pubs.rsc.org/am/content/articlehtml/2020/cp/d0cp03835k)

mikemhenry:
@zhang-ivy create tests

dominicrufa:
> Should the maximum 4th dimension "lifting" be `r_cutoff`, or should this be a user-specified parameter?
> Perhaps the parameter could be dimensionless between 0 and 1, and scale `r_cutoff`?

this is now a user-defined parameter (with a default) given by `w_RF_scale` and `w_sterics_scale`

> Would it be possible to document new_bool_string and old_bool_string here in the note somewhere?

see docstrings under `RxnHybridTopologyFactory` class. also, we use the same logic in the valence terms, too.

> Could we add a bit more documentation about how each of these blocks works?

added a bit more description

-addressed in [ 2c3d670](https://github.com/choderalab/perses/pull/785/commits/2c3d6707856d246732abd4eadc0dff905b0c5708). 

TODOs/mentions:
- still need to handle nonbonded exceptions and test. 
- the nonbonded force expressions now have several user-defined parameters. the defaults are not necessarily optimal.

jchodera:
@dominicrufa : Feel free to tag me in if you want me to try some testing (hopefully with info on what tests you'd like me to run).

mikemhenry:
before doing too much work on this PR, be sure to merge in master/resolve the conflicts so that the CI can run and pickup on any errors (or at least run your tests locally like this:  `GITHUB_ACTIONS=true CUDA_VISIBLE_DEVICES= pytest -v -a "not advanced"`

zhang-ivy:
>before doing too much work on this PR, be sure to merge in master/resolve the conflicts so that the CI can run and pickup on any errors (or at least run your tests locally like this: GITHUB_ACTIONS=true CUDA_VISIBLE_DEVICES= pytest -v -a "not advanced"
I actually don't think we should merge master in yet, there is something broken (which I caught in the #770). This should be a high priority item, so I'll try to look into it today -- if you have time as well, that would be great

mikemhenry:
fyi the tests are falining from what looks like a mangled merge:
```
>           from perses.app.setup_relative_calculation import run_setup
E             File "/home/runner/work/perses/perses/perses/app/setup_relative_calculation.py", line 533
E               <<<<<<< HEAD
E               ^
E           SyntaxError: invalid syntax
```

peastman:
Following up on our meeting, I looked through the code to see if there were ways it could be simplified.  It's complicated, but largely that's because what it's doing is complicated.  You're simultaneously modifying the potential in three ways: to alchemically modify atoms, to create soft-core interactions, and to scale interactions to improve sampling.  And all of those are done independently to different groups of atoms, or in some cases to multiple groups of atoms.  The code for doing all of that will necessarily be a bit complicated.

More comments in the code would help to make it easier to understand.  The docstring at the top of RxnHybridTopologyFactory is great.  But once you get into the details of the code, there's a lot less.  For example `_transcribe_nonbondeds()` has no docstring  at all.  And while there are some comments scattered through it, a lot of them don't mean much unless you already understand how the code works.  For example,

https://github.com/choderalab/perses/blob/66f341a478b5e032e793a21ee9e86f1edf177c92/perses/annihilation/relative.py#L3806

Just changing that to

```python
# Loop over particles in the old system and add them to the nonbonded forces
```

would make the code easier to understand.

I think the logic for assembling per-particle parameters could be simplified.  Consider this code:

https://github.com/choderalab/perses/blob/66f341a478b5e032e793a21ee9e86f1edf177c92/perses/annihilation/relative.py#L3811-L3831

Several things make this more confusing than it needs to be.

- `string_identifier` isn't a very informative name if you don't already know what it is.  How about `alchemical_type`?
- It looks at `string_identifier` to try to figure out whether the particle exists in the new system, and has a comment to explain that.  Why not just check directly?

```python
if hybrid_idx in self._hybrid_to_new_map:
```

- There's logic scattered through here for setting `alchemical_type_id`.  You have to read through it carefully before you realize there's just a simple one-to-one correspondence between values of `string_identifier` and `alchemical_type_id`.  Not only that, but the exact same logic gets repeated in a different form a bit further down:

https://github.com/choderalab/perses/blob/66f341a478b5e032e793a21ee9e86f1edf177c92/perses/annihilation/relative.py#L3882-L3888

So why not centralize it?  You could replace that with a single lookup in a dict.  Or it could be returned directly by `get_alch_identifier()`.  Or even better, you could combine the identification of all the different "ids" into a single method:

```python
electrostatics_force.addParticle(self.get_id_parameters(hybrid_idx) + [charge_old, charge_new])
```

The definition of the energy expressions is done in a confusing way:

https://github.com/choderalab/perses/blob/66f341a478b5e032e793a21ee9e86f1edf177c92/perses/annihilation/relative.py#L2723-L2764

I assume the reason was to avoid repeating parts of the expression that are shared between the two versions?  But this means that in order to figure out what the expression for exceptions is, you need to carefully compare two blocks of code, and count lines to figure out exactly what is being replaced.  I would probably have just written it as

```python
_default_steric_expression = """
    U_sterics;
    U_sterics = 4 * epsilon * x * (x - 1.0);
    x = (sigma / r_eff)^6;
    sigma = (sigma1 + sigma2) / 2;
    sigma1 = (sigma_old1 * {old_bool_string1}) + (sigma_new1 * {new_bool_string1});
    sigma2 = (sigma_old2 * {old_bool_string2}) + (sigma_new2 * {new_bool_string2});
    epsilon = sqrt(epsilon1 * epsilon2);
    epsilon1 = ((epsilon_old1 * {old_bool_string1}) + (epsilon_new1 * {new_bool_string1})) * {scale_bool_string1};
    epsilon2 = ((epsilon_old2 * {old_bool_string2}) + (epsilon_new2 * {new_bool_string2})) * {scale_bool_string2};
    r_eff = sqrt(r^2 + w^2);
    w = step(unique_new1 + unique_new2 + unique_old1 + unique_old2 - 0.1) * r_cutoff * (1. - {old_bool_string1} - {new_bool_string1}) * {w_scale};
    r_cutoff = {r_cutoff};
    w_scale = {w_scale};"""

_default_steric_exception_expression = """
    U_sterics * {scale_bool_string};
    U_sterics = 4 * epsilon * x * (x - 1.0);
    x = (sigma / r_eff)^6;
    sigma = select(1 - environment_region, sigma_old * {old_bool_string} + sigma_new * {new_bool_string}, sigma_old);
    epsilon = select(1 - environment_region, epsilon_old * {old_bool_string} + epsilon_new * {new_bool_string}, epsilon_old);
    r_eff = sqrt(r^2 + w^2);
    w = step(unique_old + unique_new - 0.1) * r_cutoff * (1. - {old_bool_string} - {new_bool_string}) * {w_scale};
    r_cutoff = {r_cutoff};
    w_scale = {w_scale};"""
```

That makes it easier for someone reading the code to see exactly what the expressions are.

Do you really need as many parameters as you have?  For example,

> sigma1 = (sigma_old1 * {old_bool_string1}) + (sigma_new1 * {new_bool_string1});

Don't the two parameters need to add to 1?  If so, you could eliminate one and just make that

> sigma1 = (sigma_old1 * {old_bool_string1}) + (sigma_new1 * (1-{old_bool_string1}));

This is purely a matter of style, so feel free to ignore it if you want, but I found these variable names a bit awkward.  I prefer to name variables based on what they *mean* rather than what they *are*.  For example, `is_old1` rather than `old_bool_string1`.  Also, are those names even correct?  Isn't it a floating point number between 0 and 1, not a boolean?

You have a few lines like

https://github.com/choderalab/perses/blob/66f341a478b5e032e793a21ee9e86f1edf177c92/perses/annihilation/relative.py#L3905

Although it's very unlikely to happen, that code could potentially produce wrong results.  A correct check would be

```python
if all([v.value_in_unit_system(unit.md_unit_system) == 0.0 for v in (chargeProd_old, chargeProd_new, epsilon_old, epsilon_new)]):
```

jchodera:
Thanks for the detailed feedback, @peastman!

> You're simultaneously modifying the potential in three ways: to alchemically modify atoms, to create soft-core interactions, and to scale interactions to improve sampling.

@zhang-ivy @dominicrufa : Would there be any significant simplification obtained by unifying the treatment of these three groups? All three can be considered special cases of a single general case:
1. **Atoms being created/deleted** interpolate parameters between zero and nonzero parameters _and_ "lift" the atom into the fourth dimension for the zero parameter state
2. **Atoms being interpolated** interpolate parameters between two nonzero parameter sets and keep the fourth dimension fixed at zero
3. **Atoms being RESTed** interpolate parameters between two nonzero parameter sets and keep the fourth dimension fixed at zero

That would suggest that each atom would simply need to have defined:
* alchemical group index (presuming they are disjoint)
* old epsilon, old sigma, old charge, old w
* new epsilon, new sigma, new charge, new w

In fact, since we would only ever want to have a nonzero `w` for the case where epsilon=0, we could potentially just define `w_max` and tie the logic for w to approach `w_max` as epsilon approaches 0. But this might prohibit interesting experiments, like per-atom definitions of `w`.

There's not an easy way to select an parameter out of a vector of global parameters by alchemical group index, so it may be necessary to replace "alchemical group index" with the array of alchemical parameters one might want to use, one or two per group, as you do now. But they would be unified in type.

perses/annihilation/relative.py
```diff
@@ -2629,3 +2629,1504 @@ def _handle_exceptions(self):
...
+
+                             #define reaction field u_RF
+                             "u_RF = (1 / r_eff) + ((eps_RF - 1) * (r_eff^2)) / ((r_cutoff^3) * (1 + (2 * eps_RF)));",
+
+                             #r_eff
+                             "r_eff = sqrt(r^2 + w^2);",
+
+                             #4th dimension:
+                             #"w = step(unique_new1 + unique_new2 + unique_old1 + unique_old2 - 0.1) * r_cutoff * ({old_bool_string1} - {new_bool_string1}) * {w_scale};",
+                             "w = is_unique_old * r_cutoff * w_scale * (1 - lambda_0_electrostatics_old) + is_unique_new * w_scale * (1 - lambda_0_electrostatics_new);",
```
jchodera, line 2699: Why does `r_cutoff * w_scale` appear in the first term but only `w_scale` in the second term?

perses/annihilation/relative.py
```diff
@@ -2629,3 +2629,1504 @@ def _handle_exceptions(self):
...
+    # predefined RF parameters (fed to the `expression` via `format`): [scale_bool_string, old_bool_string, new_bool_string, w_scale, switching_ratio, eps_RF, r_cutoff]
+    # RF per_particle_parameters: [scale_bools, old/new_bool_string, unique_old, unique_new, core]
+
+    #electrostatic exception
+    _default_RF_exception_expr_list = copy.deepcopy(_default_RF_expr_list)
+    _default_RF_exception_expr_list[0] = "U_electrostatics * {scale_bool_string};"
+    _default_RF_exception_expr_list[5] = "chargeProd = select(1 - environment_region, chargeProd_old * {old_bool_string} + chargeProd_new * {new_bool_string}, chargeProd_old);"
+    _default_RF_exception_expr_list[6] = ''
+    _default_RF_exception_expr_list[7] = ''
+    _default_RF_exception_expr_list[10] = "w = (is_unique_old * r_cutoff * w_scale * (1 - lambda_0_electrostatics_exceptions_old)) + (is_unique_new * w_scale * (1 - lambda_0_electrostatics_exceptions_new));"
```
jchodera, line 2720: Again, why does `r_cutoff * w_scale` appear in the first term, but only `w_scale` in the second term?

perses/annihilation/relative.py
```diff
@@ -2629,3 +2629,1504 @@ def _handle_exceptions(self):
...
+                                  "epsilon = sqrt(epsilon1 * epsilon2);",
+                                  "epsilon1 = ((epsilon_old1 * {old_bool_string1}) + (epsilon_new1 * {new_bool_string1})) * {scale_bool_string1};",
+                                  "epsilon2 = ((epsilon_old2 * {old_bool_string2}) + (epsilon_new2 * {new_bool_string2})) * {scale_bool_string2};",
+
+                                   #r_eff
+                                   "r_eff = sqrt(r^2 + w^2);",
+                                   
+                                   #4th dimension:
+                                   #"w = step(unique_new1 + unique_new2 + unique_old1 + unique_old2 - 0.1) * r_cutoff * ({old_bool_string1} - {new_bool_string1}) * {w_scale};",
+                                   "w = is_unique_old * r_cutoff * w_scale * (1 - lambda_0_sterics_old) + is_unique_new * w_scale * (1 - lambda_0_sterics_new);",
```
jchodera, line 2747: Here as well: `r_cutoff * w_scale` appears in the first term, but only `w_scale` in the second term.

perses/annihilation/relative.py
```diff
@@ -2629,3 +2629,1504 @@ def _handle_exceptions(self):
...
+    #steric exception
+    _default_steric_exception_expr_list = copy.deepcopy(_default_steric_expr_list)
+    _default_steric_exception_expr_list[0] = "U_sterics * {scale_bool_string};"
+    _default_steric_exception_expr_list[3] = "sigma = select(1 - environment_region, sigma_old * {old_bool_string} + sigma_new * {new_bool_string}, sigma_old);"
+    _default_steric_exception_expr_list[4] = ''
+    _default_steric_exception_expr_list[5] = ''
+    _default_steric_exception_expr_list[6] = "epsilon = select(1 - environment_region, epsilon_old * {old_bool_string} + epsilon_new * {new_bool_string}, epsilon_old);"
+    _default_steric_exception_expr_list[7] = ''
+    _default_steric_exception_expr_list[8] = ''
+    _default_steric_exception_expr_list[10] = "w = (is_unique_old * r_cutoff * w_scale * (1 - lambda_0_sterics_exceptions_old)) + (is_unique_new * w_scale * (1 - lambda_0_sterics_exceptions_new));"
```
jchodera, line 2767: And again: `r_cutoff * w_scale` appears in the first term, but only `w_scale` in the second term.

zhang-ivy:
Thanks for both of your feedback. I've spent some time carefully going through your suggestions and pushed the changes that I thought were relatively low effort yet high impact for improving clarity. 
In summary: 
- Added more comments, docstrings (if missing)
- Explicitly copied over the the custom exceptions expressions s.t. its easier to read
- Renamed variables/functions such as `old_bool_string`  to `old_scalar` and `render_bool_string` to `render_scalar_string` to better reflect what they mean
- Renamed `string_identifier` to `atom_class` to better reflect what this variable means
- Adapted how we check if charge/sigma/epsilon == 0 per Peter's recommendation
- Fixed asymmetry in `w_scale` as John pointed out
- Edited the formatting of comments

I did not make any major changes to the custom expressions, as we use similar logic in the custom valence forces, so any changes to the custom nonbonded/bond force logic should be mimicked there too.

Please let me know if you have any additional feedback (@peastman , @jchodera , @dominicrufa), otherwise I will proceed with using the implementation in this PR to draft the implementation with PME tomorrow!

jchodera:
This will be deprecated in favor of #848, but keeping this open for now until we can borrow RF implementation.

mikemhenry:
Going to close since #848 has been merged

ijpulidos:
Just a note for future reference. We never really fully implemented reaction field in #848 .

PR #776, opened 2021-03-05
fix nightly builds

mikemhenry:
Namespace changes are coming in openmm, they are backwards compatible if we modify our import slightly.

https://github.com/openmm/openmm/issues/3022

PR #786, opened 2021-04-26
Feat/lazy load imports

mikemhenry:
Moving all of our `openeye` imports inside the methods/functions that need them. I didn't do this to the examples folder but everything else has been moved except for `perses/rjmc/topology_proposal.py` which uses `openeye` extensively for some global constants. I will have to think about the best way to edit that file, thoughts @jaimergp @jchodera for that one?

### DIFF
--- a/perses/rjmc/topology_proposal.py
+++ b/perses/rjmc/topology_proposal.py
@@ -8,7 +8,6 @@ import copy
 import logging
 import itertools
 import os
-import openeye.oechem as oechem
 import numpy as np
 import networkx as nx
 import openmoltools.forcefield_generators as forcefield_generators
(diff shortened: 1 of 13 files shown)


jaimergp:
One approach would be encapsulating those constants in a "constant factory function" that gets called when needed. Another approach would be guarding those constants in a `try / except` clause, since they are key to the operational state of the module it doesn't make sense to err out gracefully, but you can always add a LARGE warning. At that point I don't know how beneficial the refactor would be because it'd make the module unusable without OE, as it is right now 🤷

mikemhenry:
I went with the try/except idea. Just for reference, perses feedstock is here:
https://github.com/conda-forge/perses-feedstock

perses/rjmc/topology_proposal.py
```diff
@@ -24,32 +23,36 @@
...
+    # bonds are same as default, require them to match in bond order
+    STRONG_ATOM_EXPRESSION = oechem.OEExprOpts_Hybridization | oechem.OEExprOpts_HvyDegree | oechem.OEExprOpts_DefaultAtoms
+    STRONG_BOND_EXPRESSION = oechem.OEExprOpts_DefaultBonds
+
+    # specific to proteins
+    # PROTEIN_ATOM_EXPRESSION = oechem.OEExprOpts_Hybridization | oechem.OEExprOpts_EqAromatic
+    # PROTEIN_BOND_EXPRESSION = oechem.OEExprOpts_Aromaticity
+    PROTEIN_ATOM_EXPRESSION = DEFAULT_ATOM_EXPRESSION
+    PROTEIN_BOND_EXPRESSION = DEFAULT_BOND_EXPRESSION
+except(ImportError):
```
jaimergp, line 54: I don't think parentheses are needed here. Also, you might need to check for OE license errors.

jchodera:
> I didn't do this to the examples folder but everything else has been moved except for perses/rjmc/topology_proposal.py which uses openeye extensively for some global constants. I will have to think about the best way to edit that file, thoughts @jaimergp @jchodera for that one?

We shouldn't be using global constants. If anything, these constants should be default values for class/object fields. Let's just move them inside the class that needs them.

mikemhenry:
> We shouldn't be using global constants for these defaults. If anything, these constants should be default values for class/object fields. Let's just move them inside the class that needs them.

Can do! Are those constants "expensive" to calculate? A few different classes need them which is why I didn't copy them into the spots they needed to be.

mikemhenry:
I moved all the global constants in that file to where they were used, so now this should be good to merge!

jchodera (approved):
This works great for now! Thanks!

jchodera:
We can deal with the nightly failures separately in https://github.com/choderalab/perses/pull/776, so merging this now.

PR #794, opened 2021-05-26
run tests in parallel

mikemhenry:
just a quick test to see if this speeds up running our tests

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -91,7 +91,7 @@ jobs:
           export OE_LICENSE="$HOME/oe_license.txt"
           export TRAVIS=true
           pushd .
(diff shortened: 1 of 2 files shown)


mikemhenry:
Looks like it shaves a few minutes off of the tests and they still all pass!

jchodera:
Nice! As long as it doesn't make it crazy difficult to track down test failures, that's great!
Even if it does make the output confusing, if we can *just re-run the failed tests serially at the end*, that would still be an immense help.

mikemhenry:
Locally the speed-up is much more dramatic (I've got 16 cores)  since github only gives us two cores when running on their servers, but it saves some time! In my experience the output doesn't end up too messed up since pytest is capturing the output per worker and dumps it also per worker, so it it doesn't look like mpi spaghetti

Issue #754, opened 2020-11-26
`atom_map` doesn't do anything in the relative FEP setup


hannahbrucemacdonald:
https://github.com/choderalab/perses/blob/33a6b4c41da2414c727c7980cb400d6402b24f3d/perses/app/relative_setup.py#L60

This variable doesn't do anything

mikemhenry:
I was just about to make an issue after I noticed the same thing!

PR #796, opened 2021-05-26
removed atom_map from FEP constructor and updated tests

mikemhenry:
fixes issue #754


jchodera (approved):
Yeah, I think it makes sense to delete this non-working "feature" to not cause confusion, and then to add an issue feature-request to make sure we incorporate an appropriate API for manually specifying atom maps between small molecules into the new API design.

Issue #759, opened 2021-02-01
Update Install Instructions 


mikemhenry:
It looks like the `conda-forge` channel is necessary for installation, without it, we get this really confusing error from conda:

```
➜ conda install -c omnia perses
Collecting package metadata (current_repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.
Solving environment: failed with repodata from current_repodata.json, will retry with next repodata source.
Collecting package metadata (repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.
Solving environment: | 
Found conflicts! Looking for incompatible packages.
This can take several minutes.  Press CTRL-C to abort.
failed

UnsatisfiableError: The following specifications were found to be incompatible with each other:

Output in format: Requested package -> Available versions
```

(I've never seen this error fail so hard that it doesn't even say which package it can't find)

When I add `-c conda-forge` it works:

<details>
  <summary>Click to expand!</summary>

```
➜ conda install -c omnia -c conda-forge perses                                                                                                                                       
Collecting package metadata (current_repodata.json): done
Solving environment: done

## Package Plan ##

  environment location: /home/mmh/miniconda3/envs/perses

  added / updated specs:
    - perses


The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    alabaster-0.7.12           |             py_0          15 KB  conda-forge
    ambertools-18.0            |                0       114.9 MB  omnia
    argon2-cffi-20.1.0         |   py37h4abf009_2          47 KB  conda-forge
    astunparse-1.6.3           |     pyhd8ed1ab_0          15 KB  conda-forge
    async_generator-1.10       |             py_0          18 KB  conda-forge
    attrs-20.3.0               |     pyhd3deb0d_0          41 KB  conda-forge
    babel-2.9.0                |     pyhd3deb0d_0         6.2 MB  conda-forge
    backports-1.0              |             py_2           4 KB  conda-forge
    backports.functools_lru_cache-1.6.1|             py_0           8 KB  conda-forge
    blas-1.0                   |              mkl           6 KB
    bleach-3.2.3               |     pyh44b312d_0         111 KB  conda-forge
    blosc-1.20.1               |       hd408876_0          71 KB
    bokeh-1.4.0                |   py37hc8dfbb8_1        13.5 MB  conda-forge
    brotlipy-0.7.0             |py37hb5d75c8_1001         346 KB  conda-forge
[...]

PR #797, opened 2021-05-26
Update installation instructions

mikemhenry:
Fixes issue #759﻿

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -22,7 +22,7 @@ You can install the latest stable release build of perses via the ``conda`` pack
 
 .. code-block:: none
 
-   $ conda config --add channels conda-forge
+   $ conda config --add channels conda-forge openeye
    $ conda install perses
 
 This version is recommended for all users not actively developing new algorithms for alchemical free energy calculations.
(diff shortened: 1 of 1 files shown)


docs/installation.rst
```diff
@@ -43,7 +22,7 @@ You can install the latest stable release build of perses via the ``conda`` pack
 
 .. code-block:: none
 
-   $ conda config --add channels omnia
+   $ conda config --add channels conda-forge
```
jchodera, line 25: We also need to mention the openeye-toolkits installation step if we don't already!
```bash
conda install -c openeye openeye-toolkits
```
Or, if we want folks to be able to `conda update` everything, including the toolkit, we should add `openeye` to the `conda config --add channels` and then add `conda install --yes openeye-toolkits` somewhere.

jchodera (changes requested):
Thanks for updating this!

docs/installation.rst
mikemhenry: good catch! I added `openeye` to the list of channels to add

docs/installation.rst
```diff
@@ -43,7 +22,7 @@ You can install the latest stable release build of perses via the ``conda`` pack
 
 .. code-block:: none
 
-   $ conda config --add channels omnia
+   $ conda config --add channels conda-forge openeye
```
jchodera, line 25: I think we need to still explicitly advise the user install `openeye-toolkits` if we don't currently do so.

mikemhenry: You are correct!

PR #804, opened 2021-06-07
Quick fix for issue 802

mikemhenry:
We still need to figure out the root cause but this PR sets the default temperature back to 300 K (which reverts PR #783 ) and also adds a quick test that will help catch regressions.


mikemhenry:
Need to commit the PDB file as well...

PR #808, opened 2021-06-17
Allow global setting of random number seed + better RepartitionedHybridTopologyFactory test

mikemhenry:
Right now test_RepartitionedHybridTopologyFactory_energies fails sometimes depending on which amino is selected and the order that torsions are mapped.  You can test this by running `while CUDA_VISIBLE_DEVICES= pytest -k test_RepartitionedHybridTopologyFactory_energies; do echo "works"; done` and eventually the test will fail. I looked at the output and traced the difference between a test passing and a test failing to atom index proposal order, which is random and we currently don't support setting the seed. 

There also is some randomness from the amino that is selected to mutate. I think instead of mutating a random amino for testing, we should mutate all of them. I will check to see how much time this adds to the test. 
 

Some ideas:

Set random number seed in a util file, and import that generator in tests that need it instead of doing ` rng = np.random.RandomState(42)` everywhere


zhang-ivy:
Interesting.. so none of the other test functions in `test_relative.py` fail (e.g. `test_HybridTopologyFactory_energies()`)? is it just `test_RepartitionedHybridTopologyFactory()`?

Also, `test_RepartitionedHybridTopologyFactory()` has 3 cases: alanine dipeptide in vacuum, alanine dipeptide in solvent, and an 8-mer peptide in solvent. Which one(s) are failing?

jchodera:
May also help with #815

Issue #805, opened 2021-06-07
When using geometry-based mapping, de-map problematic transformations


jchodera:
Currently, the geometry-based mapping can cause some problematic transformation mappings to be used, e.g.:
![image](https://user-images.githubusercontent.com/3656088/121083657-b4b59200-c7ad-11eb-90cf-9b4d794189fa.png)
We should make sure to de-map these.

Synergizes with #792, which can catch these issues more generally.

jchodera:
Here's another problematic map, which maps only part of the ring between initial and final molecules:
![image](https://user-images.githubusercontent.com/3656088/122515809-d2130900-cfc2-11eb-9ca7-5751b4aa4659.png)
This leads to problematic overlap in solvent:
![image](https://user-images.githubusercontent.com/3656088/122515882-eb1bba00-cfc2-11eb-926c-63b2d75c028d.png)

Here is a ZIP file containing the two molecules involved in the transformation, correctly overlaid:
[RUN344-transformation.zip](https://github.com/choderalab/perses/files/6675178/RUN344-transformation.zip)

Despite the correct overlay, where all atoms (except the H -> F) are on top of each other, an incorrect atom mapping is generated by perses 0.9.0. All atoms except the H -> F should be mapped on top of each other.

@dominicrufa : Any insight into what's going on here?

dominicrufa:
if you open the molecules in pymol and switch between them, you'll see that
the double ring structure not being mapped appropriately is actually
reoriented ever so slightly.  this is a consequence of the relative and
absolute tolerances for the deterministic atom map: here
<https://github.com/choderalab/perses/blob/d215e7a440a6c92ae585c5fea7cb34e654fca512/perses/rjmc/topology_proposal.py#L311>.
one can change the tolerance or modify the sdf-generating scheme so that
atom positions of scaffold atoms are not reoriented.

jchodera:
Thanks! 

@dominicrufa : Since there are no docstrings, could you help me figure out
* What units are `rtol` and `atol` in?
* What do `rtol` and `atol` mean?
* Are these propagated in through any arguments path so they could be set in the YAML file or something?

dominicrufa:
it looks like rtol and atol are the same arguments as in:
https://numpy.org/doc/stable/reference/generated/numpy.allclose.html, and
are in units w.r.t. whatever units
[...]

Issue #792, opened 2021-05-25
Check factorization of partition function


mikemhenry:
We can use the code from `timemachine` as a guide for our implementation.

https://github.com/proteneer/timemachine/blob/c08bfda717d535b0ae9c0ec71f661138819192fe/fe/topology.py#L452-L529

@jchodera if you let me know where we should add this I can take a stab at it.

jchodera:
Fixed by #809.

jchodera:
We determined the checks in https://github.com/proteneer/timemachine/blob/c08bfda717d535b0ae9c0ec71f661138819192fe/fe/topology.py#L452-L529 are too restrictive in that they only permit a _single_ substituent to be alchemically modified at a time, whereas perses permits multiple substituents to be simultaneously changed. Our solution in #809 is therefore more general.

PR #809, opened 2021-06-19
Overhaul atom mapping

### LINKED ISSUE #826: Sporadic unspecified stereochemistry issue with generate_atom_mapping_from_positions
I've run into an issue where occasionally we encounter an OpenEye toolkit unspecified stereochemistry issue with the new `AtomMapping.generate_atom_mapping_from_positions()` functionality in #809, in this case from `ADA-UCB-6c2cb422-1_1` to `MIC-UNK-b12b7f76-2_1 ` (FAH Sprint 10, `RUN1083`, job id 1084):
```
2021-07-26 22:18:38,073:(0.18s):openff.toolkit.typing.engines.smirnoff.parameters:90 angles added (0 skipped due to constraints)
2021-07-26 22:18:53,425:(15.35s):proposal_generator:determining atom map between old and new molecules...
2021-07-26 22:18:53,426:(0.00s):proposal_generator:the atom map is not specified; proceeding to generate an atom map...
  File "/lila/home/chodera/miniconda/envs/perses-fah-0.9.1/bin/perses-fah", line 33, in <module>
    sys.exit(load_entry_point('perses==0.9.0+46.g32c8652.dirty', 'console_scripts', 'perses-fah')())
  File "/lila/home/chodera/miniconda/envs/perses-fah-0.9.1/lib/python3.7/site-packages/perses-0.9.0+46.g32c8652.dirty-py3.7.egg/perses/app/fah_generator.py", line 498, in run
    run_neq_fah_setup(**setup_options)
  File "/lila/home/chodera/miniconda/envs/perses-fah-0.9.1/lib/python3.7/site-packages/perses-0.9.0+46.g32c8652.dirty-py3.7.egg/perses/app/fah_generator.py", line 370, in run_neq_fah_setup
    setup_dict = run_setup(setup_options, serialize_systems=False, build_samplers=False)
  File "/lila/home/chodera/miniconda/envs/perses-fah-0.9.1/lib/python3.7/site-packages/perses-0.9.0+46.g32c8652.dirty-py3.7.egg/perses/app/setup_relative_calculation.py", line 461, in run_setup
    use_given_geometries=use_given_geometries, given_geometries_tolerance=given_geometries_tolerance)
  File "/lila/home/chodera/miniconda/envs/perses-fah-0.9.1/lib/python3.7/site-packages/perses-0.9.0+46.g32c8652.dirty-py3.7.egg/perses/app/relative_setup.py", line 382, in __init__
    current_mol_id=0, proposed_mol_id=1)
  File "/lila/home/chodera/miniconda/envs/perses-fah-0.9.1/lib/python3.7/site-packages/perses-0.9.0+46.g32c8652.dirty-py3.7.egg/perses/rjmc/topology_proposal.py", line 2313, in propose
    atom_mapping = atom_mapper.generate_atom_mapping_from_positions(self.current_molecule, self.proposed_molecule)
  File "/lila/home/chodera/miniconda/envs/perses-fah-0.9.1/lib/python3.7/site-packages/perses-0.9.0+46.g32c8652.dirty-py3.7.egg/perses/rjmc/atom_mapping.py", line 1205, in generate_atom_mapping_from_positions
    new_mol = Molecule(new_mol)
  File "/lila/home/chodera/miniconda/envs/perses-fah-0.9.1/lib/python3.7/site-packages/openff/toolkit/topology/molecule.py", line 5407, in __init__
    super(Molecule, self).__init__(*args, **kwargs)
  File "/lila/home/chodera/miniconda/envs/perses-fah-0.9.1/lib/python3.7/site-packages/openff/toolkit/topology/molecule.py", line 1870, in __init__
    _cls=self.__class__,
  File "/lila/home/chodera/miniconda/envs/perses-fah-0.9.1/lib/python3.7/site-packages/openff/toolkit/utils/toolkits.py", line 5826, in call
    raise e
  File "/lila/home/chodera/miniconda/envs/perses-fah-0.9.1/lib/python3.7/site-packages/openff/toolkit/utils/toolkits.py", line 5822, in call
    return method(*args, **kwargs)
  File "/lila/home/chodera/miniconda/envs/perses-fah-0.9.1/lib/python3.7/site-packages/openff/toolkit/utils/toolkits.py", line 692, in from_object
    oemol=obj, allow_undefined_stereo=allow_undefined_stereo, _cls=_cls
  File "/lila/home/chodera/miniconda/envs/perses-fah-0.9.1/lib/python3.7/site-packages/openff/toolkit/utils/toolkits.py", line 1371, in from_openeye
    raise UndefinedStereochemistryError(msg)
openff.toolkit.utils.toolkits.UndefinedStereochemistryError: Unable to make OFFMol from OEMol: OEMol has unspecified stereochemistry. oemol.GetTitle(): MOL
Problematic atoms are:
Atom atomic num: 6, name: C14, idx: 13, aromatic: False, chiral: True with bonds:
bond order: 1, chiral: False to atom atomic num: 6, name: C8, idx: 7, aromatic: False, chiral: False
bond order: 1, chiral: False to atom atomic num: 6, name: C9, idx: 8, aromatic: False, chiral: False
bond order: 1, chiral: False to atom atomic num: 6, name: C11, idx: 10, aromatic: False, chiral: False
bond order: 1, chiral: False to atom atomic num: 1, name: H17, idx: 37, aromatic: False, chiral: False
Atom atomic num: 6, name: C15, idx: 14, aromatic: False, chiral: True with bonds:
bond order: 1, chiral: False to atom atomic num: 6, name: C8, idx: 7, aromatic: False, chiral: False
bond order: 1, chiral: False to atom atomic num: 6, name: C10, idx: 9, aromatic: False, chiral: False
bond order: 1, chiral: False to atom atomic num: 6, name: C12, idx: 11, aromatic: False, chiral: False
bond order: 1, chiral: False to atom atomic num: 1, name: H18, idx: 38, aromatic: False, chiral: False
```

jchodera:
This PR attempts to address some of the issues described in #805

- [x] Refactor `SmallMoleculeSetProposalEngine` to move all user-modifiable attributes from `propose()` to be class attributes. These attributes should impact all proposals, so they make sense to have at the class level, rather than as optional arguments to `propose()`.
- [x] Expose option to modify cutoff used for geometry-derived mappings
- [x] Enable YAML to specify cutoff for geometry-derived mappings
- [x] Check for problematic mappings using scheme in #792
- [x] Clean up problematic mappings (e.g. de-mapping partially mapped rings)
- [x] Add tests for compounds that led to problematic mappings
- [x] Fix #826

@dominicrufa : I've got a start at addressing #804, but could use some pointers on how to add new options to the YAML file to carry over into geometry-derived mappings!


jchodera:
For the geometry-derived mappings, we should be using the largest connected subgraph of matched nodes that does not contain only a subset of any cycles. Cycles should be either contained in their entirety or have all nodes omitted.

We can use an algorithm that first computes the largest connected subset of all matched nodes and then iterates over all cycles, removing any nodes for which the cycle only contains a partial (rather than complete or empty) set of matched nodes.

Alternatively, we could try to match the whole cycle if they share common subgraphs, but we'd need to be more clever about how exactly we do this graph isomorphism "helped" by partial matches.

jchodera:
@zhang-ivy : I'm seeing test failures on protein mutation tests---any idea what this might mean?
```

    def test_protein_counterion_topology_fix_negitive():
        """
        mutate alanine dipeptide into ARG dipeptide and assert that the appropriate number of water indices are identified
        """
        from perses.rjmc.topology_proposal import PolymerProposalEngine
        new_res = 'ARG'
        charge_diff = -1
    
        # Make a vacuum system
        atp, system_generator = generate_atp(phase='vacuum')
    
        # Make a solvated system/topology/positions with modeller
        modeller = app.Modeller(atp.topology, atp.positions)
        modeller.addSolvent(system_generator.forcefield, model='tip3p', padding=9*unit.angstroms, ionicStrength=0.15*unit.molar)
        solvated_topology = modeller.getTopology()
        solvated_positions = modeller.getPositions()
    
        atp.positions = unit.quantity.Quantity(value=np.array([list(atom_pos) for atom_pos in solvated_positions.value_in_unit_system(unit.md_unit_system)]), unit=unit.nanometers)
        atp.topology = solvated_topology
    
        atp.system = system_generator.create_system(atp.topology)
    
        # Make a topology proposal and generate new positions
        top_proposal, new_pos, _, _ = generate_dipeptide_top_pos_sys(topology = atp.topology,
                                       new_res = new_res,
                                       system = atp.system,
                                       positions = atp.positions,
                                       system_generator = system_generator,
                                       conduct_geometry_prop = True,
                                       conduct_htf_prop = False,
                                       validate_energy_bookkeeping=True,
                                       )
    
        # Get the charge difference
        charge_diff_test = PolymerProposalEngine._get_charge_difference(top_proposal._old_topology.residue_topology.name,
                                                                    top_proposal._new_topology.residue_topology.name)
        assert charge_diff_test == charge_diff
    
        # Get the array of water indices (w.r.t. new topology) to turn into ions
>       water_indices = PolymerProposalEngine.get_water_indices(charge_diff = charge_diff_test,
                                                 new_positions = new_pos,
                                                 new_topology = top_proposal._new_topology,
                                                 radius=0.8)

perses/tests/test_topology_proposal.py:868: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
perses/rjmc/topology_proposal.py:577: in get_water_indices
    traj = md.Trajectory(new_positions[np.newaxis, ...], md.Topology.from_openmm(new_topology))
../../../miniconda/envs/perses-dev/lib/python3.8/site-packages/mdtraj/core/trajectory.py:1169: in __init__
    self.xyz = xyz
../../../miniconda/envs/perses-dev/lib/python3.8/site-packages/mdtraj/core/trajectory.py:882: in xyz
    value = ensure_type(value, np.float32, 3, 'xyz', shape=shape,
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

val = Quantity(value=array([[[ 2.00000100e-01,  1.00000000e-01, -1.30000000e-07],
        [ 2.00000100e-01,  2.09000000e-01,...-01,  5.23500000e-01,  9.63099985e-01],
        [ 3.55762500e-02,  6.86000000e-02,  1.78259999e+00]]]), unit=nanometer)
dtype = <class 'numpy.float32'>, ndim = 3, name = 'xyz', length = None, can_be_none = False, shape = (None, 1571, 3), warn_on_cast = False, add_newaxis_on_deficient_ndim = True

    def ensure_type(val, dtype, ndim, name, length=None, can_be_none=False, shape=None,
        warn_on_cast=True, add_newaxis_on_deficient_ndim=False):
        """Typecheck the size, shape and dtype of a numpy array, with optional
        casting.
    
        Parameters
        ----------
        val : {np.ndaraay, None}
            The array to check
        dtype : {nd.dtype, str}
            The dtype you'd like the array to have
        ndim : int
            The number of dimensions you'd like the array to have
        name : str
            name of the array. This is used when throwing exceptions, so that
            we can describe to the user which array is messed up.
        length : int, optional
            How long should the array be?
        can_be_none : bool
            Is ``val == None`` acceptable?
        shape : tuple, optional
            What should be shape of the array be? If the provided tuple has
            Nones in it, those will be semantically interpreted as matching
            any length in that dimension. So, for example, using the shape
            spec ``(None, None, 3)`` will ensure that the last dimension is of
            length three without constraining the first two dimensions
        warn_on_cast : bool, default=True
            Raise a warning when the dtypes don't match and a cast is done.
        add_newaxis_on_deficient_ndim : bool, default=True
            Add a new axis to the beginining of the array if the number of
            dimensions is deficient by one compared to your specification. For
            instance, if you're trying to get out an array of ``ndim == 3``,
            but the user provides an array of ``shape == (10, 10)``, a new axis will
            be created with length 1 in front, so that the return value is of
            shape ``(1, 10, 10)``.
    
        Notes
        -----
        The returned value will always be C-contiguous.
    
        Returns
        -------
        typechecked_val : np.ndarray, None
            If `val=None` and `can_be_none=True`, then this will return None.
            Otherwise, it will return val (or a copy of val). If the dtype wasn't right,
            it'll be casted to the right shape. If the array was not C-contiguous, it'll
            be copied as well.
    
        """
        if can_be_none and val is None:
            return None
    
        if not isinstance(val, np.ndarray):
            if isinstance(val, collections.abc.Iterable):
                # If they give us an iterator, let's try...
                if isinstance(val, collections.abc.Sequence):
                    # sequences are easy. these are like lists and stuff
                    val = np.array(val, dtype=dtype)
                else:
                    # this is a generator...
>                   val = np.array(list(val), dtype=dtype)
E                   ValueError: setting an array element with a sequence.

../../../miniconda/envs/perses-dev/lib/python3.8/site-packages/mdtraj/utils/validation.py:104: ValueError
```

jchodera:
@dominicrufa : I'd like to add atom mapping validation and refinement steps to occur after the [coordinate-derived mapping step](https://github.com/choderalab/perses/pull/809/files#diff-8f7684f72557823abcc54734edc5aaa7b730fb8bfda1b43cda8ab6a511bfc6d6R135-R174). Are there methods already implemented in `AtomMapper` that I should be using to check the validity of mappings or de-map ring systems that are only partially mapped? Or am I better off implementing new methods for this purpose?

zhang-ivy:
@mikemhenry:  I think you re-worked the counterion tests.. did you see the above error John mentioned before you merged? 

I've run the counterion code for my RBD:ACE2 D501Y simulations and it runs fine. It seems like this is error only comes up in the tests. Something is going wrong when the mdtraj trajectory is created:
```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
perses/rjmc/topology_proposal.py:577: in get_water_indices
    traj = md.Trajectory(new_positions[np.newaxis, ...], md.Topology.from_openmm(new_topology))
../../../miniconda/envs/perses-dev/lib/python3.8/site-packages/mdtraj/core/trajectory.py:1169: in __init__
    self.xyz = xyz
../../../miniconda/envs/perses-dev/lib/python3.8/site-packages/mdtraj/core/trajectory.py:882: in xyz
    value = ensure_type(value, np.float32, 3, 'xyz', shape=shape,
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

```
@mikemhenry  could you help debug this further? I'm guessing something weird is happening when `new_positions[np.newaxis, ...]` is fed into `md.Trajectory`

jchodera:
I'm tremendously confused as to why the `AtomMapper` class exists only to hold a bunch of private underscore-appended static methods, rather than use a simple factory idiom like
```python
# Create an atom mapping factory
atom_mapper = AtomMapper(atom_expr=atom_expr, bond_expr=bond_expr, map_strategy=map_strategy)
# Use the factory
for target_molecule in target_molecules:
   atom_mapping = AtomMapper.create_map(source_molecule, target_molecule)
   ...
```
I'm going to at least make some notes about how to refactor this class, but may end up doing some refactoring in this PR if it seems straightforward to fix.

mikemhenry:
Looks like the tests are passing now, not sure why since the committed code -- or are you talking about some failing tests happing locally?

jchodera:
@mikemhenry @zhang-ivy : This must be a stochastic test failure then. Can you add this to the list of stochastic test failures to revisit?

jchodera:
@mikemhenry : Fixing the atom mapping issues turned into a complete overhaul of how we create atom maps. We now have an `AtomMapper` factory that creates `AtomMapping` objects:
```python
# Create an AtomMapper factory
from perses.rjmc.atom_mapping import AtomMapper
atom_mapper = AtomMapper(use_positions=True)
# You can configure options later
atom_mapper.use_positions = False # ignore positions in scoring mappings
# Now you can create an AtomMapping object
# This can take either OEMol or OpenFF Molecule objects
atom_mapping = atom_mapper.get_best_atom_mapping(old_molecule, new_molecule)
# We can retrieve these objects again
old_mol = atom_mapping.old_mol
new_mol = atom_mapping.new_mol
# And get consistent atom index maps
atom_mapping.old_to_new_atom_map
atom_mapping.new_to_old_atom_map
# We can even create an image of the mapping
atom_mapping.render_image('mapping.png')
# We can also modify this in place, and the old-to-new and new-to-old atom maps will stay fully consistent
atom_mapping.old_to_new_atom_map = { 0:0, 1:1, 2:2 }
```
Encapsulating all the logic within the `AtomMapping` and `AtomMapper` classes greatly simplified the code.

I still need to add more unit test coverage and spend more time inspecting results, but I should have this wrapped up in another day. There's one test that was manually constructed that is failing that I still need to understand, as well as likely a few others (at least) I had been ignoring.

Any feedback/comments on the idioms used here would be appreciated while this is still in progress!

jchodera:
I should also mention this was not intended to be a complete conversion of the atom mapping facility from the OpenEye toolkit to the OpenFF Molecule objects. We still need to rely on the OpenEye toolkit for the foreseeable future due to its MCSS and image generation capabilities, but moving to Molecule objects _has_ simplified a lot of things (except stereochemistry---grr!).

perses/rjmc/atom_mapping.py
```diff
@@ -0,0 +1,1305 @@
...
+    molb : oechem.OEMol
+    """
+
+    import openeye.oechem as oechem
+    for a_atom, b_atom in atommap.items():
+        if mola.GetAtom(oechem.OEHasAtomIdx(a_atom)).GetAtomicNum() == 1 or molb.GetAtom(oechem.OEHasAtomIdx(b_atom)).GetAtomicNum() == 1:
+            return True
+
+    return False
+
```
mikemhenry, line 40: Noting uses it, we can drop it.

```suggestion
```

jchodera:
Here's an example where my proposed `AtomMapping.unmap_partially_mapped_cycles()` doesn't work as expected:
```
AtomMapping : [H:39][c:35]1[c:34]([c:33]([c:32]2[c:31]([c:36]1[H:40])[c:17]([c:18]([c:19]([c:16]2[O:15][C:14]([H:28])([H:29])[C:13]([H:26])([H:27])[C:12]([H:24])([H:25])[C:11]3=[C:4]([O:43][c:5]4[c:10]3[c:9]([c:8]([c:7]([c:6]4[H:20])[H:21])[H:22])[H:23])[C:2](=[O:1])[O-:3])[H:30])[H:41])[H:42])[H:37])[H:38] -> [H:22][c:8]1[c:7]([c:6]([c:5]2[c:10]([c:9]1[H:23])[C:11](=[C:4]([N:33]2[H:34])[C:2](=[O:1])[O-:3])[C:12]([H:24])([H:25])[C:13]([H:26])([H:27])[C:14]([H:28])([H:29])[O:15][c:16]3[c:19]([c:18]([c:17]([c:37]4[c:38]3[C:39](=[C:35]([N:36]4[H:40])[H:42])[H:41])[H:32])[H:31])[H:30])[H:20])[H:21] : mapped atoms {0: 0, 1: 1, 2: 2, 3: 3, 4: 4, 5: 5, 6: 6, 7: 7, 8: 8, 9: 9, 10: 10, 11: 11, 12: 12, 13: 13, 14: 14, 15: 15, 16: 16, 17: 36, 18: 37, 19: 19, 20: 20, 21: 21, 22: 22, 23: 24, 24: 23, 25: 26, 26: 25, 27: 27, 28: 28, 30: 17, 31: 18, 41: 31}
```
Initial map:
image
Map after `unmap_partially_mapped_cycles()`:
image

It looks like I'll need to find the maximum subgraph of contiguously mapped atoms and then de-map everything outside of that.

jchodera:
I think I'll need to fully implement #792 if the cycle basis check we already have implemented is failing here.

jchodera:
My simple idea of checking all cycles are either fully mapped or not mapped isn't quite working, it seems. I'm digging into this further, but in this case, it woudl de-map the 6-5 ring at the top but would not also de-map the carboxylic acid moiety on the upper right:
image
Note that we should really permit the mapping of O -> N in this case, but there seems to be no single optimal `OEExprOpts` choice for `atom_expr`:
* If we choose `OEExprOpts_Hybridization`, then we don't map O -> N on the top here:
image
* If we choose `OEExprOpts_HvyDegree`, then we have to unmap a heavy atom if we change a -H substituent to a halogen:
image

In addition, the scheme from @proteneer in [this code](https://github.com/proteneer/timemachine/blob/c08bfda717d535b0ae9c0ec71f661138819192fe/fe/topology.py#L452-L529) is far too restrictive, since it will only allow a _single_ substituent to be deleted or added at once. We want to be able to permit multiple substituents to be added or deleted simultaneously, so we'll need to come up with another approach to check for (and enforce) factorizability of partition functions.

jchodera:
I've updated the `default` atom mapping criteria based on examining Mcl1 failures, but our ring demapping algorithm still needs to be updated to avoid situations like this, where substituents are not de-mapped if they're on the wrong side of the maximal connected mapped subgraph:
image
Perhaps we just need to enumerate all connected subgraphs in the old and new molecules and de-map any atoms that are not in the largest connected subgraph?

jchodera:
I've overhauled the algorithm for detecting transformations that create/break cycles, as well as repairing these transformations, but now I'm running into some weird mapping issues. Consider this mapping:
image
Note the weird mapping of atom 17.

I need to track down which code path is doing this and then see if there's an indexing issue somewhere.

jchodera:
The updated `AtomMapper` code now generates considerably improved atom mappings for the JACS test set.

I'm now working on trying to figure out how the atom mapping facility is supposed to be used for biopolymer residues like amino acids. Hopefully I can sort this out and revive the rest of the tests.

jchodera:
Just noting this here for my future self: The major issue with why the current biopolymer atom mapping is failing is that the current scheme seems to make use of the fact that OEMol atom indices can be arbitrarily set, and this facility is used to keep track of which atoms in part of the residue are mapped. My atom mapping tool instead returns indices of the atoms within the OpenFF `Molecule` object, which does not support the concept of attached arbitrary atom indices.

This is easy to work around, but the ability to tag atoms of a `Molecule` with indices may be useful in future.

jchodera:
We're down to two test failures remaining!

- [ ] @dominicrufa @zhang-ivy : I'll need your help to sort out this issue with the repartitioned topology that likely appeared previously as a stochastic error since we were not exhaustively testing transformations here before:
<details>

<summary>Failure log for `test_RepartitionedHybridTopologyFactory_energies`</summary>

```

=================================== FAILURES ===================================
_______________ test_RepartitionedHybridTopologyFactory_energies _______________
[gw0] linux -- Python 3.9.6 /usr/share/miniconda/envs/test/bin/python

    def test_RepartitionedHybridTopologyFactory_energies():
        """
        Test whether the difference in the nonalchemical zero and alchemical zero states is the forward valence energy.  Also test for the one states.
        """
    
        from perses.tests.test_topology_proposal import generate_atp
        from openmmforcefields.generators import SystemGenerator
    
        # Test alanine dipeptide in vacuum
        atp, system_generator = generate_atp('vacuum')
        RepartitionedHybridTopologyFactory_energies(atp.topology, '1', atp.system, atp.positions, system_generator)
    
        # Test alanine dipeptide in solvent
        atp, system_generator = generate_atp('solvent')
>       RepartitionedHybridTopologyFactory_energies(atp.topology, '1', atp.system, atp.positions, system_generator)

perses/tests/test_relative.py:784: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
perses/tests/test_relative.py:665: in RepartitionedHybridTopologyFactory_energies
    RepartitionedHybridTopologyFactory_energy_mutant(topology, chain, system, positions, system_generator, mutant)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

topology = <Topology; 3 chains, 516 residues, 1557 atoms, 1043 bonds>
chain = '1'
system = <simtk.openmm.openmm.System; proxy of <Swig Object of type 'OpenMM::System *' at 0x7fddf8b4bea0> >
positions = Quantity(value=array([[ 2.00000100e-01,  1.00000000e-01, -1.30000000e-07],
       [ 2.00000100e-01,  2.09000000e-01,  ...0e-01, -6.70000000e-01,  1.08389999e+00],
       [ 1.29177625e+00,  1.14910000e+00,  1.20299985e-01]]), unit=nanometer)
system_generator = <openmmforcefields.generators.system_generators.SystemGenerator object at 0x7fde0c30b4f0>
mutant = 'GLU'

    def RepartitionedHybridTopologyFactory_energy_mutant(topology, chain, system, positions, system_generator, mutant):
        """
        Test whether the difference in the nonalchemical zero and alchemical zero states is the forward valence energy.  Also test for the one states.
        Note that two RepartitionedHybridTopologyFactorys need to be generated (one for each endstate) because the energies need to be validated separately for each endstate.
        Tests a single mutant.
        """
    
        from perses.rjmc.topology_proposal import PointMutationEngine
        from perses.annihilation.relative import RepartitionedHybridTopologyFactory
        from perses.tests.utils import validate_endstate_energies
    
        ENERGY_THRESHOLD = 1e-4
    
        for res in topology.residues():
            if res.id == '2':
                wt_res = res.name
    
        print(f'Making mutation {wt_res}->{mutant}')
    
        # Create point mutation engine to mutate residue at id 2 to random amino acid
        point_mutation_engine = PointMutationEngine(wildtype_topology=topology,
                                                    system_generator=system_generator,
                                                    chain_id=chain,
                                                    max_point_mutants=1,
                                                    residues_allowed_to_mutate=['2'],  # the residue ids allowed to mutate
                                                    allowed_mutations=[('2', mutant)],
                                                    aggregate=True)  # always allow aggregation
    
        # Create topology proposal
        topology_proposal = point_mutation_engine.propose(current_system=system, current_topology=topology)
    
        # Create geometry engine
        from perses.rjmc.geometry import FFAllAngleGeometryEngine
        geometry_engine = FFAllAngleGeometryEngine(metadata=None,
                                                   use_sterics=False,
                                                   n_bond_divisions=100,
                                                   n_angle_divisions=180,
                                                   n_torsion_divisions=360,
                                                   verbose=True,
                                                   storage=None,
                                                   bond_softening_constant=1.0,
                                                   angle_softening_constant=1.0,
                                                   neglect_angles=False,
                                                   use_14_nonbondeds=True)
    
        # Create geometry proposal
        new_positions, logp_proposal = geometry_engine.propose(topology_proposal, positions, beta,
                                                                       validate_energy_bookkeeping=True)
        logp_reverse = geometry_engine.logp_reverse(topology_proposal, new_positions, positions, beta,
                                                    validate_energy_bookkeeping=True)
    
        if not topology_proposal.unique_new_atoms:
            assert geometry_engine.forward_final_context_reduced_potential == None, f"There are no unique new atoms but the geometry_engine's final context reduced potential is not None (i.e. {self._geometry_engine.forward_final_context_reduced_potential})"
            assert geometry_engine.forward_atoms_with_positions_reduced_potential == None, f"There are no unique new atoms but the geometry_engine's forward atoms-with-positions-reduced-potential in not None (i.e. { self._geometry_engine.forward_atoms_with_positions_reduced_potential})"
            vacuum_added_valence_energy = 0.0
        else:
            added_valence_energy = geometry_engine.forward_final_context_reduced_potential - geometry_engine.forward_atoms_with_positions_reduced_potential
    
        if not topology_proposal.unique_old_atoms:
            assert geometry_engine.reverse_final_context_reduced_potential == None, f"There are no unique old atoms but the geometry_engine's final context reduced potential is not None (i.e. {self._geometry_engine.reverse_final_context_reduced_potential})"
            assert geometry_engine.reverse_atoms_with_positions_reduced_potential == None, f"There are no unique old atoms but the geometry_engine's atoms-with-positions-reduced-potential in not None (i.e. { self._geometry_engine.reverse_atoms_with_positions_reduced_potential})"
            subtracted_valence_energy = 0.0
        else:
            subtracted_valence_energy = geometry_engine.reverse_final_context_reduced_potential - geometry_engine.reverse_atoms_with_positions_reduced_potential
    
        # Generate repartitioned htf at lambda = 0
        zero_htf = RepartitionedHybridTopologyFactory(topology_proposal=topology_proposal,
                              current_positions=positions,
                              new_positions=new_positions,
                              endstate=0)
    
        # Compute error at lambda = 0 endstate
        zero_state_error, _ = validate_endstate_energies(zero_htf._topology_proposal,
                                                         zero_htf,
                                                         added_valence_energy,
                                                         subtracted_valence_energy,
                                                         ENERGY_THRESHOLD=ENERGY_THRESHOLD,
                                                         platform=openmm.Platform.getPlatformByName('Reference'),
                                                         repartitioned_endstate=0)
        # Generate repartitioned htf at lambda = 1
        one_htf = RepartitionedHybridTopologyFactory(topology_proposal=topology_proposal,
                                                      current_positions=positions,
                                                      new_positions=new_positions,
                                                      endstate=1)
    
        # Compute error at lambda = 1 endstate
        _, one_state_error = validate_endstate_energies(one_htf._topology_proposal,
                                                         one_htf,
                                                         added_valence_energy,
                                                         subtracted_valence_energy,
                                                         ENERGY_THRESHOLD=ENERGY_THRESHOLD,
                                                         platform=openmm.Platform.getPlatformByName('Reference'),
                                                         repartitioned_endstate=1)
    
        # Check that endstate errors are below threshold
        assert abs(zero_state_error) < ENERGY_THRESHOLD, f"The zero state alchemical and nonalchemical energy absolute difference {abs(zero_state_error)} is greater than the threshold of {ENERGY_THRESHOLD}."
>       assert abs(one_state_error) < ENERGY_THRESHOLD, f"The one state alchemical and nonalchemical energy absolute difference {abs(one_state_error)} is greater than the threshold of {ENERGY_THRESHOLD}."
E       AssertionError: The one state alchemical and nonalchemical energy absolute difference 0.2041063642262273 is greater than the threshold of 0.0001.
E       assert 0.2041063642262273 < 0.0001
E        +  where 0.2041063642262273 = abs(-0.2041063642262273)

perses/tests/test_relative.py:764: AssertionError
----------------------------- Captured stdout call -----------------------------
```
</details>

<summary>Energy breakdown</summary>

```
Making mutation ALA->GLU
Warning: Returning Reference platform instead of requested platform CUDA
conducting subsequent work with the following platform: Reference
conducting subsequent work with the following platform: CPU
conducting subsequent work with the following platform: CPU
conducting subsequent work with the following platform: CPU
conducting subsequent work with the following platform: CPU
Warning: Returning Reference platform instead of requested platform CUDA
conducting subsequent work with the following platform: Reference
conducting subsequent work with the following platform: CPU
added energy components: [('CustomBondForce', 1.580827770516917), ('CustomAngleForce', 4.3543592013601105), ('CustomTorsionForce', 7.050911762338636), ('CustomBondForce', 20.71734407448997)]
Warning: Returning Reference platform instead of requested platform CUDA
conducting subsequent work with the following platform: Reference
conducting subsequent work with the following platform: CPU
conducting subsequent work with the following platform: CPU
conducting subsequent work with the following platform: CPU
conducting subsequent work with the following platform: CPU
Warning: Returning Reference platform instead of requested platform CUDA
conducting subsequent work with the following platform: Reference
conducting subsequent work with the following platform: CPU
added energy components: [('CustomBondForce', 0.0), ('CustomAngleForce', 0.00011854972790545768), ('CustomTorsionForce', 0.0014456143524425994), ('CustomBondForce', 4.034319661495377)]
conducting subsequent work with the following platform: Reference
conducting subsequent work with the following platform: CPU
			HarmonicBondForce: 0.034551580662335844
			HarmonicAngleForce: 0.6072024919297015
			PeriodicTorsionForce: 16.176829149610796
			NonbondedForce: -796.2718391330847
			AndersenThermostat: 0.0
			MonteCarloBarostat: 0.0
conducting subsequent work with the following platform: CPU
			HarmonicBondForce: 1.615379351179253
			HarmonicAngleForce: 4.961561693289812
			PeriodicTorsionForce: 23.22774091194943
			NonbondedForce: -775.5554960693804
			AndersenThermostat: 0.0
			MonteCarloBarostat: 0.0
conducting subsequent work with the following platform: CPU
			HarmonicBondForce: 1.615379351179253
			HarmonicAngleForce: 4.9614431435619055
			PeriodicTorsionForce: 23.22629529759699
			NonbondedForce: 2162531545690.6074
			AndersenThermostat: 0.0
			MonteCarloBarostat: 0.0
conducting subsequent work with the following platform: Reference
conducting subsequent work with the following platform: CPU
			HarmonicBondForce: 0.034551580662335844
			HarmonicAngleForce: 0.6072024919297015
			PeriodicTorsionForce: 16.176829149610796
			NonbondedForce: -796.2718391330847
			AndersenThermostat: 0.0
			MonteCarloBarostat: 0.0
conducting subsequent work with the following platform: CPU
			HarmonicBondForce: 1.615379351179253
			HarmonicAngleForce: 4.961561693289811
			PeriodicTorsionForce: 23.22774091194943
			NonbondedForce: 2162531545748.2427
			AndersenThermostat: 0.0
			MonteCarloBarostat: 0.0
conducting subsequent work with the following platform: CPU
			HarmonicBondForce: 1.615379351179253
			HarmonicAngleForce: 4.9614431435619055
			PeriodicTorsionForce: 23.22629529759699
			NonbondedForce: 2162531545690.6074
			AndersenThermostat: 0.0
			MonteCarloBarostat: 0.0
------------------------------ Captured log call -------------------------------
```
</details>

- [ ] Here's a failed test that I need to update that had simply hard-coded the "correct" mapping results:
```
[gw0] linux -- Python 3.9.6 /usr/share/miniconda/envs/test/bin/python

pairs_of_smiles = [('Cc1ccccc1', 'c1ccc(cc1)N'), ('CC(c1ccccc1)', 'O=C(c1ccccc1)'), ('Oc1ccccc1', 'Sc1ccccc1')]
test = True

    def test_mapping_strength_levels(pairs_of_smiles=[('Cc1ccccc1','c1ccc(cc1)N'),('CC(c1ccccc1)','O=C(c1ccccc1)'),('Oc1ccccc1','Sc1ccccc1')],test=True):
    
        correct_results = {0:{'default': (3,2), 'weak':(3,2), 'strong':(4,3)},
                           1:{'default': (7,3), 'weak':(6,2), 'strong':(7,3)},
                           2:{'default': (1,1), 'weak':(1,1), 'strong':(2,2)}}
    
        mapping = ['weak','default','strong']
    
        for example in mapping:
            for index, (lig_a, lig_b) in enumerate(pairs_of_smiles):
                print(f"conducting {example} mapping with ligands {lig_a}, {lig_b}")
                initial_molecule = smiles_to_oemol(lig_a)
                proposed_molecule = smiles_to_oemol(lig_b)
                molecules = [Molecule.from_openeye(mol) for mol in [initial_molecule, proposed_molecule]]
                system_generator = SystemGenerator(forcefields = forcefield_files, barostat=barostat, forcefield_kwargs=forcefield_kwargs,nonperiodic_forcefield_kwargs=nonperiodic_forcefield_kwargs,
                                                     small_molecule_forcefield = 'gaff-1.81', molecules=molecules, cache=None)
                proposal_engine = SmallMoleculeSetProposalEngine([initial_molecule, proposed_molecule], system_generator)
                proposal_engine.map_strength = example
                initial_system, initial_positions, initial_topology = OEMol_to_omm_ff(initial_molecule, system_generator)
                print(f"running now with map strength {example}")
                proposal = proposal_engine.propose(initial_system, initial_topology)
                print(lig_a, lig_b,'length OLD and NEW atoms',len(proposal.unique_old_atoms), len(proposal.unique_new_atoms))
                if test:
                    render_atom_mapping(f'{index}-{example}.png', initial_molecule, proposed_molecule, proposal._new_to_old_atom_map)
>                   assert ( (len(proposal.unique_old_atoms), len(proposal.unique_new_atoms)) == correct_results[index][example]), f"the mapping failed, correct results are {correct_results[index][example]}"
E                   AssertionError: the mapping failed, correct results are (7, 3)
E                   assert (6, 2) == (7, 3)
E                     At index 0 diff: 6 != 7
E                     Full diff:
E                     - (7, 3)
E                     + (6, 2)

perses/tests/test_topology_proposal.py:99: AssertionError
----------------------------- Captured stdout call -----------------------------
```

jchodera:
OK, the hard-coded map test is fixed, and this PR should be ready for review!
I'll incorporate input from @zhang-ivy @dominicrufa when they can provide it about the failing ALA->GLU mutation energy comparison test, but I'm guessing that test failure occurs because the nonbonded forces are just very large:
```
conducting subsequent work with the following platform: CPU
			HarmonicBondForce: 1.615379351179253
			HarmonicAngleForce: 4.961561693289811
			PeriodicTorsionForce: 23.22774091194943
			NonbondedForce: 2162531545748.2427
			AndersenThermostat: 0.0
			MonteCarloBarostat: 0.0
conducting subsequent work with the following platform: CPU
			HarmonicBondForce: 1.615379351179253
			HarmonicAngleForce: 4.9614431435619055
			PeriodicTorsionForce: 23.22629529759699
			NonbondedForce: 2162531545690.6074
			AndersenThermostat: 0.0
			MonteCarloBarostat: 0.0
```

jchodera:
One way we might fix the `test_RepartitionedHybridTopologyFactory_energies` issue is to either minimize the energy or generate multiple conformers until we get one with a non-crazy energy, and then compare the energies for that conformation. But I suggest we create a different issue for that.

Tagging @mikemhenry to review this on Monday if possible!

jchodera:
I've now enabled `AtomMapping` classes to render themselves inside Jupyter notebooks!
image
I've used this feature to overhaul the atom mapping example notebook so we can show off the new features at Monday's meeting:
https://github.com/choderalab/perses/blob/refactor-small-molecule-proposal/examples/atom-mapping/Atom-mapping.ipynb

zhang-ivy:
>Just noting this here for my future self: The major issue with why the current biopolymer atom mapping is failing is that the current scheme seems to make use of the fact that OEMol atom indices can be arbitrarily set, and this facility is used to keep track of which atoms in part of the residue are mapped. My atom mapping tool instead returns indices of the atoms within the OpenFF Molecule object, which does not support the concept of attached arbitrary atom indices.
@jchodera : Does this PR break biopolymer atom mapping? If so, what is the short term solution here for making sure that we can map protein mutations?

> One way we might fix the test_RepartitionedHybridTopologyFactory_energies issue is to either minimize the energy or generate multiple conformers until we get one with a non-crazy energy, and then compare the energies for that conformation. But I suggest we create a different issue for that.

I can look into adding a minimization step to the tests today.

jchodera:
> @jchodera : Does this PR break biopolymer atom mapping? If so, what is the short term solution here for making sure that we can map protein mutations?

No. All tests pass. So if the test coverage in `master` covered your use cases, we should not break them.

If you have unit tests _outside_ of `master` that are important, it would be critical to share those now!

> I can look into adding a minimization step to the tests today.

Let's tackle that in a separate PR!

perses/app/relative_setup.py
```diff
@@ -122,27 +122,21 @@ def __init__(self,
...
             if True, `complex` must be in `phases` and .sdf or .mol2 file of ligand must be provided
+        given_geometries_tolerance : simtk.unit.Quantity with units of length, default=0.2*angstrom
+            If use_given_geometries=True, use this tolerance for identifying mapped atoms
         """
         from openeye import oechem
 
+        # TODO: Refactor this class into initializer that configures options and factory methods that use them.
+        # QUESTION: Why do we add these as private object attributes?
+        #   If we want users to be able to modify them, they should not be private.
+        #   Also, these are all used immediately within __init__, so it doesn't make sense to store them this way.
```
ijpulidos, line 139: Yeah, I agree. It should be better to separate these into different methods that use the attributes.

perses/app/relative_setup.py
```diff
@@ -122,27 +122,21 @@ def __init__(self,
...
-            - core will return the map with the largest number of atoms in the core. If there are multiple maps with the same highest score, then `matching_criterion` is used to tie break
-            - geometry uses the coordinates of the molB oemol to calculate the heavy atom distance between the proposed map and the actual geometry
-            this can be vital for getting the orientation of ortho- and meta- substituents correct in constrained (i.e. protein-like) environments.
-            this is ONLY useful if the positions of ligand B are known and/or correctly aligned.
-            - matching_criterion uses the `matching_criterion` flag to pick which of the maps best satisfies a 2D requirement.
-            - random will use a random map of those that are possible
-            - weighted-random uses a map chosen at random, proportional to how close it is in geometry to ligand B. The same as for 'geometry', this requires the coordinates of ligand B to be meaninful
-            - return-all BREAKS THE API as it returns a list of dicts, rather than list. This is intended for development code, not main pipeline.
```
ijpulidos, line 136: Wouldn't removing these would affect the expected behavior of some current/older code or uses?

perses/app/setup_relative_calculation.py
```diff
@@ -447,7 +458,7 @@ def run_setup(setup_options, serialize_systems=True, build_samplers=True):
                                           trajectory_directory=trajectory_directory, trajectory_prefix=setup_options['trajectory_prefix'], nonbonded_method=setup_options['nonbonded_method'],
 
                                           complex_box_dimensions=setup_options['complex_box_dimensions'],solvent_box_dimensions=setup_options['solvent_box_dimensions'], ionic_strength=ionic_strength, remove_constraints=setup_options['remove_constraints'],
-                                          use_given_geometries = use_given_geometries)
+                                          use_given_geometries=use_given_geometries, given_geometries_tolerance=given_geometries_tolerance)
```
ijpulidos, line 461: This part is kind of messy. Don't we have a code formatter (such as `black`) to handle these things? If not we should set something like that, eventually.

jchodera: Great idea! Can you suggest we adopt `black` in the issue tracker? We can implement that cleanup separately.

perses/app/relative_setup.py
jchodera: I changed all uses of `map_strategy` within `perses`, but if there are external users (or unmerged branches), these would need to update to the new API. But this can't be helped, and we're still pre-1.0 without a stable API.

perses/app/relative_setup.py
jchodera: I believe @mikemhenry is working on refactoring this file right now for 0.10.0

mikemhenry (approved):
LGTM! I really like the rework of the atom mapping API

perses/rjmc/atom_mapping.py
```diff
@@ -0,0 +1,1475 @@
...
+        if (atom_mappings is None) or len(atom_mappings)==0:
+            return None
+
+        scores = np.array([ self.score_mapping(atom_mapping) for atom_mapping in atom_mappings ])
+        best_map_index = np.argmax(scores)
+
+        elapsed_time = time.time() - initial_time
+        _logger.debug(f'get_best_mapping took {elapsed_time:.3f} s')
+
+        return atom_mappings[best_map_index]
```
ijpulidos, line 992: Can we get multiple mappings with the same max score? What happens when that's the case?

perses/rjmc/atom_mapping.py
```diff
@@ -0,0 +1,1475 @@
...
+                for atom, oeatom in zip(offmol.atoms, oemol.GetAtoms()):
+                    atom.name = oeatom.GetName()
+            elif hasattr(mol, 'GetAtoms'):
+                # OpenEye OEMol
+                offmol = Molecule(mol)
+                oemol = oechem.OEMol(mol)
+            else:
+                raise ValueError(f'Not sure how to copy {mol}; should be openff.toolkit.topology.Molecule or openeye.oechem.OEMol')
+
+            return offmol, oemol
```
ijpulidos, line 794: Not sure how bad we are in memory usage or if we would ever need a better memory footprint. But this might be a candidate for improvement, maybe returning just copies of the same type as the input. That is, if input is an `OEMol` output is just `OEMol` as well (similar with `offmol`).

perses/rjmc/atom_mapping.py
```diff
@@ -0,0 +1,1475 @@
...
+        Sample a mapping stochasticaly between ethane and ethanol
+
+        >>> from openff.toolkit.topology import Molecule
+        >>> ethane = Molecule.from_smiles('[C:0]([H:1])([H:2])([H:3])[C:4]([H:5])([H:6])([H:7])')
+        >>> ethanol = Molecule.from_smiles('[C:0]([H:1])([H:2])([H:3])[C:4]([H:5])([H:6])[O:7][H:8]')
+        >>> atom_mapping = atom_mapper.get_sampled_mapping(ethane, ethanol)
+
+        """
+        import numpy as np
+        atom_mappings = self.get_all_maps(old_mol, new_mol)
```
ijpulidos, line 1020: I believe `get_all_maps` doesn't exist as a public/user-visible method. Is this supposed to be `_get_all_maps`? Now that makes me wonder if this method is being tested at all (probably not).

perses/app/setup_relative_calculation.py
jchodera: Issue opened in #827

perses/rjmc/atom_mapping.py
```diff
@@ -0,0 +1,1475 @@
...
+        Derive atom mapping from positions:
+
+        >>> atom_mapper = AtomMapper()
+        >>> old_mol = Molecule.from_file('old_mol.sdf')
+        >>> new_mol = Molecule.from_file('new_mol.sdf')
+        >>> atom_mapping = atom_mapper.generate_atom_mapping_from_positions(old_mol, new_mol)
+
+        """
+        from openff.toolkit.topology import Molecule
+        distance_unit = unit.angstroms # distance unit for comparisons
```
ijpulidos, line 1201: Are units always expected to be angstroms? I guess this works for the most part but maybe we could try to get the actual units from the molecule conformers themselves. Something like `old_mol.conformers[0].unit` (assuming they all have the same units as the first occurrence)

ijpulidos (approved):
This is a great PR. Lots of improvements in the API. There are some comments to be discussed as usual but nothing blocking. This can be merged.

perses/rjmc/atom_mapping.py
ijpulidos: Actually I now realize as well that the `_get_all_maps` static method has an undefined reference to `self` if no atom or bond expressions are passed.

perses/rjmc/atom_mapping.py
jchodera: The `distance_unit` here is arbitrary, but defined here to be consistent in how both the measured distances and absolute distance threshold are converted into unitless numbers. So the choice is totally arbitrary.

perses/rjmc/atom_mapping.py
jchodera: I am hoping to refactor the private `_get_all_map()` method in a future PR---once I fully understand what it does!
This was just a partial refactor, cleaning up the public API, eliminating references to the private API when possible, and refactoring the bits of the code I could understand. We'll need @dominicrufa's help (after he returns from his internship) for the next phase of refactoring/cleanup.

perses/rjmc/atom_mapping.py
jchodera: Great point! The hope is to eliminate `OEMol` objects entirely when possible, and just use `Molecule` in the future. This is just a temporary hack.

perses/rjmc/atom_mapping.py
jchodera: Great point! Added a comment to address this ambiguity.

perses/rjmc/atom_mapping.py
```diff
@@ -0,0 +1,1475 @@
...
+                oemol = oechem.OEMol(mol)
+            else:
+                raise ValueError(f'Not sure how to copy {mol}; should be openff.toolkit.topology.Molecule or openeye.oechem.OEMol')
+
+            return offmol, oemol
+
+        old_offmol, old_oemol = copy_molecule(old_mol)
+        new_offmol, new_oemol = copy_molecule(new_mol)
+
+        if (old_offmol.n_atoms==0) or (new_offmol.n_atoms==0):
```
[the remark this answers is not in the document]
jchodera: That would take more code, since we don't know whether they are `OEMol` or `Molecule` objects at the beginning.

perses/rjmc/atom_mapping.py
```diff
@@ -0,0 +1,1475 @@
...
+
+        # Store atom maps
+        if (old_to_new_atom_map is not None) and (new_to_old_atom_map is not None):
+            raise ValueError('Only one of old_to_new_atom_map or new_to_old_atom_map can be specified')
+        if (old_to_new_atom_map is None) and (new_to_old_atom_map is None):
+            raise ValueError('One of old_to_new_atom_map or new_to_old_atom_map must be specified')
+        if old_to_new_atom_map is not None:
+            self.old_to_new_atom_map = old_to_new_atom_map
+        if new_to_old_atom_map is not None:
+            self.new_to_old_atom_map = new_to_old_atom_map
```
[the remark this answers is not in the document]
jchodera: The idea is for `AtomMapping` to provide any "views" to the data that will be convenient to minimize the complexity of downstream code that uses this object. There are some places where `old_to_new` is easier to use, other places where `new_to_old` is easier to use.

perses/rjmc/atom_mapping.py
```diff
@@ -0,0 +1,1475 @@
...
+        self.new_mol = Molecule(new_mol, allow_undefined_stereo=True)
+
+        # Fix atom name retrieval from OEMol objects
+        # TODO: When https://github.com/openforcefield/openff-toolkit/issues/1026 is fixed, remove this
+        if hasattr(old_mol, 'GetAtoms'):
+            for index, atom in enumerate(old_mol.GetAtoms()):
+                self.old_mol.atoms[index].name = atom.GetName()
+        if hasattr(new_mol, 'GetAtoms'):
+            for index, atom in enumerate(new_mol.GetAtoms()):
+                self.new_mol.atoms[index].name = atom.GetName()
```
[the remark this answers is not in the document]
jchodera: For now, this code doesn't do any harm, and is compatible with the `openff-toolkit` current and future release.
At some point in the future, we can change the dependency to require the new version of `openff-toolkit` or later, and then remove this code---hence the `#TODO`!

perses/tests/test_atom_mapping.py
```diff
@@ -0,0 +1,268 @@
...
+                    else:
+                        _logger.info(f'{mol1_smiles} -> {mol2_smiles} using map strength {map_strength} : {atom_mapping}')
+
+                else:
+                    # Check that expected number of mapped atoms are provided
+                    n_mapped_atoms = 0
+                    if atom_mapping is not None:
+                        n_mapped_atoms = atom_mapping.n_mapped_atoms
+
+                    assert n_mapped_atoms==expected_n_mapped_atoms, "Number of mapped atoms does not match hand-calibrated expectation"
```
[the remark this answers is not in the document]
jchodera: Yes! Great idea---please open an issue!

perses/tests/test_atom_mapping.py
```diff
@@ -0,0 +1,268 @@
...
+
+################################################################################
+# AtomMapping
+################################################################################
+
+class TestAtomMapping(unittest.TestCase):
+    """Test AtomMapping object."""
+    def setUp(self):
+        """Create useful common objects for testing."""
+        from openff.toolkit.topology import Molecule
```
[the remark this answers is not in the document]
jchodera: Good catch!

perses/tests/test_atom_mapping.py
```diff
@@ -0,0 +1,268 @@
...
+        assert atom_mapping.n_mapped_atoms == n_mapped_atoms_old
+
+        # Test methyl-cyclohexane -> methyl-cyclopentane, demapping the ring transformation
+        from openff.toolkit.topology import Molecule
+        old_mol = Molecule.from_smiles('[C:0][C:1]1[C:2][C:3][C:4][C:5][C:6]1') # methyl-cyclohexane
+        new_mol = Molecule.from_smiles('[C:0][C:1]1[C:2][C:3][C:4][C:5]1') # methyl-cyclopentane
+        old_to_new_atom_map = { 0:0, 1:1, 2:2, 3:3, 5:4, 6:5 }
+        new_to_old_atom_map = dict(map(reversed, self.old_to_new_atom_map.items()))
+        atom_mapping = AtomMapping(old_mol, new_mol, old_to_new_atom_map=old_to_new_atom_map)
+        assert atom_mapping.creates_or_breaks_rings() == True
```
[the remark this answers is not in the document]
jchodera: Extra test coverage doesn't help! There are a lot of combinations of options being tested here.

perses/tests/test_atom_mapping.py
```diff
@@ -0,0 +1,268 @@
...
+    def test_str(self):
+        atom_mapping = AtomMapping(self.old_mol, self.old_mol, old_to_new_atom_map=self.old_to_new_atom_map)
+        str(atom_mapping)
+
+    def test_render_image(self):
+        import tempfile
+        atom_mapping = AtomMapping(self.old_mol, self.old_mol, old_to_new_atom_map=self.old_to_new_atom_map)
+        for suffix in ['.pdf', '.png', '.svg']:
+            with tempfile.NamedTemporaryFile(suffix=suffix) as tmpfile:
+                atom_mapping.render_image(tmpfile.name)
```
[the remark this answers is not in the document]
jchodera: Test parameterization is probably the right way to go here, as you suggested! I'll leave that for further cleanup.

jchodera:
Changes should be complete. Just waiting on tests to pass now.

Issue #803, opened 2021-05-31
Use versioneer to manage versions


jchodera:
It looks like we have the [version number hard-coded into `setup.py`](https://github.com/choderalab/perses/blob/master/setup.py#L37), meaning we just released a package that thinks it is `0.8.2dev0` as `0.9.0` on `conda-forge`!

mikemhenry:
Fixed with #812

Issue #788, opened 2021-05-12
Update setup.py


mikemhenry:
Right now `setup.py` lists the incorrect license + python versions we support. This should be fixed.

mikemhenry:
Fixed with #812

PR #812, opened 2021-06-21
Add versioneer to track package version automatically 

mikemhenry:
Fixes issue #803 

Also since I had to modify setup.py, I fixed the license and supported python versions, which fixes issue #788


setup.py
```diff
@@ -51,33 +49,49 @@ def find_package_data(data_root, package_root):
...
-       'openmm.forcefielddir' : ['perses=perses:get_datadir'],
-       'console_scripts' : ['perses-relative = perses.app.setup_relative_calculation:run','perses-fah = perses.app.fah_generator:run']
-      },
-      ext_modules=extensions,
-      install_requires=[],
-      )
+setup(
+    name="perses",
+    author="Patrick Grinaway",
+    author_email="patrick.grinaway@choderalab.org",
```
jchodera, line 62: We should update the authorship lines here. I could not find [a way to specify multiple authors](https://stackoverflow.com/questions/9999829/how-to-specify-multiple-authors-emails-in-setup-py), so how about I claim responsibility for now?
```python
    author="John D. Chodera",
    author_email="john.chodera@choderalab.org",
```
```

jchodera (approved):
Thanks so much! Please merge after updating the `author` fields.

setup.py
mikemhenry: Sounds good! I view scientific software authorship like it works with papers, last name got the funding and is the person to email if you have questions!

jchodera:
> Sounds good! I view scientific software authorship like it works with papers, last name got the funding and is the person to email if you have questions!

Well, it's more for maintainership purposes right now: I'll still be the one users can bug even when folks move on!

PR #819, opened 2021-07-15
Create RestCapablePMEHybridTopologyFactory (OLD)

zhang-ivy:
I've subclassed `RxnHybridTopologyFactory` to create `RestCapablePMEHybridTopologyFactory` (please let me know if you have class name suggestions) that handles:
        - alchemical modification of atoms
        - REST scaling
        - 4th dimension softcore
      
For nonbonded interactions/exceptions:
- We use a CustomNonbondedForce and CustomBondForce (with PME) to handle the direct space. 
- (Not yet implemented) We will leave the reciprocal space in a NonbondedForce and disable the direct space using [this PR](https://github.com/openmm/openmm/pull/3173#issuecomment-880839738)
- Unlike RxnHybridTopologyFactory, we wrote this factory to only handle one REST region and one alchemical region, which simplifies the code a lot.

We handle valence terms in the same way as RxnHybridTopologyFactory (though we will probably want to re-factor the valence functions to match the custom nb/exceptions forces described here).

@dominicrufa could you review this and then tag Peter for feedback?


perses/annihilation/relative.py
```diff
@@ -2629,3 +2629,2206 @@ def _handle_exceptions(self):
...
+        "old_sigma_scalar1 = lambda_sterics_old * sigma_old1;",
+        "new_sigma_scalar1 = lambda_sterics_new * sigma_new1;",
+
+        "sigma2 = (is_unique_old2 * old_sigma_scalar2) + (is_unique_new2 * new_sigma_scalar2) + is_core2 * (old_sigma_scalar2 + new_sigma_scalar2) + is_environment2 * (old_sigma_scalar2 + new_sigma_scalar2);",
+        "old_sigma_scalar2 = lambda_sterics_old * sigma_old2;",
+        "new_sigma_scalar2 = lambda_sterics_new * sigma_new2;",
+
+        # Define epsilon (with rest scaling)
+        "epsilon = sqrt((epsilon1 * p1_rest_scale) * (epsilon2 * p2_rest_scale));",
+        # Q for DAR: Are we sure we want to apply the rest scaling before combining epsilons?
```
dominicrufa, line 4304: no i'm not sure. it might make sense to combine eps with the mixing rule and then multiply by the scaling factor

perses/annihilation/relative.py
```diff
@@ -2629,3 +2629,2206 @@ def _handle_exceptions(self):
...
+        "alpha = sqrt(-log(2 * delta) / r_cutoff);",
+        "delta = {delta};",
+        "r_cutoff = {r_cutoff};",
+
+        # Define r_eff
+        "r_eff_electrostatics = sqrt(r^2 + w_electrostatics^2);",
+        "r_eff_sterics = sqrt(r^2 + w_sterics^2);",
+
+        # Define 4th dimension terms:
+        "w_electrostatics = is_unique_old * lambda_electrostatics_new * w_scale + is_unique_new * lambda_electrostatics_new * w_scale;", # Q for DAR: do we also want to multiply by r_cutoff (as we do in RxnHybridTopologyFactory)?
```
dominicrufa, line 4325: i guess the w value should go from 0 to some >0 value, (or backward). but the size of the scaling factor is an open parameter. i guess we can multiply it by `w_scale` * `r_cutoff` where `w_scale` should be a scalar between 0 and 1?

perses/annihilation/relative.py
dominicrufa: i think the maths require writing `(p1_rest_scale * p2_rest_scale * sqrt(epsilon1 * epsilon2))`

perses/annihilation/relative.py
```diff
@@ -2629,3 +2629,2206 @@ def _handle_exceptions(self):
...
+
+    def __init__(self,
+                 topology_proposal,
+                 current_positions,
+                 new_positions,
+
+                 # scaling arguments
+                 scale_regions=None,
+
+                 r_cutoff=1.4 * unit.nanometers,  # default cutoff for reaction field electrostatics
```
zhang-ivy, line 4391: change this to default to None
In the init, set it to whatever the og system nb cutoff is. If the og system's nb method is NoCutoff, set r_cutoff to 100

perses/annihilation/relative.py
zhang-ivy: default w_scale to 0.1

perses/annihilation/relative.py
```diff
@@ -2629,3 +2629,2206 @@ def _handle_exceptions(self):
...
+        standard_nonbonded_method = old_system_nbf.getNonbondedMethod()
+        if standard_nonbonded_method in [openmm.NonbondedForce.CutoffPeriodic, openmm.NonbondedForce.PME,
+                                         openmm.NonbondedForce.Ewald]:
+            if is_exception:
+                custom_force.setUsesPeriodicBoundaryConditions(True)
+            else:
+                custom_force.setNonbondedMethod(self._translate_nonbonded_method_to_custom(standard_nonbonded_method))
+                custom_force.setUseSwitchingFunction(False)
+                custom_force.setCutoffDistance(self._r_cutoff)
+                # Should we also explictly set the long range correction to be False?
```
zhang-ivy, line 4824: yes

perses/annihilation/relative.py
zhang-ivy: Fixed

perses/annihilation/relative.py
zhang-ivy: fixed

perses/annihilation/relative.py
zhang-ivy: fixed

jchodera:
Supplanted by #848

Issue #815, opened 2021-06-22
Stochastic test failures in RepartitionedHybridTopologyFactory_energies


jchodera:
I'm noticing some [stochastic test failures](https://github.com/choderalab/perses/pull/809/checks?check_run_id=2881527173#step:9:359) in `RepartitionedHybridTopologyFactory_energies`. I'm opening this thread to start logging them for future assessment.

mikemhenry:
Here is an example where one test passed then I reran it and it failed. 
The best way to view it (IMHO) is to use `vimdiff`. On line 52 to you can see the different index proposal order and then the different energy values calcualted. 

[works.txt](https://github.com/choderalab/perses/files/6696138/works.txt)
[failed.txt](https://github.com/choderalab/perses/files/6696139/failed.txt)

mikemhenry:
Another example:
https://github.com/choderalab/perses/runs/2880262661?check_suite_focus=true#step:9:356

jchodera:
Thanks! This should be enough for me to dig into!

mikemhenry:
https://github.com/choderalab/perses/runs/2937235508?check_suite_focus=true#step:9:327
```
        assert abs(zero_state_error) < ENERGY_THRESHOLD, f"The zero state alchemical and nonalchemical energy absolute difference {abs(zero_state_error)} is greater than the threshold of {ENERGY_THRESHOLD}."
>       assert abs(one_state_error) < ENERGY_THRESHOLD, f"The one state alchemical and nonalchemical energy absolute difference {abs(one_state_error)} is greater than the threshold of {ENERGY_THRESHOLD}."
E       AssertionError: The one state alchemical and nonalchemical energy absolute difference 5.307491402817277e-06 is greater than the threshold of 1e-06.
E       assert 5.307491402817277e-06 < 1e-06
E        +  where 5.307491402817277e-06 = abs(-5.307491402817277e-06)
```
Just another one coming from our nightly CI

mikemhenry:
Another example https://github.com/choderalab/perses/pull/809#issuecomment-865523341

zhang-ivy:
[...]

PR #828, opened 2021-07-29
Fix RepartitionedHybridTopologyFactory tests to reduce stochastic failures

zhang-ivy:
I’ve added a fix to the RepartitionedHybridTopologyFactory tests to reduce stochastic failures — we now check the potential energy of the proposed positions before generating the `RepartitonedHybridTopologyFactory`. If the potential energy is greater than 1e8, we try to generate another proposal with energy < 1e8. We do this 5x, and if on the 5th time we still fail to generate a proposal with energy < 1e8, we raise an exception.

Addresses: https://github.com/choderalab/perses/issues/815


mikemhenry:
Looks like we are still getting a failure:
test_flattenedHybridTopologyFactory_energies
https://github.com/choderalab/perses/pull/828/checks?check_run_id=3196637909#step:9:353

mikemhenry:
I'm also worried that this strategy is just re-rolling the randomness of the way we are generating the atom map.

zhang-ivy:
@mikemhenry: The failure you linked above is in the nightly -- which I thought wasn't required for merging? Also upon further inspection of the failure, it looks like the code that is being run in that version doesn't contain my changes in this PR, check the lines below [here](https://github.com/choderalab/perses/pull/828/checks?check_run_id=3196637909#step:9:296) -- they don't contain the for loop I introduced

zhang-ivy:
> I'm also worried that this strategy is just re-rolling the randomness of the way we are generating the atom map.

I don't think its randomness in the atom map that's causing the stochastic failures, I think its randomness in the geometry engine's proposal order. The geometry engine by nature was designed to use randomness to generate new proposals, so I don't think that part would change. If anyone has any suggestions for alternative ways to solve this, feel free to share here! I just implemented a version of what @jchodera  suggested [here](https://github.com/choderalab/perses/pull/809#issuecomment-886260884)

jchodera (approved):
This is certainly a significant improvement!
We can continue to improve this to be more deterministic and thorough in future iterations.

zhang-ivy:
@mikemhenry : I think you are putting out a release today -- feel free to merge this and include it in the release if you see fit!

mikemhenry:
That sounds good, I will get it into the release today!

RE: that failure not containing the PR changes: I will investigate that since we should be testing the code in the PR, thanks for the great work @zhang-ivy

perses/tests/test_relative.py
```diff
@@ -711,10 +712,25 @@ def RepartitionedHybridTopologyFactory_energy_mutant(topology, chain, system, po
...
+        context.setPeriodicBoxVectors(*topology_proposal.new_system.getDefaultPeriodicBoxVectors())
+        context.setPositions(new_positions)
+        context.setVelocitiesToTemperature(temperature)
+        potential_energy = context.getState(getEnergy=True).getPotentialEnergy()
+
+        if potential_energy < 1e8 * unit.kilojoule_per_mole:
+            break
+        else:
+            if i == 4:
+                raise Exception("Was not able to generate a decent proposal after 5 iterations")
```
ijpulidos, line 733: This is nice! I believe a more readable way would be using a `for-else` statement here. I will be making the changes to accomplish that.

ijpulidos (commented):
Nice change, this will definitely help with the issue.

perses/tests/test_relative.py
ijpulidos: Fixed in https://github.com/choderalab/perses/pull/828/commits/867b40e48d249213a83349d28b4123fe7c8637e1

zhang-ivy: Cool, thanks! I didn't know about `for-else` statements!

zhang-ivy:
@mikemhenry @ijpulidos : It looks like one of the required tests is now failing. When I inspect the error, it again looks like this PR is not being tested properly. There should be a for loop starting around [here](https://github.com/choderalab/perses/pull/828/checks?check_run_id=3203785416#step:9:312).

mikemhenry:
It is because we need to copy this fix to another test as well

mikemhenry:
In the future we can refactor things to use fixtures so we can re-use bits of code for testing instead of copying and pasting the same bits in each test, this will make maintenance like this much easier!

zhang-ivy:
>It is because we need to copy this fix to another test as well

Ah good catch! I had forgotten about the other test 

>In the future we can refactor things to use fixtures so we can re-use bits of code for testing instead of copying and pasting the same bits in each test, this will make maintenance like this much easier!

This sounds great!

mikemhenry:
I copied the fix, so once it passes we should be good!

mikemhenry (approved):

PR #830, opened 2021-07-30
Add support for resuming with CLI/YAML

mikemhenry:


mikemhenry:
Not sure why it can't find the file, this works for me locally:
```
# Checkout PR
gh pr checkout 830
# Run the test
pytest -v -k test_cli_resume_repex
```

perses/app/setup_relative_calculation.py
```diff
@@ -874,5 +887,49 @@ def run(yaml_filename=None):
...
+            print(f"Comparing ligand {setup_options['old_ligand_index']} to {setup_options['new_ligand_index']}")
+            print(f"{phase} phase has a free energy of {free_energies[phase]}")
+
+    elif setup_options['fe_type'] == 'repex':
+        for phase in setup_options['phases']:
+            print(f'Running {phase} phase')
+            trajectory_directory = setup_options['trajectory_directory']
+            trajectory_prefix = setup_options['trajectory_prefix']
+
+            reporter_file = str(trajectory_directory)+'/'+str(trajectory_prefix)+'-'+str(phase)+'.nc'
```
ijpulidos, line 921: Same :) `f's{trajectory_directory}/{trajectory_prefix}-{phase}.nc'`

perses/app/setup_relative_calculation.py
```diff
@@ -874,5 +887,49 @@ def run(yaml_filename=None):
...
+            simulation = HybridRepexSampler.from_storage(reporter)
+            total_steps = setup_options['n_cycles']
+            run_so_far = simulation.iteration
+            left_to_do = total_steps - run_so_far
+            _logger.info(f"\t\textending simulation...\n\n")
+            simulation.extend(n_iterations=left_to_do)
+            _logger.info(f"\n\n")
+            _logger.info(f"\t\tFinished phase {phase}")
+    else:
+        raise("Can't resume")
```
ijpulidos, line 932: I am not familiar with the setup options and what does `fe_type` actually means, yet. But I do see in some places that there are options different to `sams` or `repex`. Can we safely assure that only these two are the ones to be used when resuming runs?

docs/changelog.rst
```diff
@@ -14,6 +14,8 @@ The full release history can be viewed `at the GitHub yank releases page <https:
 
 Bugfixes
 ^^^^^^^^
+- (PR `#830 <https://github.com/choderalab/perses/pull/830>`_)
+  Added support for resuming simulations from the CLI
```
ijpulidos, line 18: Maybe we can be a bit more specific? As I understand, this resumes simulations in some cases but not all of them, is that correct? Maybe we can specify that a bit.

ijpulidos (approved):
Nice work! It certainly helps when resuming from CLI/YAML files. However, we should keep in mind that maybe only changing and checking the iterations is not enough. What if the simulation crashes and/or other parameters are changed before trying to resume?

perses/app/setup_relative_calculation.py
```diff
@@ -681,6 +680,20 @@ def run(yaml_filename=None):
 
     _logger.info(f"Getting setup options from {yaml_filename}")
     setup_options = getSetupOptions(yaml_filename)
+
+    # The name of the reporter file includes the phase name, so we need to check each
+    # one
+    for phase in setup_options['phases']:
+        trajectory_directory = setup_options['trajectory_directory']
+        trajectory_prefix = setup_options['trajectory_prefix']
+        reporter_file = str(trajectory_directory)+'/'+str(trajectory_prefix)+'-'+str(phase)+'.nc'
```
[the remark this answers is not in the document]
jchodera: Yes! f-strings are strongly encouraged.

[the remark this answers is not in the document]
jchodera: For maximum compatibility, we would use `os.path.join`, but we have no need to support `win` platforms here.

perses/app/setup_relative_calculation.py
jchodera: We can eventually unify this code, since the only thing that is different (aside from debug output) is the class name of the `MultiStateSampler` subclass. But for now, this is fine---we'll be reworking this entirely in future milestones, so it's not necessary to spend much more effort on it.

The [multistate samplers](https://openmmtools.readthedocs.io/en/0.18.1/multistate.html) in openmmtools are intended to be drop-in replacements for each other, and share nearly identical APIs (except for some small algorithmic differences). We can eventually add `NonequilibriumSwitchingSampler` too.

perses/tests/test_resume.py
```diff
@@ -64,6 +64,17 @@ def test_cli_resume():
             yaml.dump(y_doc, outfile)
         subprocess.run(["perses-relative", "test.yml"])
```
jchodera, line 65: Am I missing something, or does this only run perses _once_? We need to run it, stop it early (or extend it after it finishes), and then run it again.

jchodera (changes requested):
The test for whether repex can be resumed no longer actually tests resuming: It just runs perses once. We need to run it once, stop it early (or extend it after it finishes), and then run it again, and demonstrate that works gracefully. We should fix the test, make sure it works, then merge.

mikemhenry:
I'll use an f-string, but I agree that using pathlib stuff is best when dealing with paths since it can make sure the separators are correct and while paths are string like, its best to treat them more a separate thing. I just was doing this with the least amount of effort so I coped the line here https://github.com/choderalab/perses/blob/02df62f7fc7c93e56e74eb956790e73a41fbb761/perses/app/setup_relative_calculation.py#L616

I'll get the test working as well, it passed for me locally but then failed on GHA so I commented it out so if we needed this *today* it could be merged in. I'm working on fixing the tests now :)

perses/app/setup_relative_calculation.py
mikemhenry: Those are the only two types I belive that we can have with the yaml so it is safe to just assume we will have to handle those two cases.

docs/changelog.rst
mikemhenry: I don't think there are any simulations we can't resume, but I will add that the support is limited/fragile

jchodera:
@mikemhenry : One issue could be that the [OpenEye license files are not being found](https://github.com/choderalab/perses/pull/830/checks?check_run_id=3215955594#step:9:430), which I think means that the `perses-relative` entry point is failing to return an error code or we are failing to check for it after calling `subprocess.run()`: 
```
2021-08-01 23:58:52,344:(0.00s):relative_setup:Setting non bonded method to PME
2021-08-01 23:58:52,344:(0.00s):relative_setup:Handling files for ligands and indices...
2021-08-01 23:58:52,344:(0.00s):relative_setup:Detected .sdf format.  Proceeding...
LICENSE: Could not open license file specified by OE_LICENSE environment variable "/home/runner/oe_license.txt"
LICENSE: Could not open license file "oe_license.txt" in local directory
LICENSE: N.B. OE_DIR environment variable is not set
LICENSE: No product keys!
LICENSE: No valid license found for oechem
```
I think this means this test is failing to generate any output the first time because of a failure to find the license file in the `subprocess.run()` environment.

ijpulidos:
@mikemhenry I think the tests are expecting this `cdk2_repex_hbonds/cdk2-vacuum.nc` file that I cannot see where it comes from. That may be something you have only on your local tree, hence why it is passing there? Just a guess.

mikemhenry:
@ijpulidos it is created when we run the yaml file
@jchodera good catch -- I thought that was a red herring since the other tests seem to run fine and I expected them to blow up if there really was an issue with the license file -- looking into that now

mikemhenry:
AHH I forgot how subprocess will have a different env, I bet that is it @jchodera, testing now! 
This would explain why it works locally, I have the license file in the spot where it expects to be and not set via an envar.

mikemhenry:
I am also going to quickly add `LOGLEVEL = os.environ.get("LOGLEVEL", "INFO").upper()` in this PR as well since the log level is hard coded and I remember relay having an issue with that when we told them to more information. When using the API you can set the log level but there wasn't a way to do that with the CLI until now.

mikemhenry:
If this last fix doesn't work, I will copy the license into the local dir where these tests are running.

mikemhenry:
:tada: okay it is passing now! This is what I get for doing TDD locally instead of on GHA :shrug:

mikemhenry:
@ijpulidos @jchodera would you mind taking a quick look again? Really the only changes were made to the testing, but I did add functionality for reading `LOGLEVEL` from the environment. Once this is merged we can cut that release!

docs/changelog.rst
```diff
@@ -14,6 +14,10 @@ The full release history can be viewed `at the GitHub yank releases page <https:
 
 Bugfixes
 ^^^^^^^^
+- (PR `#830 <https://github.com/choderalab/perses/pull/830>`_)
+  Added limited support for resuming simulations from the CLI. 
+  Assumes simulations are only going to be resumed from the production step and not equilibration step.
+  ``LOGLEVEL`` can now be set with an environmental variable when using the CLI tool.
```
ijpulidos, line 20: Maybe we can say something about extending the simulation `n_cycles` now? I think that's a good thing to show.

perses/app/setup_relative_calculation.py
```diff
@@ -298,8 +299,7 @@ def getSetupOptions(filename):
         _logger.info(f"\t'softcore_v2' not specified: default to 'False'")
 
     _logger.info(f"\tCreating '{trajectory_directory}'...")
-    assert (not os.path.exists(trajectory_directory)), f'Output trajectory directory "{trajectory_directory}" already exists. Refusing to overwrite'
-    os.makedirs(trajectory_directory)
+    os.makedirs(trajectory_directory, exist_ok=True)
```
ijpulidos, line 302: What happens if the directory already exists? Do contents get overwritten? If that's the case I think maybe we shouldn't do that or at least we should warn the users. What do you think?

ijpulidos (approved):
Nice changes, good to see the debugging levels capabilities.

perses/app/setup_relative_calculation.py
mikemhenry: If it exists, it does not get overwritten. That function acts like the UNIX utility `mkdir` where if the directory already exists, you get an error ie:
```
$ mkdir foo
$ mkdir foo
mkdir: cannot create directory ‘foo’: File exists
```

docs/changelog.rst
mikemhenry: Good point!

perses/app/setup_relative_calculation.py
ijpulidos: Yeah, I agree. So, I don't know if I'm missing something else, but when the directory exists there wouldn't be an error and the contents in it then could get overwritten?

jchodera (approved):
Thanks for fixing this!

perses/app/setup_relative_calculation.py
mikemhenry: It won't overwrite the contents:
```
$ ls foo
bar
$ python
Python 3.8.5 (default, Sep  4 2020, 07:30:14) 
[GCC 7.3.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.makedirs("foo", exist_ok=True)
>>> 
$ ls foo/
bar
```
Also the docs don't mention that the contents will be overwritten 
https://docs.python.org/3/library/os.html#os.makedirs

ijpulidos: I will be making a new issue to discuss about this a bit further. What if the user uses the same directory name without realizing it already exists?

PR #834, opened 2021-08-04
Enable protein mutations involving nonstandard protonation state aas

### LINKED ISSUE #831: Enable protein mutation transformations for amino acids with nonstandard protonation states
Since we are failing to achieve convergence for K27A and D39A in barnase:barstar, @jchodera suggested that I run these mutations with alternate protonation states (LYN instead of LYS and ASH instead of ASP). If we can demonstrate that the DDG is sensitive to the protonation state, then we can say that our lack of convergence for these mutations could likely be solved with constant pH simulations. 

Currently, the `PointMutationEngine`/`PolymerProposalEngine` do not support nonstandard amino acids. When constructing the atom map, the PolymerProposalEngine relies on generating an oemol from one of the PDBs in [perses/data/amino_acid_templates](https://github.com/choderalab/perses/tree/master/perses/data/amino_acid_templates).  The code for this is [here](https://github.com/choderalab/perses/blob/222e1c7003bcd5b8a8e0c9f28396301e5e2bdaaf/perses/rjmc/topology_proposal.py#L1228-L1232)

To enable protein mutation transformations involving LYN and ASH, is is sufficient to add PDBs for LYN and ASH in `amino_acid_templates`? If so, how were these PDBs generated?
Are there other changes (to topology_proposal.py or geometry.py) necessary to handle these types of amino acids?

Tagging @dominicrufa  and @jchodera  for feedback!

zhang-ivy:
## Description
Enable protein mutation transformations involving nonstandard amino acids, specifically: ASH, GLH, HID, HIE, HIP, LYN
- LYN is deprotonated LYS (neutral)
- ASH is protonated ASP (neutral)
- GLH is protonated GLU (neutral)
- HIE is histidine protonated at N-epsilon (neutral)
- HID is histidine protonated at N-delta (neutral)
- HIP is histidine protonated at both N-delta and N-epsilon (+1 charge)

Note that if HIS is specified, we use the HIE protonation state.

Changes:
- Enable `perses.app.relative_point_mutation_setup.PointMutationExecutor` to take in an optional argument that specifies a nonstandard amino acid name for the old residue. 
    - If there are nonstandard residues, the input PDBs should contain the standard residue name but the atoms/positions should correspond to the nonstandard residue. 
    - The `PointMutationExecutor` will generate the old and new systems with the standard amino acid name (as OpenMM requires standard amino acid names even for nonstandard amino acids) and then it will modify the old residue name to the nonstandard name specified. 
    - From there on, the nonstandard name will be used in subsequent calls to generate a topology proposal, geometry proposals, and hybrid topology factories. 
    - Note: If the user wants to mutate to a nonstandard amino acid, simply supply the nonstandard amino acid name as the `proposed_residue` argument, per normal (nothing changed here)
- Expanded `ring_amino_acids` in `relative_point_mutation_setup.py` to include HID, HIE, HIP
- Added PDB templates (generated in PyMOL) for the aforementioned nonnstandard amino acids in `perses/data/aminio_acid_templates` -- these are used to generate oemols of the old and new residues
- Expand `perses.rjmc.topology_proposal.PolymerProposalEngine._aminos` to include `['ASH', 'GLH', 'HID', 'HIE', 'HIP', 'LYN']`. Changed `PolymerProposalEngine._positive_aminos` to be `['ARG', 'HIP', 'LYS']`
- Deleted lines in `topology_proposal.py` that force HID/HIE/HIP to HIS  -- previously we didn't have HID/HIE/HIP PDB templates, so we had to use the HIS template, but we don't need this behavior anymore
- Deleted unused attribute `self._aminos_3letter_to_1letter_map` from `topology_proposal.py`
- Expanded `test_topology_proposal.py`'s `test_mutate_from_alanine` to include the aforementioned nonstandard amino acids 
- Removed the automatic conversion of HID/HIE to HIS in `test_topology_proposal.py`'s `test_mutate_from_every_amino_to_every_other`

## Motivation and context

Resolves #831

## How has this been tested?

Expanded `test_topology_proposal.py`'s `test_mutate_from_alanine` to include the aforementioned nonstandard amino acids.

## Change log

```

```


zhang-ivy:
@mikemhenry : I suspect `test_mutate_from_alanine` is failing in GHA because the PDB files I added to `perses/data/amino_acid_templates` aren't being included in the repo that is being tested -- how can I check whether the files (e.g. `HID.pdb`, `LYN.pdb`, etc) are being included?

mikemhenry:
It looks like this is the error, I would expect it to be a file not found if it was missing the files:
```
>   new_res_to_oemol_map = {atom.index: proposed_oemol.GetAtom(oechem.OEHasAtomName(atom.name)).GetIdx() for atom in new_res.atoms()}
E   AttributeError: 'NoneType' object has no attribute 'GetIdx'

perses/rjmc/topology_proposal.py:1214: AttributeError

```
It looks like the method being used to read in the templete is correct:
```
current_residue_pdb_filename = resource_filename('perses', os.path.join('data', 'amino_acid_templates', f"{old_res_name}.pdb"))
```

So I don't really have any reason to believe those files are not being included correctly.

zhang-ivy:
@mikemhenry : Yes, I've seen that error locally before and its misleading because 
```
resource_filename('perses', os.path.join('data', 'amino_acid_templates', f"{old_res_name}.pdb"))
```
will not throw an error even if the file does not exist. I just double checked this locally with `old_res_name = 'HIT'` , and it does not throw an error.  

I'm almost certain that's it because the files I added to `perses/data/amino_acid_templates` are missing.

> If you want to test, you can try and adding some print statements like print(os.path.isfile(current_residue_pdb_filename)) and if it prints True then we know the files are being included.

Was hoping there would be a way to check this without rerunning the whole workflow (its very slow). But I'll try this!

mikemhenry:
You can try it locally like this:

`pytest -v -k test_mutate_from_alanine`

mikemhenry:
Something I've also done is change the CI to only run one test while I debug, like I did with `.github/workflows/CI.yaml` on this commit https://github.com/choderalab/perses/pull/830/commits/ddee79b4f8dd44df2efabf2e4ead35f294619f62

zhang-ivy:
Ah, I found the problem, its actually not that the PDB files I added are missing. There is a difference in the residue templates in OpenMM vs Perses for the `LYN` amino acid. Fixing now.

zhang-ivy:
@mikemhenry : The tests should be passing now, could you review and merge? 
@dominicrufa  already let me know via slack that he reviewed this and approves, but if you could formally "approve" that would be great!

perses/rjmc/topology_proposal.py
```diff
@@ -1216,18 +1200,10 @@ def _construct_atom_map(self,
...
-        if new_res_name in his_templates:
-            new_res_name = 'HIS'
-        else:
-            pass
-
         current_residue_pdb_filename = resource_filename('perses', os.path.join('data', 'amino_acid_templates', f"{old_res_name}.pdb"))
         proposed_residue_pdb_filename = resource_filename('perses', os.path.join('data', 'amino_acid_templates', f"{new_res_name}.pdb"))
-
+        print(f"is the old res {old_res_name} pdb present? ", os.path.isfile(current_residue_pdb_filename))        
+        print(f"is the new res {new_res_name} pdb present? ", os.path.isfile(current_residue_pdb_filename))
```
dominicrufa, line 1206: looks good, but before merge, can you make this a logger comment rather than a print, plz

dominicrufa (commented):
this looks good; this is just pending a fix of the error in the test i think

perses/rjmc/topology_proposal.py
zhang-ivy: oops sorry! fixing now

perses/rjmc/topology_proposal.py
```diff
@@ -910,6 +910,10 @@ def _identify_differences(self, topology, residue_map):
             # template_atom_names : dict, key : template atom index, value : template atom name
             template_atom_names = {}
             for atom in template.atoms:
+                if replace_with == 'LYN' and atom.name == 'HZ3': # Rename HZ3 to HZ1 s.t. the topology matches the naming in amino_acid_templates/LYN.pdb
+                    atom.name = 'HZ1'
+                    template.atomIndices['HZ1'] = template.atomIndices['HZ3']
+                    del template.atomIndices['HZ3']
```
mikemhenry, line 916: Will this ever find more than one match? Deleting things as you iterate can cause issues, it might be better to first loop over the atoms to rename them.
I don't understand why `template.atomIndices['HZ1'] = template.atomIndices['HZ3']` is in the loop since it operates on `template.atomIndices` and not `atom`, plus then if you find another atom with that name then you will get a key error.
Does this work?

```suggestion
                if replace_with == 'LYN' and atom.name == 'HZ3': # Rename HZ3 to HZ1 s.t. the topology matches the naming in amino_acid_templates/LYN.pdb
                    atom.name = 'HZ1'
            template.atomIndices['HZ1'] = template.atomIndices['HZ3']
            del template.atomIndices['HZ3']
```

mikemhenry: Or if the logic of this only needs to happen then `replace_with == 'LYN'` might have to be moved

zhang-ivy: This logic should only happen for `LYN`. Basically the problem here is that the `LYN` amber template atoms do match the atoms in `LYN.pdb` (generated by pymol). Since LYN is deprotonated LYS, this means that terminal carbon on the LYN side chain should only contain 2 hydrogens, not 3. In the amber template, `LYN` contains HZ2 and HZ3, whereas in `LYN.pdb`, it contains HZ1 and HZ2. So here I was trying to rename the HZ3 atom (there's only one of them) in the amber template for LYN. I also need to update the atom indices dictionary (key: atom name, value: atom index), because otherwise the dictionary will contain the wrong atom name. I think my original implementation works better than your suggestion, because your suggestion allows the `template.atomIndices` dict to be changed for the HZ1 atom in any residue (which is not the behavior we want). Hopefully this clarifies things, but let me know if you still think some of the logic can be improved upon

mikemhenry: How does having `template.atomIndices['HZ1'] = template.atomIndices['HZ3']` in the loop ensure that HZ1 atom is changed only for the LYN residue? Since we are only looping over `for atom in template.atoms` I would expect that it would be okay if this was outside of the loop.

Okay I looked at more context and see now that the check  `replace_with == 'LYN'` is what is guarding which residue we are changing. 

@zhang-ivy  would you mind reorganizing this slightly so it is more clear when we are doing things for only one residue? something like this (I can't suggest changes that don't happen in the PR which is why I can't use that feature here).

``` python
for k, (residue, replace_with) in enumerate(residue_map):
            # Load residue template for residue to replace with
            if replace_with =='HIS':
                replace_with = 'HIE'
            template = self._templates[replace_with]
            # template_atom_names : dict, key : template atom index, value : template atom name
            template_atom_names = {}
            if replace_with == 'LYN':
                for atom in template.atoms:
                    if atom.name == 'HZ3': # Rename HZ3 to HZ1 s.t. the topology matches the naming in amino_acid_templates/LYN.pdb
                        atom.name = 'HZ1'
                 template.atomIndices['HZ1'] = template.atomIndices['HZ3']
                 del template.atomIndices['HZ3']
```
I think something like that would work? Moving the `if` logic to be as high as possible to modify things per-residue makes it more clear what is going on.

zhang-ivy: The reason why I didnt implement it like your suggestion above is because I was trying to avoid adding an extra for loop that iterates over the template.atoms, since [there is already a for loop](https://github.com/choderalab/perses/blob/master/perses/rjmc/topology_proposal.py#L932-L933) that iterates over the atoms. 
If you feel strongly that your suggestion is more clear, we I can replace my original implementation with yours, however I think your implementation is a bit redundant.

zhang-ivy: Or we could keep my implementation and I can add better comments that clarify what is happening the code? 
Something like `iterate over atoms in the mutant residue template. if mutant is LYN, we need to edit one of the atom names"

zhang-ivy: Mike and I moved this discussion. We settled on the changes in the following commit.

zhang-ivy:
@mikemhenry : I think this is ready for merge!

mikemhenry (approved):

PR #835, opened 2021-08-04
Write out YAML file after parsing

### LINKED ISSUE #817: Write out YAML after parsing
Currently we don't require users to specify all options in the YAML file and instead we provide some defaults in accordance to best practices. Some options are also set dynamically depending on what other options a users sets. After we parse the YAML file and set all the options, we should write out the full YAML file. This will make it easier for users to see exactly which options are being used as well as to improve reproducibility in case our default options change later.

mikemhenry:
## Description

Write out YAML file after parsing.

## Motivation and context

In order to improve reproducibility and quickly see which options were used in the simulation, it would be nice to be able to inspect the YAML that has all the options set as parsed.

Resolves #817 

## How has this been tested?

Tested on GHA and locally manually.

## Change log

```
Write out YAML file after all options are parsed and set. Saved as YAML original file name + date + time. 
```

### DIFF
--- a/perses/app/setup_relative_calculation.py
+++ b/perses/app/setup_relative_calculation.py
@@ -299,6 +299,10 @@ def getSetupOptions(filename):
         _logger.info(f"\t'softcore_v2' not specified: default to 'False'")
 
     _logger.info(f"\tCreating '{trajectory_directory}'...")
+
+    if not 'rmsd_restraint' in setup_options:
+        setup_options['rmsd_restraint'] = False
+
     os.makedirs(trajectory_directory, exist_ok=True)
 
 
@@ -327,8 +331,6 @@ def run_setup(setup_options, serialize_systems=True, build_samplers=True):
     #   such as deferring to defaults for modules we call unless the user
     #   chooses to override them.
 
-    if not 'rmsd_restraint' in setup_options:
-        setup_options['rmsd_restraint'] = False
 
     if 'use_given_geometries' not in list(setup_options.keys()):
         use_given_geometries = False
@@ -680,7 +682,10 @@ def run(yaml_filename=None):
            _logger.critical(f"You must specify the setup yaml file as an argument to the script.")
 
     _logger.info(f"Getting setup options from {yaml_filename}")
-    setup_options = getSetupOptions(yaml_filename)
+    from types import MappingProxyType
+
+    setup_options_temp = getSetupOptions(yaml_filename)
+    setup_options = MappingProxyType(setup_options_temp)
 
     # The name of the reporter file includes the phase name, so we need to check each
     # one
--- a/perses/app/setup_relative_calculation.py
+++ b/perses/app/setup_relative_calculation.py
@@ -5,6 +5,7 @@ import os
 import sys
 import simtk.unit as unit
 import logging
+from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


mikemhenry:
It isn't actually because it is an object, the parsing code doesn't handle the case where `complex_box_dimensions` is set to `None` or (`null`) but checks to see if the key exits and if it does, assumes it can be cast as a float.

``` python
    if 'complex_box_dimensions' not in setup_options:
        setup_options['complex_box_dimensions'] = None
    else:
        setup_options['complex_box_dimensions'] = tuple([float(x) for x in setup_options['complex_box_dimensions']])
```

mikemhenry:
So I'll just need to make the parsing logic a bit more robust.

ijpulidos:
@mikemhenry I actually think having all the values in the parsed YAML is a good idea for replicability (even the ones inferred by the software). If we don't have all of these in the parsed YAML and if we end up changing the defaults in the future we then would get different results for the same YAML file, which I think is not something we want.

Another solution would be to just not try to put these `None` when the original input YAML file doesn't have them, but actually just trust the class defaults and output the parsed YAML after it has passed through the classes and such. So, for example for the `setup_options['complex_box_dimensions']` case you would just have to overwrite it if it is in the input YAML file, else just use the class defaults, i.e. just:
```python
    if 'complex_box_dimensions' in setup_options:
        setup_options['complex_box_dimensions'] = tuple([float(x) for x in setup_options['complex_box_dimensions']])
```
or something like that.

Of course this means we have to be sure the classes have actual sane and working defaults and this may be a hard task.

ijpulidos:
Hmm, actually that code snippet wouldn't work as well, maybe you need to check also that they are not `None` up there, but the idea of having the default behavior be determined by the defaults in the classes still holds in my opinion.

jchodera:
> So even if I drop the null values, we don't lose any information that was in the original and then we can re-run a parsed yaml file.

As long as we can re-parse the YAML file, that works for now! We'll be replacing all of this functionality when we bring the new API online.

mikemhenry:
Yes, I want to do the least amount of work possible on this since it is due for an overhaul. Ready for review now @ijpulidos and @jchodera

ijpulidos (changes requested):
Nice! It works, even though I think this way of dealing with setup options from CLI can be done in a better way. Hopefully something for future releases.

I think we do require a test for this, though. Something that starts a sim from a parsed YAML.

ijpulidos:
Is it crazy to ask also for a test that verifies that the parsed YAML has *all* the possible parameters/options? Thinking again about reproducibility.

jchodera:
> Is it crazy to ask also for a test that verifies that the parsed YAML has all the possible parameters/options? Thinking again about reproducibility.

I think all we need right now is "good enough" because all this functionality will be replaced in an [upcoming milestone](https://github.com/choderalab/perses/pulls?q=is%3Aopen+is%3Apr+milestone%3A%220.11.0+New+API+for+relative+free+energy+calculations%22). What about creating a new issue to discuss what the _new_ YAML functionality should look like with the new API, and making sure we spec out exactly what behavior we want for that? If we design it right, it will be simple to implement and achieve the desired behavior while not compromising flexibility. Tests could, for example, use introspection to regress over all available options.

jchodera (approved):
Good enough for now!

PR #838, opened 2021-08-05
Feat/add docker support

### LINKED ISSUE #832: Maintain official docker container
We should bundle an official `Dockerfile` in the repo and host docker builds on dockerhub.

mikemhenry:
## Description

This PR adds a docker file that we will maintain as an official docker image.

## Motivation and context

This is a first draft, we can optimize the container size and build time later. I would like to get it into our users hands so we can iterate on it.

Resolves #832 

## How has this been tested?

I've tested this image locally by building it, then starting a shell in the container, cloning the `perses` repo, then running pytest. You should be able to play around with it, `docker pull choderalab/perses:0.9.1` (uploading the image may take some time, so if you get image not found, that's why).

## Change log

```
Official Docker image hosted on docker hub `docker pull choderalab/perses:0.9.1`
```


docker/Dockerfile-nvidia-10
```diff
@@ -0,0 +1,32 @@
...
+    micromamba install -n "$ENV_NAME" -c conda-forge -c openeye --yes perses==0.9.1 openeye-toolkits google-cloud-storage && \
+    micromamba clean -a
+
+
+RUN mkdir /perses
+RUN mkdir /data
+ENV OE_LICENSE=/perses/oe_license.txt
+
+ENV PATH $PATH:/perses
+ENV OPENMM_CPU_THREADS="2"
```
jchodera, line 32: Is this a reasonable choice given that most AWS instances like [`p3.2xlarge`](https://aws.amazon.com/ec2/instance-types/p3/) have 8 vCPUs for every 1 GPU? 

We should also document in the `README.md` what the recommended instance types are for AWS and GCE.

docker/README.md
```diff
@@ -0,0 +1,8 @@
+
+The open eye license file is not included and must be passed in as a mount point.
+I have an `oe_license.txt` file in `/home/mmh/.openeye/`.
+The container expects `oe_license.txt` in `/perses/` because we set `ENV OE_LICENSE=/perses/oe_license.txt` in the Dockerfile.
+
+To do that use the `--mount` option.
+For example and to to test:
+docker run --rm --mount type=bind,source=/home/mmh/.openeye/,target=/perses/,readonly mmh42/perses:0.9.1  python -c "import openeye; assert openeye.oechem.OEChemIsLicensed(), 'OpenEye license checks failed!'"
```
jchodera, line 8: Can you change `/home/mmh` to `$HOME`?

docker/Dockerfile-nvidia-10
```diff
@@ -0,0 +1,32 @@
...
+    ./bin/micromamba shell init -s bash -p "$MAMBA_ROOT_PREFIX" && \
+    echo "micromamba activate $ENV_NAME" >> "$BASH_ENV" && \
+    chmod -R a+rwx "$HOME" "$BASH_ENV" "$MAMBA_ROOT_PREFIX" && \
+    micromamba install -n "$ENV_NAME" -c conda-forge -c openeye --yes perses==0.9.1 openeye-toolkits google-cloud-storage && \
+    micromamba clean -a
+
+
+RUN mkdir /perses
+RUN mkdir /data
+ENV OE_LICENSE=/perses/oe_license.txt
```
jchodera, line 29: Why not change this to `/openeye/oe_license.txt` to be clearer? Would have to update the `README` as well.

docker/Dockerfile-nvidia-10
```diff
@@ -0,0 +1,32 @@
...
+ENV BASH_ENV="$HOME/.bashrc"
+ENV PATH "$MAMBA_ROOT_PREFIX/bin:$PATH"
+
+
+RUN wget -qO- https://micromamba.snakepit.net/api/micromamba/linux-64/latest | tar -xvj bin/micromamba && \
+    mkdir -p "$HOME" && \
+    ./bin/micromamba shell init -s bash -p "$MAMBA_ROOT_PREFIX" && \
+    echo "micromamba activate $ENV_NAME" >> "$BASH_ENV" && \
+    chmod -R a+rwx "$HOME" "$BASH_ENV" "$MAMBA_ROOT_PREFIX" && \
+    micromamba install -n "$ENV_NAME" -c conda-forge -c openeye --yes perses==0.9.1 openeye-toolkits google-cloud-storage && \
```
jchodera, line 23: What's behind `google-cloud-storage`?

docker/Dockerfile-nvidia-10
```diff
@@ -0,0 +1,32 @@
+FROM nvidia/cuda:10.1-runtime-ubuntu18.04
```
jchodera, line 1: Why CUDA 10.1 and not 11.x?

jchodera (changes requested):
This is great! 

Can we document the `Dockerfile` with some comments that make it clearer what options might be changed and why, and why some weird choices were selected?

mikemhenry:
@jchodera Thanks for the feedback! Those weird choices are copied over from a docker file I was using to troubleshoot a different issue -- I'll make this much more generic and include in the README how to add customizations that someone might need.

docker/README.md
mikemhenry: Done! I used `~` originally and docker complained, but `$HOME` worked!

docker/Dockerfile-nvidia-10
mikemhenry: Done!

docker/Dockerfile-nvidia-10
mikemhenry: It was just left over from a copy and paste, removed it!

docker/Dockerfile-nvidia-10
mikemhenry: At one point I was having issues with CUDA 11 and the openmm version I was using, 11.1 works just fine and I've updated the container.

docker/Dockerfile-nvidia-10
mikemhenry: I've removed that env argument so openmm will use all the threads it can.

ijpulidos (approved):
Nice work, this is going to be very helpful!

docker/Dockerfile-nvidia-11
```diff
@@ -0,0 +1,43 @@
...
+# Instead of using conda which is slow,
+# We will use micromamba which is fast
+# and quick to install and works like conda
+# See https://github.com/mamba-org/mamba#micromamba for more details
+RUN wget -qO- https://micromamba.snakepit.net/api/micromamba/linux-64/latest | tar -xvj bin/micromamba && \
+    mkdir -p "$HOME" && \
+    ./bin/micromamba shell init -s bash -p "$MAMBA_ROOT_PREFIX" && \
+    echo "micromamba activate $ENV_NAME" >> "$BASH_ENV" && \
+    chmod -R a+rwx "$HOME" "$BASH_ENV" "$MAMBA_ROOT_PREFIX" && \
+    micromamba install -n "$ENV_NAME" -c conda-forge -c openeye --yes perses==0.9.1 openeye-toolkits && \
```
jchodera, line 37: @jaimergp may know the answer here, but we may want to install the version of OpenMM _without_ the CUDA Toolkit if the docker image already _has_ the toolkit? We need to be careful about keeping versions consistent.

This is likely just a size optimization (?) so it could be punted to a future iteration.

jchodera (approved):
Looks great! Just flagged one issue.

docker/Dockerfile-nvidia-11
jaimergp: You can probably force remove it later as long as it matches. I'd add `cudatoolkit==x.y` to the package list. Also, you can add that experimental channel for the cudatoolkit shims so nothing is downloaded to begin with.

mikemhenry: I will play around and see

mikemhenry:
I was able to get things to work by using the conda shim stuff to prevent the condatoolkit from being installed. Later we can work on making it smaller, perhaps with a multi-layer build.

PR #839, opened 2021-08-07
Store new and old positions and maps

### LINKED ISSUE #752: hybrid_{complex,solvent}.pdb written by FAH generation code is missing CONECT records for ligands B state
The current `hybrid_{complex,solvent}.pdb` written by the FAH generation code appears to be missing `CONECT` records for the ligand B state. It also incorrectly writes the `MOL` residue as `ATOM` instead of `HETATM`.

@dominicrufa : Any idea where I would fix this? It would seem to be an issue with the OpenMM topology we use for writing if we're writing via OpenMM's `PDBFile`.

ijpulidos:
## Description
This allow the serialization of new and old positions in pdb files as well as atom mappings in numpy object file, when creating hybrid topologies. 

## Motivation and context

Resolves #752 
## How has this been tested?
Tested using [minimal-neq-sprint.tar.gz](https://github.com/choderalab/perses/files/6948343/minimal-neq-sprint.tar.gz) by running `python 05-prepare-single-transformation.py 0`

## Change log
Addresses (issue #752 ) by storing positions and maps for new and old molecules when creating a hybrid topology.

```

```

### DIFF
--- a/perses/app/fah_generator.py
+++ b/perses/app/fah_generator.py
@@ -18,6 +18,8 @@ from simtk import openmm
 import logging
 
 import datetime
+
+
 class TimeFilter(logging.Filter):
     def filter(self, record):
         try:
@@ -133,7 +135,7 @@ def relax_structure(temperature,
                     positions,
                     nequil=1000,
                     n_steps_per_iteration=250,
-                    platform_name='OpenCL',
+                    platform_name='CPU',
                     timestep=4.*unit.femtosecond,
                     collision_rate=90./unit.picosecond,
                     **kwargs):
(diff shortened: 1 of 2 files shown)


perses/app/fah_generator.py
```diff
@@ -133,7 +135,7 @@ def relax_structure(temperature,
                     positions,
                     nequil=1000,
                     n_steps_per_iteration=250,
-                    platform_name='OpenCL',
+                    platform_name='CPU',
```
jchodera, line 138: We probably shouldn't change the default platform here, since we should still prefer the GPU if possible.
This may need to be addressed more generally later by allowing OpenMM to simply use the fastest platform by default rather than specifying the platform, but this is a bit complex since we need at least `mixed` precision, which is not the default platform property.

perses/app/fah_generator.py
```diff
@@ -409,17 +411,33 @@ def run_neq_fah_setup(ligand_file,
...
 
         import mdtraj as md
-        top = htfs[phase].hybrid_topology
-        np.save(f'{dir}/hybrid_topology', top)
-        traj = md.Trajectory(pos, top)
-        traj.remove_solvent(exclude=['CL', 'NA'], inplace=True)
-        traj.save(f'{dir}/hybrid_{phase}.pdb')
-
-        #lastly, make a core.xml
+        # Save old trajectory
```
jchodera, line 419: Perhaps we could change
```python 
# Save old trajectory
```
to
```python
# Store initial configuration/topology for old system sliced form hybrid system as PDB file
```
and same with the new one.

jchodera (approved):
This is great! Just suggested a bit of cleanup.

zhang-ivy (approved):
LGTM!

PR #840, opened 2021-08-09
Minor improvements to point mutation executor

zhang-ivy:
## Description

Here, I've included some minor fixes to the `PointMutationExecutor`:
1. Make sure reverse geometry proposal happens right after forward geometry proposal.
2. Add naked charge fix
3. Fix `complex_positions` formatting issue 
See details below.

## Motivation and context

Changes:
1. Move the line for generating the reverse geometry proposal s.t. it is directly after the forward proposal line. We want to do this to make sure that the forward and reverse proposals are executed before the counterion code, otherwise we end up running the forward proposal using the unmodified new system and the reverse proposal using the modified (by counterion code) new system. 
2. Add a fix that adds padding if there are naked charges. John suggested this as amber forcefields allow for naked charges, e.g. on hydroxyl hydrogens
3. Resolve a formatting problem for `complex_positions` -- When visualizing the complex positions after adding solvent, the protein was translated away from the solvent. This is because `complex_positions` is a numpy array, but openmm stores positions as Vec3 objects. When calling `modeller.addSolvent()`, there is likely a translation step that moves the solvent to the protein, but this step can only be done when the positions are in the Vec3 format. The solution here is to convert `complex_positions` to Vec3 format before adding solvent. 

## How has this been tested?

As long as [this test](https://github.com/choderalab/perses/blob/master/perses/tests/test_relative_point_mutation_setup.py) passes, we should be fine. 

To make sure change 2 is working correctly, I manually inspected the PDB in PyMOL to make sure the protein positions are no longer translated away from solvent.

## Change log

```

```


perses/app/relative_point_mutation_setup.py
```diff
@@ -251,13 +257,29 @@ def __init__(self,
...
+                    nb_force = force_dict['NonbondedForce']
+                    for i in range(nb_force.getNumParticles()):
+                        charge, sigma, epsilon = nb_force.getParticleParameters(i)
+                        if sigma == 0*unit.nanometer:
+                            sigma = 0.06*unit.nanometer
+                            nb_force.setParticleParameters(i, charge, sigma, epsilon)
+                        if epsilon == 0*unit.kilojoule_per_mole:
+                            epsilon = 0.0001*unit.kilojoule_per_mole
+                            nb_force.setParticleParameters(i, charge, sigma, epsilon)
+
```
dominicrufa, line 273: this is primarily for hydroxyl hydrogens, right?

dominicrufa (approved):

perses/app/relative_point_mutation_setup.py
zhang-ivy: That is one of main examples, yes. I'm not sure if there are other examples of naked charges

mikemhenry:
Looks like the tests are throwing some NaNs

```
 >           raise SimulationNaNError(message)
E           openmmtools.multistate.utils.SimulationNaNError: Propagating replica 0 at state 0 resulted in a NaN!
E           The state of the system and integrator before the error were saved in /tmp/pytest-of-runner/pytest-0/popen-gw1/test_resume_protein_mutation_n0/nan-error-logs

/usr/share/miniconda/envs/test/lib/python3.9/site-packages/openmmtools/multistate/multistatesampler.py:1234: SimulationNaNError
```

mikemhenry:
@zhang-ivy 
So the tests that are failing:

```
FAILED perses/tests/test_resume.py::test_resume_protien_mutation_with_checkpoint
FAILED perses/tests/test_resume.py::test_resume_protein_mutation_no_checkpoint
```

are ones that I wrote 

https://github.com/choderalab/perses/blob/master/perses/tests/test_resume.py#L158
https://github.com/choderalab/perses/blob/master/perses/tests/test_resume.py#L222

But I don't see anything weird with tests, and they are passing on master without these modifications, so something is weird. Before you dig too deep, check to make sure the tests look like they are using the classes as you would expect them to.

zhang-ivy:
@mikemhenry I fixed the changes in this PR s.t. the tests are passing for me locally now. Please review/merge!
Also btw there was a typo in `perses/tests/test_resume.py::test_resume_protien_mutation_with_checkpoint` -- I renamed `protien` to `protein` in the last commit

perses/app/relative_point_mutation_setup.py
```diff
@@ -209,6 +209,12 @@ def __init__(self,
             complex_positions[:protein_n_atoms, :] = protein_positions
             complex_positions[protein_n_atoms:, :] = ligand_positions
 
+            # Convert positions back to openmm vec3 objects
```
mikemhenry, line 212: Have you had a chance to make an issue about this?

mikemhenry (approved):

perses/app/relative_point_mutation_setup.py
zhang-ivy: I made an issue about this in OpenMM: https://github.com/openmm/openmm/issues/3213

PR #855, opened 2021-08-30
Fix geometry engine bug

zhang-ivy:
## Description

I found a bug when playing around with the geometry engine. I think the changes in this PR are necessary because if `topology_index_map` is `None`, then line 2232 will throw an error because you cannot retrieve four indices from an object that is `None`. Therefore, line 2232 and subsequent lines must be indented to be within the `if topology_index_map is not None:` statement.

## Motivation and context

Resolves #???

## Change log

```

```


jchodera:
@zhang-ivy : Did you have a test case that would cause things to fail without this change?

zhang-ivy:
I encountered this error when I was working on this PR: https://github.com/choderalab/perses/pull/834
At one point, there was a bug in the PR which caused `topology_index_map` to be None, leading me to discover this bug. 

If the topology proposal is being generated correctly, `topology_index_map` should never be `None`. But if for some reason it isn't generated properly, the geometry engine will handle it incorrectly, as I described above.

jchodera:
@zhang-ivy : Do you know if this bug is still relevant?

zhang-ivy:
Its not a high priority bug, but its certainly still relevant. Can someone (@mikemhenry or @ijpulidos ) review and merge?

PR #844, opened 2021-08-12
[WIP] Minimal working, automated and tested examples

ijpulidos:
## Description

Removing old and obsolete examples. Leaving current atom-mapping notebook and cleaning up the protein mutation non-equilibrium switching API example.

Still missing protein-ligand and only ligand examples, for both CLI and API.

## Motivation and context

Resolves #813 

## How has this been tested?

Just local tests so far (manually checking it runs and it gives reasonable output), need to write automated tests for this.

## Change log

```
Minimal examples of protein, protein-ligand and ligand mutations, with automated testing.
```

PR #858, opened 2021-09-01
Add GPU testing to our CI

### LINKED ISSUE #853: GPU continuous integration
It would be great if we could get GitHub Actions runnings running on a GPU!

mikemhenry:
## Description

This workflow spins up an ec2 instance with a GPU (`p2.xlarge`), runs our tests, then terminates the EC2 instance.  Right now I have it just use one python version and openmm from conda-forge, but we can expand this if we want either in this PR or at a later date.

## Motivation and context

I just wanted to see how long these tests take and see how it behaves. To save CI time, we might want to set this to only trigger manually so after a PR is ready to go, we run the tests on AWS. We could also only run a subset of the tests as well, such as skipping any tests that are CPU only anyway. 

I used this github action https://github.com/machulav/ec2-github-runner and had to add the following secrets to the repository to get this to work. 

```
AWS_ACCESS_KEY_ID
AWS_REGION 
AWS_SECRET_ACCESS_KEY
GH_PERSONAL_ACCESS_TOKEN  
```

Resolves #853 

## Change log

```
Added GPU testing to our CI
```

### DIFF
--- a/perses/tests/test_CUDA.py
+++ b/perses/tests/test_CUDA.py
@@ -1,12 +1,13 @@
 import pytest
 
+
 @pytest.mark.gpu_needed
 def test_example_barnase_barstar_neq_switching():
     import logging
     import os
     import pathlib
-    import time
     import tempfile
+    import time
 
     import numpy as np
     from openmmtools.integrators import PeriodicNonequilibriumIntegrator
--- a/perses/tests/test_CUDA.py
+++ b/perses/tests/test_CUDA.py
@@ -1,12 +1,12 @@
 import pytest
 
-
-@pytest.mark.gpu_ci
-def test_example_barnase_barstar_neq_switching(tmp_path):
+@pytest.mark.gpu_needed
+def test_example_barnase_barstar_neq_switching():
     import logging
     import os
     import pathlib
     import time
+    import tempfile
 
     import numpy as np
     from openmmtools.integrators import PeriodicNonequilibriumIntegrator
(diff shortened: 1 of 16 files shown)


jchodera:
Awesome!

Does this "fail safe" and time out after a while, spinning the instance down? Otherwise, failures that get "stuck on" can get very pricey!

mikemhenry:
There doesn't seem to be a fail safe if something goes really wrong, but really we want a fail safe that works outside the github action to be really safe, This https://docs.aws.amazon.com/AWSEC2/latest/WindowsGuide/UsingAlarmActions.html#AddingTerminateActions looks promising and I'll see if I can set it up to terminate after an hour.

jchodera:
Adding a separate mechanic for terminating after an hour sounds like a good way to ensure things don't run away!

ijpulidos:
@mikemhenry We should change to be triggered just on 'nightly'.

mikemhenry:
@ijpulidos 
GPU tests on aws are working now :tada: 
Right now I have it just running the tests in `tests/test_resume.py `
Thoughts on other tests that would be good to run on a GPU?

ijpulidos:
@mikemhenry This is great! :tada: :rocket: 
And yes, I can think that at least the examples could be very good candidates for this.  Basically [this test](https://github.com/choderalab/perses/blob/master/perses/tests/test_examples.py). Even though, I may need to change some things in the actual examples for them to use CUDA instead of CPU, not sure if they end up using the "fastest" one, but it's worth giving it a try as it is now and check.

perses/tests/test_CUDA.py
```diff
@@ -0,0 +1,213 @@
...
+        with open(os.path.join(out_path, f"reverse_eq_old.npy"), "wb") as out_file:
+            np.save(out_file, np.array(reverse_eq_old))
+        with open(os.path.join(out_path, f"forward_neq_old.npy"), "wb") as out_file:
+            np.save(out_file, np.array(forward_neq_old))
+        with open(os.path.join(out_path, f"forward_neq_new.npy"), "wb") as out_file:
+            np.save(out_file, np.array(forward_neq_new))
+        with open(os.path.join(out_path, f"reverse_neq_old.npy"), "wb") as out_file:
+            np.save(out_file, np.array(reverse_neq_old))
+        with open(os.path.join(out_path, f"reverse_neq_new.npy"), "wb") as out_file:
+            np.save(out_file, np.array(reverse_neq_new))
```
ijpulidos, line 213: @mikemhenry As far as I see this seems very similar (if not the same) as the example we have already for protein neq switching. Other than having to specify the CUDA platform, are there any reasons why we need this in a separate test/file?

I can think of rewriting the current examples to use the "best" platform available and it should use CUDA when available, I think.

ijpulidos (commented):
Great work, this will help a lot with testing the software with more 'real-life' examples. Other than the doubts I have with the CUDA test this is a LGTM for me. Thanks!

perses/tests/test_CUDA.py
ijpulidos: Tested this with my local setup and it runs in just a few mins (less than 10min for all of the examples). So, is it crazy to make all of these run in the GPU CI workflow? We might just use something like `configure_platform(utils.get_fastest_platform().getName())` for the platform in the examples scripts, and my understanding is that they should use CUDA when available. What do you think?

mikemhenry: I'm afraid that if we don't explicitly specify `CUDA` then we won't know if the GPU stuff stops working and falls back to CPU. So maybe we just have a quick test where we specifically ask for CUDA, and for the rest we just ask for the fastest?

jchodera: Can we also run one test (doesn't have to be all) on `OpenCL` as well?

ijpulidos:
@mikemhenry I have been tracking the progress here and I believe we are hitting some kind of limit with GHA. I still believe that it's better if we use our temp dir context cache [here](https://github.com/choderalab/perses/blob/master/perses/tests/utils.py#L44), just to avoid accidentals `os.chdir` that don't come back to the original working directory.

But that's not the issue, I have the feeling that we hit some limit because this happened before with https://github.com/choderalab/perses/pull/844 and once I stopped running the tests marked as `slow` (that is, just running one of the examples), these issues with files not being found stopped (the issue was another at the time).

ijpulidos:
I don't know if running each test example individually is easy, but if we can do that for each example, then we can be pretty certain that it is some kind of limit that we are hitting. Maybe running the example scripts using subprocesses as we are doing in https://github.com/choderalab/perses/blob/master/perses/tests/test_examples.py#L32 is not good for GHA? Not sure.

mikemhenry:
Yah there is something weird going on with GHA/subprocess/temp files. I'm not sure what the root issue is

mikemhenry:
@ijpulidos what if I undo my hacks (most recent commit) and just not test the examples for now? Then we at least have some GPU testing going

perses/tests/test_examples.py
```diff
@@ -15,14 +15,14 @@
...
-from perses.tests.utils import enter_temp_directory
+from tempfile import TemporaryDirectory
 
 ROOT_DIR_PATH = pathlib.Path(__file__).joinpath("../../../").resolve()
 
 
 def run_script_file(file_path, cmd_args=None):
     """Run through the shell a python script."""
-    with enter_temp_directory():
+    with TemporaryDirectory():
```
ijpulidos, line 25: I believe the 'undo hacks' commit didn't really work? I think we should try using the `enter_temp_directory` from the perses test utils just to avoid accidental dictionary changes that can happen and to keep things a bit cleaner, since that one removes the tree once it's finished.

ijpulidos (approved):
Thos looks very good, great work! Other than what I mentioned about the temp directory, this is good to be merged from my part. We should eventually find out why we cannot run all the [slow] examples using GHA. Thanks!

perses/tests/test_examples.py
mikemhenry: Sorry I must have missed that! Just pushed the fix

mikemhenry:
@ijpulidos I think I've got it all fixed now, the last thing to do will be to change this to only run when we merge to master + run every night (currently set to run on each commit on a PR which we agreed is wasteful)

Any other notes or should I follow this plan (I'll wait to make sure the changes I just made pass)?

ijpulidos:
@mikemhenry seems good, thanks again! 

I agree on running these nightly. I raised an issue on the GHA weird limitation thing that we are facing, because that seems something we want to solve sometime soon-ish.

PR #860, opened 2021-09-01
Simplify `_construct_atom_map` for protein mutations

### LINKED ISSUE #857: AtomMapper returns weird maps for certain protein mutation transformations
I am trying to run a saturated mutagenesis of RBD:ACE2 N501 mutations and most of the atom maps look ok. There are 3 exceptions (here are the atom maps for the apo phase):

**N501Q:**
```
('N', 'N'), ('CA', 'CA'), ('C', 'C'), ('O', 'O'), ('H', 'H'), ('HA', 'HA'), ('CD', 'CG'), ('NE2', 'ND2'), ('OE1', 'OD1'), ('HE22', 'HD21'), ('HE21', 'HD22')
```
Note that the CB and HBs are not mapped here.

**N501G:**
```
('N', 'N'), ('CA', 'CA'), ('C', 'C'), ('O', 'O'), ('H', 'H'), ('HA2', 'HD22')
```
Note that the HA is being incorrectly mapped here.

**N501R:**
```
('N', 'N'), ('CA', 'CA'), ('C', 'C'), ('O', 'O'), ('H', 'H'), ('HA', 'HA'), ('CZ', 'CG'), ('NH1', 'ND2'), ('HH12', 'HD21'), ('HH11', 'HD22')
```
Note that the CB and HBs are not mapped here.

Another note: When I generate atom maps for N->Q, N->G, and N->R for dipeptide systems, the atom maps are similarly problematic.

The problematic code lies in the atom mappings returned by `AtomMapper` [here](https://github.com/choderalab/perses/blob/master/perses/rjmc/topology_proposal.py#L1300-L1301). 
- Note that the atom map indices returned by `AtomMapper` do not use the same numbering scheme as the residue oemols in the `construct_atom_map` function, but you can see the atom map with indices converted to residue oemol indices [here](https://github.com/choderalab/perses/blob/master/perses/rjmc/topology_proposal.py#L1340) 

However, I'm not sure specifically what's going wrong in the `AtomMapper` to cause all these problematic maps. I tried looking through the code to see what might be wrong, but its not obvious to me -- this will require more hands on debugging.

Since this issue is currently blocking my experiments for the N501 saturated mutagenesis, here are the action items I propose:

- Short term: Since the `AtomMapper` code is quite complex and atom mapping for protein mutations is more straightforward than for small molecules, I suggest we replace the `AtomMapper` call with a simpler function that generates atom mappings for standard amino acids by atom name and follows this paradigm: 
    1. always map backbone atoms (exception: GLY, for which we should not map HA2 and HA3),
    2. by default, map CBs only, but allow the user to either demap these or map these + more sidechain atoms.

- Long term: Fix the `AtomMapper` to appropriately handle protein mutations (so that we can easily generate mappings for nonstandard amino acids etc), but this would require a significant time investment.

zhang-ivy:
## Description
I've simplified the code for generating protein mutation atom mappings by generating the mapping directly from the old/new residue topologies. Generating the atom mapping no longer involves using `AtomMapper` or oemols, but oemols are still generated and attached to the output `TopologyProposal` because they are needed as input to the geometry engine.

By default, we map backbone atoms and CB atoms (except for GLY, where we map CA but not HA2 or HA3).

## Motivation and context

Resolves #857 

## How has this been tested?

I added `test_protein_atom_maps()` to `test_topology_proposal.py`, which tests the generation of atom maps for some representative transformations.

## Change log

```

```


perses/rjmc/topology_proposal.py
```diff
@@ -1109,237 +1106,144 @@ def _is_residue_equal(self, residue, other_residue):
...
-        _logger.debug(f"\t\t\told res: {old_res.name}; new res: {new_res.name}")
-
-        new_res_index_to_name = {atom.index: atom.name for atom in new_res.atoms()}
-        old_res_index_to_name = {atom.index: atom.name for atom in old_res.atoms()}
-
-        _logger.debug(f"\t\t\told topology res names: {old_res_index_to_name}")
-        _logger.debug(f"\t\t\tnew topology res names: {new_res_index_to_name}")
 
+        # Retrieve old and new residues
+        assert len(residue_map) == 1, f"residue_map is not of length 1"
```
mikemhenry, line 1159:
```suggestion
        assert len(residue_map) == 1, "residue_map is not of length 1"
```

perses/tests/test_topology_proposal.py
```diff
@@ -336,6 +337,71 @@ def test_mutate_from_alanine():
         else:
             _ = generate_dipeptide_top_pos_sys(ala.topology, amino, ala.system, ala.positions, system_generator, conduct_geometry_prop=False)
 
+def test_protein_atom_maps():
+
```
dominicrufa, line 341: it looks like you hardcoded some specific mutations. is this robust enough? i guess you can augment this test as/if you encounter problems in future?

perses/tests/test_topology_proposal.py
```diff
@@ -336,6 +337,71 @@ def test_mutate_from_alanine():
...
             _ = generate_dipeptide_top_pos_sys(ala.topology, amino, ala.system, ala.positions, system_generator, conduct_geometry_prop=False)
 
+def test_protein_atom_maps():
+    
+    # Get alanine dipeptide in vacuum test system
+    ala, system_generator = generate_atp()
+    
+    
+    # Define function for checking that the atom map is correct
+    def check_atom_map(topology_proposal, reference_map):
```
dominicrufa, line 347: can this be a utility outside the `test` fn?

perses/rjmc/topology_proposal.py
```diff
@@ -1109,237 +1106,144 @@ def _is_residue_equal(self, residue, other_residue):
...
-        current_oemol : openeye.oechem.oemol object
-            copy of modified old oemol sidechain
-        proposed_oemol : openeye.oechem.oemol object
-            copy of modified new oemol sidechain
-        old_oemol_res_copy : openeye.oechem.oemol object
-            copy of modified old oemol
-        new_oemol_res_copy : openeye.oechem.oemol object
-            copy of modified new oemol
-
+        local_atom_map : dict, key: int, value int
```
dominicrufa, line 1133: i vaguely remember there being a reason we make a deepcopy of the `oemol` to append to the `Topology` object. can we keep that?

perses/rjmc/topology_proposal.py
```diff
@@ -1109,237 +1106,144 @@ def _is_residue_equal(self, residue, other_residue):
...
         old_res_name = old_res.name
         new_res_name = new_res.name
 
-        current_residue_pdb_filename = resource_filename('perses', os.path.join('data', 'amino_acid_templates', f"{old_res_name}.pdb"))
-        proposed_residue_pdb_filename = resource_filename('perses', os.path.join('data', 'amino_acid_templates', f"{new_res_name}.pdb"))
-
-        current_oemol = PolymerProposalEngine.generate_oemol_from_pdb_template(current_residue_pdb_filename)
-        proposed_oemol = PolymerProposalEngine.generate_oemol_from_pdb_template(proposed_residue_pdb_filename)
-
-        old_oemol_res_copy = copy.deepcopy(current_oemol)
```
dominicrufa, line 1212: reference to the oemol deepcopy is here.

perses/rjmc/topology_proposal.py
dominicrufa: scratch this. i dont think we need to keep deepcopy anymore.

dominicrufa (commented):
looks good. the test doesnt provide full coverage of protein mutations, but that might be unnecessary?

perses/tests/test_topology_proposal.py
zhang-ivy: It will never be used outside of this function, so there's no need to move it outside `test_protein_atom_maps`

perses/rjmc/topology_proposal.py
zhang-ivy: agreed. the deepcopy was necessary before because we were deleting backbone atoms from the original oemol (which would then be inputted to `AtomMapper`). However, we are no longer using the oemol for generating the map.

perses/tests/test_topology_proposal.py
zhang-ivy: I think this is robust enough, as the test cases are representative of the different cases handled by the code: 
1) mutations where the old or new residue is GLY
2) mutations where the old or new residue is in ['THR', 'ILE', 'VAL']
3) all other mutations

perses/tests/test_topology_proposal.py
```diff
@@ -336,6 +337,71 @@ def test_mutate_from_alanine():
...
+        # Retrieve atom index to name mapping for old and new residues
+        old_res = [res for res in topology_proposal.old_topology.residues() if res.name == topology_proposal.old_residue_name][0]
+        new_res = [res for res in topology_proposal.new_topology.residues() if res.name == topology_proposal.new_residue_name][0]
+        old_res_index_to_name = {atom.index: atom.name for atom in old_res.atoms()}
+        new_res_index_to_name = {atom.index: atom.name for atom in new_res.atoms()}
+        print("old map: ", old_res_index_to_name)
```
zhang-ivy, line 353:
```suggestion
       
```

perses/tests/test_topology_proposal.py
```diff
@@ -336,6 +337,71 @@ def test_mutate_from_alanine():
...
+    
+    # Define function for checking that the atom map is correct
+    def check_atom_map(topology_proposal, reference_map):
+        # Retrieve atom index to name mapping for old and new residues
+        old_res = [res for res in topology_proposal.old_topology.residues() if res.name == topology_proposal.old_residue_name][0]
+        new_res = [res for res in topology_proposal.new_topology.residues() if res.name == topology_proposal.new_residue_name][0]
+        old_res_index_to_name = {atom.index: atom.name for atom in old_res.atoms()}
+        new_res_index_to_name = {atom.index: atom.name for atom in new_res.atoms()}
+       
+        print("new map: ", new_res_index_to_name)
```
zhang-ivy, line 354:
```suggestion
           
```

zhang-ivy:
@jchodera Could you review this and let us know if its ok to merge?

PR #861, opened 2021-09-03
Add scaled water implementation

zhang-ivy:
## Description
This PR augments the existing RESTTopologyFactory to scale rest-water interactions by beta/beta_0 (previously was being scaled by sqrt(beta/beta_0).

To do:
- [ ] Add tests (I have these in a notebook, but need to revise them based on JDC's recommendations)
- [ ] Investigate/solve weird solvent artifact issue seen in barnase:barstar scaled water rest caches (but not in RBD:ACE2 caches)

## Motivation and context

Resolves #???

## Change log

```

```


jchodera:
Oh! I thought we wanted to keep intra-REST regions scaled by `beta/beta0`, but REST-environment `sqrt(beta/beta0)`. Perhaps we want to add an optional argument about whether nearby water should be included in the REST region, in which case the scaling would be distance-dependent (go smoothly from `beta/beta0` at `r < r0` to `sqrt(beta/beta0)` at `r > r0`)?

zhang-ivy:
>I thought we wanted to keep intra-REST regions scaled by beta/beta0, but REST-environment sqrt(beta/beta0)

This is what the original REST2 paper describes, so there wouldn't be any need for the scaled water rest implementation if we were following these scaling rules, we could just use the [`RESTToplogyFactory`](https://github.com/choderalab/perses/blob/master/perses/annihilation/rest.py).

To enhance water sampling, we decided to further scale the REST-water interactions (by beta/beta0), but not the water-water interactions (these are left unscaled). We don't actually want to include waters in the REST region, because if we didm, water-water interactions would be scaled by beta/beta0, which we worried may cause weird artifacts like waters flying away.

mikemhenry:
@zhang-ivy is this PR still something we want/need?

zhang-ivy:
Closing this, since we didn't observe that this addressed our sampling problems

Issue #864, opened 2021-09-07
docker container working with GPU


ijpulidos:
It has been reported that the docker container is failing with running actual examples of perses using the GPU. 

This issue is raised to test this is working correctly. Probably by running it in some AWS/GCP instances and checking.

mikemhenry:
Report from a user:

```
 Existing docker image (Perses 0.8.1) on the new A100 GPUs on GCP:

Desired platform not supported. exception raised: Error compiling program: nvrtc: error: invalid value for --gpu-architecture (-arch)

Warning: Returning Reference platform instead of requested platform CUDA

conducting subsequent work with the following platform: Reference

conducting subsequent work with the following platform: CPU



New docker image (Perses 0.9.1) on the V100s on GCP:


Desired platform not supported. exception raised: Error loading CUDA module: CUDA_ERROR_INVALID_PTX (218)

Warning: Returning Reference platform instead of requested platform CUDA

conducting subsequent work with the following platform: Reference

conducting subsequent work with the following platform: CPU



New docker image (Perses 0.9.1) on the A100s on GCP:


Desired platform not supported. exception raised: Error loading CUDA module: CUDA_ERROR_INVALID_PTX (218)

Warning: Returning Reference platform instead of requested platform CUDA

conducting subsequent work with the following platform: Reference

conducting subsequent work with the following platform: CPU


New docker image (Perses 0.9.1) locally (GeForce GTX 1080, Ubuntu 20.04):

Desired platform not supported. exception raised: Error initializing CUDA: CUDA_ERROR_COMPAT_NOT_SUPPORTED_ON_DEVICE (804) at /home/conda/feedstock_root/build_artifacts/openmm_1622798693339/work/platforms/cuda/src/CudaContext.cpp:138

Warning: Returning Reference platform instead of requested platform CUDA

conducting subsequent work with the following platform: Reference

conducting subsequent work with the following platform: CPU


nvidia-smi works in the local container.
```
On these nodes
![image](https://user-images.githubusercontent.com/11765982/132422410-99dbeefc-a1ea-4b9f-96d7-d116de5be778.png)
![image](https://user-images.githubusercontent.com/11765982/132422441-a7dd25d8-6b57-49fc-af1f-b1c02ac10860.png)

jchodera:
> Desired platform not supported. exception raised: Error compiling program: nvrtc: error: invalid value for --gpu-architecture (-arch)

[...]

PR #869, opened 2021-09-17
Version CUDA Support Better With Dockerfiles

mikemhenry:
## Description

Using CUDA 11 as a descriptor isn't enough as cuda 11.0.3 on the host didn't work with a cuda image built with 11.2.2. 

## Motivation and context

This PR adds support for a cuda image that will work on GCP's container optimized os (cos) by providing a docker image that is built with cuda 11.0.3.

## How has this been tested?

Tested manually on GCP on a V100 GPU.

## Change log

```
Added a docker image for CUDA 11.0.3
```


mikemhenry:
Still need to add a note in the readme

mikemhenry:
Ready for review! Also see this comment, https://github.com/choderalab/perses/issues/864#issuecomment-921847591 looks like the changes worked!

ijpulidos (approved):
Great work, any ideas why is it failing with CUDA 11.2? Is that only on GCP?

mikemhenry:
It works fine with CUDA 11.2, but just not on systems that only support CUDA 11.0

mikemhenry:
Superseded by PR #1087

PR #873, opened 2021-10-05
Add scaled water rest implementation (second attempt)

zhang-ivy:
## Description
This PR creates `RESTTopologyFactoryV3`, a class that generates a rest-like system where the energies are scaled according to rest with the following exception: scale rest-water interactions by beta/beta_0 (previously was being scaled by sqrt(beta/beta_0).

## Change log

```

```


mikemhenry:
@zhang-ivy is this PR still something we want/need?

PR #878, opened 2021-10-12
Fix stochastic failures in RepartitionedHybridTopologyFactory test

### LINKED ISSUE #815: Stochastic test failures in RepartitionedHybridTopologyFactory_energies

zhang-ivy:
## Description

Changes  to `test_relative.py::test_RepartitionedHybridTopologyFactory_energies`:
- Only test ALA -> SER (not ALA -> all other amino acids)
- Only propose geometry once (not 5x)

Changes to `RepartitionedHybridTopologyFactory`:
- Fix bug in instantiation of `NonbondedForce`

## Motivation and context

We were previously seeing occasional failures with `test_relative.py::test_RepartitionedHybridTopologyFactory_energies`. This was due to the geometry engine proposing clashy positions, which have very, very large energy, causing the `validate_endstate_energies()` check to fail because the energy threshold of 1e-4 becomes difficult to pass with very large energies.

The changes introduced here stop testing ALA-> all other amino acids and instead restricts testing to ALA -> SER mutations, because its unlikely that the geometry engine will propose clashy positions for this conservative mutation.

I also caught a bug in the `RepartitionedHybridTopologyFactory` class implementation where the a `NonbondedForce` was not instantiated properly.

Resolves #815 

## How has this been tested?

See `test_relative.py::test_RepartitionedHybridTopologyFactory_energies`

## Change log

```

```

### DIFF
--- a/perses/app/relative_point_mutation_setup.py
+++ b/perses/app/relative_point_mutation_setup.py
@@ -285,11 +285,17 @@ class PointMutationExecutor(object):
                             continue
                         charge, sigma, epsilon = nb_force.getParticleParameters(idx)
                         if sigma == 0*unit.nanometer:
-                            sigma = 0.06*unit.nanometer
-                            nb_force.setParticleParameters(idx, charge, sigma, epsilon)
+                            new_sigma = 0.06*unit.nanometer
+                            nb_force.setParticleParameters(idx, charge, new_sigma, epsilon)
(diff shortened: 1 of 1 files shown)


zhang-ivy:
@mikemhenry @ijpulidos : Please review and merge!

PR #884, opened 2021-10-27
WIP: Optional oechem for AtomMapper

### LINKED ISSUE #820: Use open source MCSS implementation
Eventually, we might consider incorporating open MCSS implementations from [LoMap](https://github.com/MobleyLab/Lomap/blob/55dd9818f43e4cb238b599d45ac777d6583d713b/lomap/mcs.py).

richardjgowers:
## Description

An attempt to make the `AtomMapper` optionally require OEChem, by either taking an OEMol or openff.Molecule object.  N.B. I'm not sure if this is even the right approach, as an openff Molecule can also be either an rdkit or oechem currently, so maybe a better approach is to completely commit to OpenFF objects (as it is already also OEChem) and instead "undress" the openff molecule invisibly to the user.

## Motivation and context

Resolves #820 ?

This is mostly just an afternoon of seeing how baked in OEChem is, and if removal is possible, it seems reasonable.

I was also reading up on the state of LOMAP ([here](https://github.com/MobleyLab/Lomap/pull/54) in particular) and it seems like ultimately many of the implementations I'll need to do this already will exist there (or directly in rdkit, see above), but I agree with the discussion there that this API is more manageable.

I'm happy developing on this branch (and into perses) for now, but ultimately a way forward might be to lift AtomMapper (et al) out into a "Lomap3" package which is maintained by OpenFE.  Still mulling that idea over, but what are the thoughts of removing this class from Perses and having OpenFE supply a dependency?

### DIFF
--- a/perses/rjmc/atom_mapping.py
+++ b/perses/rjmc/atom_mapping.py
@@ -19,10 +19,16 @@ try:
     from openeye import oechem
 except ImportError:
     HAS_OECHEM = False
-    from perses.utils.off import get_scaffold
+    from perses.utils.off import get_scaffold as off_get_scaffold
+
+    def is_oechem(mol):
+        return False
 else:
     HAS_OECHEM = True
-    from perses.utils.openeye import get_scaffold
+    from perses.utils.openeye import get_scaffold as oe_get_scaffold
+
+    def is_oechem(mol):
+        return hasattr(mol, 'GetAtoms')
 
 ################################################################################
 # LOGGER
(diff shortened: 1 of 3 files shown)


jchodera:
@richardjgowers We should chat about this soon since there is a straightforward way to address this in stages.

richardjgowers:
@jchodera I think I've come to think that a "pure" OFF toolkit implementation either won't work (as OFF lacks the cheminformatics bells and whistles) or would be too slow, so they'll always have to be a "real" cheminformatics toolkit under the hood (also tangentially I'm not sure networkx can replace these either).  So then this issue turned into rewriting the AtomMapper to use rdkit in place of oechem, which I'm not sure is productive.  I'm also wary about trying to make an AtomMapper give identical results for more than one toolkit backend, as there's lots of weird corner cases you can soon hit.

PR #887, opened 2021-11-01
use beta of openeye toolkit

### LINKED ISSUE #883: Add OpenEye beta tests to CI
We should add OpenEye toolkits beta candidates to our CI.

mikemhenry:
## Description

This PR installs openeye toolkit with the `beta` label https://anaconda.org/OpenEye/OpenEye-toolkits/files
Instead of expanding the matrix, I opted to do this as part of the openmm nightly check, since it isn't required to pass + didn't want to eat up too much CI time. 

## Motivation and context

We want to be able to check if any changes to the toolkit cause our tests to fail. We also want to be able to report bugs back to openeye.

Resolves #883 

## Change log

```

```

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -55,7 +55,7 @@ jobs:
           case ${{ matrix.openmm }} in
             nightly)
               echo "Using nightly dev build of OpenMM."
-              conda install -y -c omnia-dev openmm --override-channels --force-reinstall;;
+              conda install -y -c omnia-dev openmm --override-channels --force-reinstall
               echo "Using beta build of openeye toolkit"
               conda install -y -c openeye/label/beta openeye-toolkits --override-channels --force-reinstall;;
             *)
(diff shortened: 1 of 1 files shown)


mikemhenry:
Two things to conisder:

- We will probably want to change this to the release candidate when it comes out
- Once the new release is cut, we may need to remove this in case it pulls in an older version

ijpulidos:
@mikemhenry Thanks for making the PR and changes. Do we know what beta we are getting? We know the OpenEye toolkits releases a lot of betas before the actual rc, so I don't really know if we can be sure we are testing the "right" one or whether we should be just testing the rc that will eventually come from the beta. Thoughts?

mikemhenry:
Looks like it pulled in 
`openeye-toolkits          2021.2.0b7               py37_0    openeye/label/beta`

mikemhenry:
@ijpulidos It sounds like the Beta and RC will be close enough for our purposes, so I think this is ready to merge.

ijpulidos (approved):
LGTM :tada:

Issue #845, opened 2021-08-13
GeometryEngine proposes wrong stereochemistry for CA in RBD:ACE2 mutations


zhang-ivy:
@jchodera and I found in issue with the geometry proposals in WT RBD:ACE2 N501Y and beta variant RBD:ACE2 Y501N -- the CA chirality is R, not S. This happens in both apo and complex.
Note that for these mutations, I do not map CB atoms and beyond. 

<img width="507" alt="Screen Shot 2021-08-13 at 12 04 44 PM" src="https://user-images.githubusercontent.com/35546250/129387447-49a9c7fc-ef2d-4dc9-8949-bf4efc03205c.png">

This does not seem to be a problem in barnase:barstar (I do map CBs here) or capped ASN->TYR (I do not map CBs here). 

@dominicrufa mentioned that the stereochemistry code (which starts [here](https://github.com/choderalab/perses/blob/master/perses/rjmc/geometry.py#L2253)) maintains stereochemistry for atoms in unique old and new, but not for core atoms. This is a problem for amino acid mutations when we don't map CBs, because then CA chirality can be proposed with wrong stereochemistry. 

@jchodera : If you see a quick fix for this, please let me know!
I've attached scripts you can use to generate geometry proposals for capped ASN->TYR as well as RBD:ACE2 N501Y. 
- For the former, I've included a PDB of ASN that you'll need to load into the script. 
- For the latter, I've also included an apo htf you can load in to the script so that you can access the topology proposal and old positions.

[debug_geometry_engine_stereo.zip](https://github.com/choderalab/perses/files/6983911/debug_geometry_engine_stereo.zip)

jchodera:
@zhang-ivy : I can't seem to run the RBD:ACE2 example:
```
(perses-dev) lski2838:debug_geometry_engine_stereo choderaj$ python generate_geometry_proposals_rbd_ace2.py 
INFO:numexpr.utils:NumExpr defaulting to 8 threads.
Traceback (most recent call last):
  File "generate_geometry_proposals_rbd_ace2.py", line 14, in <module>
    htf = pickle.load(f)
ModuleNotFoundError: No module named 'openmm'
```
I'm guessing you might have a more recent development snapshot of OpenMM installed? Can you dump your conda env?

jchodera:
[...]

PR #892, opened 2021-11-12
Allow `generate_dipeptide_top_pos_sys` to accept `demap_CBs`

zhang-ivy:
## Description

This PR piggy backs off of https://github.com/choderalab/perses/pull/860, where I forgot to allow `demap_CBs` to be passed to `generate_dipeptide_top_pos_sys`.

## Motivation and context

This is necessary because to enable generating protein mutation test systems with demapped CBs.

## Change log

```

```

### DIFF
--- a/perses/tests/test_topology_proposal.py
+++ b/perses/tests/test_topology_proposal.py
@@ -202,6 +202,7 @@ def generate_dipeptide_top_pos_sys(topology,
                                    positions,
                                    system_generator,
                                    extra_sidechain_map=None,
+                                   demap_CBs=False,
                                    conduct_geometry_prop=True,
                                    conduct_htf_prop = False,
                                    validate_energy_bookkeeping=True,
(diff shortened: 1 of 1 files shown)


zhang-ivy:
@mikemhenry @ijpulidos  : could you review and merge?

ijpulidos (approved):
Ah! good catch. This looks just fine. Thanks!

ijpulidos:
@zhang-ivy This might be for another issue/PR, but just wondering if we have tests for this case. Such as a test that requires `demap_CBs=True` for it to pass. Thoughts? (I guess it is related to the #845 , maybe we can engineer some test around that case).

zhang-ivy:
>@zhang-ivy This might be for another issue/PR, but just wondering if we have tests for this case. Such as a test that requires demap_CBs=True for it to pass. Thoughts? (I guess it is related to the #845 , maybe we can engineer some test around that case).

Yes, it couldn't hurt to have another test that checks whether `demap_CBs` actually demaps the CBs. Would you want to take a stab at this? It should be something like this: [choderalab](https://github.com/choderalab/perses/blob/master/perses/tests/test_topology_proposal.py#L340) -- but would just have to change the `reference map` to not include CBs.

PR #896, opened 2021-12-01
Only support openmm 7.6

mikemhenry:
This adjusts our CI to only test 7.6 + nightly, sets the min version of openmm to 7.6 and removes and only env file.


mikemhenry:
@ijpulidos I got some nan's in the python 3.9 tests so I will re-run, but assuming things pass this should be good for review and merge!

devtools/conda-envs/test_env.yaml
```diff
@@ -17,7 +17,7 @@ dependencies:
   - numexpr
   - autograd
   - pymbar
-  - openmm=7.5.1
+  - openmm=7.6
```
ijpulidos, line 20: I think we don't want to pin any specific `openmm` version. It should use the latest available and it would also not require manually changing it for future openmm releases. Isn't that the case?

ijpulidos (changes requested):
This looks good! I just think not pinning the openmm version should work better.

ijpulidos:
Why is it trying to test OpenMM 7.5.1 in the checks?

devtools/conda-envs/test_env.yaml
mikemhenry: I don't think there is an issue with having to manually change as openmm releases come out since if a new version comes out that breaks things, then we would want to keep testing against the older version while we update.

mikemhenry:
> Why is it trying to test OpenMM 7.5.1 in the checks?

In the settings for the repo, those checks are set to "required" so it expects them in the PR. Until we merge this one in, I didn't want to change the repo setting since that would affect other PRs.

devtools/conda-envs/test_env.yaml
mikemhenry: I'm open to other ideas! Also pinning the "oldest" version we support makes the logic of refining the env a bit easier in the CI, but we could probably refactor it so we don't need to do that.

jchodera: We should pin `openmm >=7.6`, not `=`!

.github/workflows/CI.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [3.7, 3.8, 3.9]
-        openmm: ["nightly", "7.5.1.", "7.6.0"]
+        openmm: ["nightly", "7.6.0"]
```
ijpulidos, line 23: I know this might not be an issue for the current OpenMM release (7.6), but I wonder if we want to use only `major.minor` and not `patch/build` versions, for future openmm releases. That is, only doing "7.6" instead of "7.6.0". 

Would that work as expected with mamba when these options are used? Such that doing something like `mamba install openmm=7.6` hopefully gets the latest 7.6.x for openmm.

ijpulidos (approved):
Minor comment, for future releases of openmm. But a non-blocker.

.github/workflows/CI.yaml
mikemhenry: That is a good idea, this is how we install it:
`mamba install -y -c conda-forge openmm=${{ matrix.openmm }}`

So doing `7.6` should pull in a newer minor version, but do want to explicitly test minor versions? OpenMM doesn't follow symver so we might want to be explicit.

mikemhenry: Going to do 7.6 :)

PR #905, opened 2021-12-23
Fix water selection for counterion

zhang-ivy:
## Description

Currently, we choose a water at least X nm from protein to change into an ion. If there are glycan residues, the current implementation is incorrect, as a water may be chosen near a glycan, which we don't want. 

This PR changes the selection string from `protein` to `not water and not resn NA and not resn CL and not resn ZN`. The ions in the selection string may be supplied to the method as a list of strings.

## Motivation and context

Resolves #???

## Change log

```

```

### DIFF
--- a/perses/rjmc/topology_proposal.py
+++ b/perses/rjmc/topology_proposal.py
@@ -531,8 +531,7 @@ class PolymerProposalEngine(ProposalEngine):
     def get_water_indices(charge_diff,
                                new_positions,
                                new_topology,
-                               radius=0.8,
-                               ion_names=['NA', 'CL', 'ZN']):
+                               radius=0.8):
         """
         Choose random water(s) (at least `radius` nm away from the protein) to turn into ion(s). Returns the atom indices of the water(s) (index w.r.t. new_topology)
 
@@ -546,8 +545,6 @@ class PolymerProposalEngine(ProposalEngine):
             topology of new system
         radius : float, default 0.8
             minimum distance (in nm) that all candidate waters must be from 'protein atoms'
-        ion_names : list of str, default ['NA', 'CL', 'ZN']
-            ion residue names to exclude when choosing waters far away from nonsolvent atoms
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


mikemhenry (approved):
LGTM! Is there a test we could add that confirms the selection is what we want? Or maybe is there a test that if we had it, we would have caught this bug?

zhang-ivy:
>Is there a test we could add that confirms the selection is what we want?

I don't think so. Any test I can think of would just involve running mdtraj again (but I think we should assume mdtraj is returning the correct indices).

> Or maybe is there a test that if we had it, we would have caught this bug?

I can't think of a test that would have caught this bug. This fix is necessary for systems with glycans, but most typical computational chemistry systems don't have glycans.

If anyone has any ideas for a test, feel free to push to this PR and I can review. Otherwise, this should be ready to merge.

jchodera:
> This PR changes the selection string from protein to not water and not resn NA and not resn CL and not resn ZN. The ions in the selection string may be supplied to the method as a list of strings.

What if there are other ions in the system that are not NA, CL, or ZN?

MDTraj has a more robust way to exclude solvent in [`remove_solvent`](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/core/trajectory.py#L1826-L1861), but this is unfortunately not something that can be used as part of the DSL. Instead, this would require using part of the private API:
```python
from mdtraj.core.residue_names import _SOLVENT_TYPES
solvent_types = list(_SOLVENT_TYPES)
solute_indices = [atom.index for atom in mdtraj_topology.atoms if atom.residue.name not in solvent_types]
```
I suggest we use this scheme and note a `TODO` in the code to replace it once we're able to either (1) get `solvent` as a keyword into the [MDTraj DSL](https://mdtraj.org/1.6.2/atom_selection.html), or (2) transition to MDAnalysis (which is probably a good idea given that the OpenFE devs are also MDAnalysis devs).

zhang-ivy:
>What if there are other ions in the system that are not NA, CL, or ZN?

I've made the `ion_names` an optional argument, so alternative ion names can be specified, but your suggestion is better, so I'll fix that now.

mikemhenry:
@zhang-ivy is this ready to merge?

zhang-ivy:
Yes, as long as others have reviewed it.

ijpulidos (approved):
LGTM! Agree with the comments/TODO.

PR #922, opened 2022-01-20
Fix box vector handling in topology proposal

zhang-ivy:
## Description

This PR makes sure that when a source topology is being copied to a destination topology, the source topology's box vectors are copied to the destination topology.

## Motivation and context

OpenMM only writes out box vectors to DCD files if the `topology.getPeriodicBoxVectors()` is not None: https://github.com/openmm/openmm/issues/3418

I ran a simulation and the box vectors were not saved because `htf._topology_proposal.old_topology` does not have box vectors. This PR addresses the problem by adding a line that copies the box vectors from the source to destination topology in `append_topology()`. It removes the line in `_propose()` that does this for the new topology only.

## Change log

```

```

### DIFF
--- a/perses/rjmc/topology_proposal.py
+++ b/perses/rjmc/topology_proposal.py
@@ -170,6 +170,7 @@ def append_topology(destination_topology, source_topology, exclude_residue_name=
         If specified, any residues matching this name are excluded.
 
     """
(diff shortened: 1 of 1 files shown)


jchodera (approved):
Thanks for adding this!

Issue #903, opened 2021-12-17
Skip codecov uploads on scheduled CI runs


mikemhenry:
RE:
https://community.codecov.com/t/too-many-uploads-to-this-commit-code-invalid/3384/2

We should skip uploading coverage reports to codecov if the CI was triggered by a schedule, since we don't expect coverage to change if code doesn't change.

PR #932, opened 2022-02-02
Stop uploading coverage reports on nightly CI runs

mikemhenry:
We had an issue where codecov would fail to upload reports when we uploaded too many with the same commit hash, this will prevent reports being uploaded when we don't expect anything to change.


mikemhenry:
Fixes issue #903

ijpulidos (approved):
Looks good!

PR #934, opened 2022-02-04
Add CLI that creates dummy output files and validates input

### LINKED ISSUE #933: Create Docker with new CLI
@jchodera I just wanted to make sure that the dummy/testing CLI has the functionality we want:

- [ ] Reads in a yaml file and checks the input
- [ ] Checks to make sure we can use a CUDA context
- [ ] Creates some dummy/mock output files 

I'm working on this right now and should have it done by tomorrow, but I wanted to make sure I wasn't missing anything.

mikemhenry:
## Description

Creates a new CLI tool `perses-cli` that takes in a yaml file and creates dummy output.
`perses-cli --yaml-path protein-ligand.yaml`

## Motivation and context

Resolves #933 

## How has this been tested?

Tested locally 

## Change log

```
Added a dumy cli tool for testing workflows.
```


mikemhenry:
Needs some tests

- [ ] add tests

mikemhenry:
@ijpulidos if you get a chance can you take a look at why the resume from checkpoint test is taking such a long time? I'm not sure why the test is suddenly taking a long time now. Looking at the CI logs, it didn't take this long of a time during the scheduled run. 

I'm also not sure why the test I added is failing on CI but passes locally (haha what else is new)

ijpulidos:
@mikemhenry Works for me locally, that's strange indeed. I wonder if there's something going on with the `test_env.yaml` + `conda` combination. I see this differences from the environment in the CI runs compared to the environment I use locally from the same `test_env.yaml` file. https://gist.github.com/ijpulidos/2674755d0b24bd7b648bca875e94b846

mikemhenry:
Thanks! I'll keep playing around with this to get the CI working, but do you mind doing a review?

perses/app/cli.py
```diff
@@ -0,0 +1,150 @@
...
+    click.echo("🕵️\t Checking OpenEye license")
+    _check_openeye_license()
+    click.echo("✅\t OpenEye license good")
+    click.echo("🖥️⚡\t Making a test system to check if we can get a GPU")
+    _test_gpu()
+    click.echo("🎉\t GPU test sucsessful!")
+    click.echo("🖨️\t Writing out files")
+    trajectory_directory = options["trajectory_directory"]
+    _write_out_files(trajectory_directory, options)
+    click.echo("🧪\t Simulation over")
```
ijpulidos, line 150: I wonder if we are missing an entry point here or doing something like
```python
if __name__ == '__main__':
    cli()
```
Might be just me ignoring how to actually use the cli, though. But even so, I guess we do want some kind of help or instructions on how to use the CLI, right?

perses/tests/test_cli.py
```diff
@@ -0,0 +1,47 @@
...
+trajectory_directory: temp/offlig10to24
+trajectory_prefix: out
+atom_selection: not water
+phases:
+  - complex
+  - solvent
+  - vacuum
+timestep: 4
+h_constraints: true
+"""
```
ijpulidos, line 37: From what I can see, this is the same yaml file in the `examples/protein-ligand/cli/protein-ligand.yaml` path. I wonder if we want to use that file directly to avoid syncronization and redundency issues once things change. So, something like `python cli.py --yaml-path=/path/to/perses/examples/protein-ligand/cli/protein-ligand.yaml`

perses/app/cli.py
ijpulidos: Ah, I see there's an entry-point in the `setup.py`, I think that works, but will leave the comment just in case.

perses/app/cli.py
```diff
@@ -0,0 +1,150 @@
...
+    click.echo("📖\t Fetching simulation options ")
+    options = getSetupOptions(yaml_path)
+    click.echo("🖨️\t Printing options")
+    click.echo(options)
+    click.echo("🕵️\t Checking OpenEye license")
+    _check_openeye_license()
+    click.echo("✅\t OpenEye license good")
+    click.echo("🖥️⚡\t Making a test system to check if we can get a GPU")
+    _test_gpu()
+    click.echo("🎉\t GPU test sucsessful!")
```
ijpulidos, line 146: Just a typo fix :).
```suggestion
    click.echo("🎉\t GPU test successful!")
```

ijpulidos (commented):
I'm really vibing with click, looks neat!! Just a few comments. Thanks!

perses/tests/test_cli.py
ijpulidos: Or whatever fix the test in this case, but I guess the point still stands on using the file we already have instead of copying it.

perses/app/cli.py
mikemhenry: That is best-practice, so I will add it.

perses/tests/test_cli.py
mikemhenry: @ijpulidos since our examples are outside our package source, I'm not sure the most reliable way to get the location of that yaml, do you have any ideas? Normally I would use something like this https://docs.python.org/3.8/library/importlib.html#importlib.resources.path but I would like to avoid doing something like `pathlib.Path(__file__).joinpath("../../../").resolve()` if I can help it.

ijpulidos: Or https://setuptools.pypa.io/en/latest/userguide/datafiles.html

mikemhenry: Actually never mind to trying to get this into the 0.9.4 release, if it ends up in that release that is fine, but once I get this PR merged I can just make a docker image with the `dev` label and share that

mikemhenry: But I still like the idea of making an issue to fix all the tests, and just keep the yaml hard coded for now

ijpulidos: Yes, I agree! I created the issue regarding this. CC #942

ijpulidos (approved):
Looks good to me!

jchodera (approved):
This is good for a first pass!
We'll want to let folks know that the YAML format may change in future releases.

jchodera:
@mikemhenry : Whoops! I had meant to select "request changes"---see comments above!

perses/app/cli.py
```diff
@@ -0,0 +1,156 @@
...
+    # add the dir prefix
+    files_in_xml_dir = [dirs_to_make.joinpath(_) for _ in files_in_xml_dir]
+
+    for _file in files_in_lower_dir:
+        with open(_file, "w") as fp:
+            pass
+
+
+@click.command()
+@click.option("--yaml-path", type=click.Path(exists=True, dir_okay=False), required=True)
```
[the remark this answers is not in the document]
mikemhenry: Good idea!

perses/app/cli.py
```diff
@@ -0,0 +1,156 @@
...
+@click.option("-rpe", "--raise-platform-error", is_flag=True, type=str)
+@click.option("--platform", type=str, default="CUDA")
```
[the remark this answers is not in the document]
mikemhenry: Done! Now we only throw an error if the platform a user asked for was unavailable

mikemhenry:
Okay @jchodera ready for another pass!

perses/app/cli.py
```diff
@@ -0,0 +1,159 @@
...
+from pathlib import Path
+
+import click
+import openmm
+from openmm import OpenMMException, unit
+from openmmtools import integrators
+from openmmtools.testsystems import HarmonicOscillator
+from perses.app.setup_relative_calculation import getSetupOptions
+
+percy = """
```
jchodera, line 12: An alternative, based on @dominicrufa's unofficial "Percy Pig" mascot:
```
MMMMMMMMMMMMXo:ccldOKNNWMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMMM
MMMMMMMMMMMWxcxOkxdodddxxxk0XWMMMMMMMMMMMMMMMMMMMMMMMMMMMMMWWWWMMWNX0Okk0WMMMMMM
MMMMMMMMMMMWxlk0OkdddddxkxdoodONWNXK0OOOOOOOOO0KXNWMMMMN0kxxxxxxxxddoddockWMMMMM
MMMMMMMMMMMMKooOOxooooc;:ldOOxooddddodddddddddodddxxkOkoodkOOkxdodxO000kcxWMMMMM
MMMMMMMMMMMMMKolk0000OdoddllxOOkkO00000O000000000OkxdookOOdl:cloooxO00kloXMMMMMM
MMMMMMMMMMMMMMNklokO0OO000OkkO000000000000O000000000O00OdllollxO0000OdldXMMMMMMM
MMMMMMMMMMMMMMMWXkdodkO00O00O0000OdlxO000000000OxxkO000OxkO00OOO00Odlo0WMMMMMMMM
MMMMMMMMMMMMMMMMMMWKkdldOOOO00000kdxOO000000000OxldO0000OO000OOxdoox0NMMMMMMMMMM
MMMMMMMMMMMMMMMMMMMMXdldOOO0000O000kkO0000000O0OOOOO0O00000Odloxk0NMMMMMMMMMMMMM
MMMMMMMMMMMMMMMMMMMKook00000000000Odlk00000000kooO000O00000OxcxNMMMMMWX0KNMMMMMM
MMMMMMMMMMMMMMMMMMKloO00000000000O0OkxdooooodxkxkO000000000O0xlkWMMMNkloook0OkxO
MMMMMMMMMMMMMMMMMNolO000000000000OxlllodddddolloxO00000000O00Odl0MMMOcx0Okooddo:
MMMMMMMMMMMMMMMMMOcx0OO00000O000kocoxxkO000OOkkocoO0000O0O0000kldWMMxck0000000xc
MMMMMMMMMMMMMMMMWdlO00O000000OOOo:dd:,:x000xc;cxd:d0000OO000O0OolXMWxck00OO0Oxlx
MMMMMMMMMMMMMMMMXooOO0000OOOOxkkclkc,,lO000k:,,lxclO00000000000dl0MNdcO0000kolkN
MMMMMMMMMMMMMMMMXolO00000kolccxOlckxook0000Oxlokx:okdkkxkO00000dl0XxlxO000kloXWM
MMMMMMMMMMMMMMMMNdlO00000OOd:lO0xclk00000000O0Oxclkkc;loxO0000OocoookO000OolKMMM
MMMMMMMMMMMMMMMMMOcx00000000xclkOkocldkOOOOOxdlldOOocoOO000000Oc,lxO00000kcxWMMM
MMMMMMMMMMMMMMMMMNdlO000000O0kocokOkdoooooooooxOkdllxO000O000Od:oO00O000OooXMMMM
MMMMMMMMMMMMMMMMMMXolk00000O00OkollodxkkkkkkkxdlllxO00000OOOOd:oO000O00OdcOMMMMM
MMMMMMMMMMMMMMMMMMM0::xO0000000O0OkdooooooooooodkO000000000OocoO000O000xlxWMMMMM
MMMMMMMMMMMMMMMMMNOoccclxO00O000000000OOOOOOO00000000OO00OdclxO00O0000kldNMMMMMM
MMMMMMMMMMMMMMMWKdlxOOkolldkO0000000000000000000000000OxolldO0O000000kldNMMMMMMM
MMMMMMMMMMMMMMWOldO00O00OxollodxOO000000000000000OOkdolloxO000O0000OxlxNMMMMMMMM
MMMMMMMMMMMMMWkldO000O00000OkdoooooooodddddddoooooooodkO000OO0O000OdlOWMMMMMMMMM
MMMMMMMMMMMMWOldO00000000000000OOkxxddooooooooddxkO0000O000000O0kdoxKWMMMMMMMMMM
MMMMMMMMMMMMXooO000000O0000000O0000000000000000OO00000OO000000kooxKWMMMMMMMMMMMM
MMMMMMMMMMMWxck00000000000000O000000000000000000O00OO000000O00dcOMMMMMMMMMMMMMMM
MMMWKkkOKWMXooO000000OkkO000000000000000000000000O000000000O00xckMMMMMMMMMMMMMMM
WMKdlcccldKOcd000000OdcdO0000000000000000000000000000000000000klkMMMMMMMMMMMMMMM
odccocccdllolk000000d:oO00000000000000000000000000000000000000OlxWMMMMMMMMMMMMMM
;:;:c::oxllolk00000kclO000000000000000000000000000000000000000OcdWMMMMMMMMMMMMMM
Oollodxdl;lc:x0000Oo:x0000000000000000000000000000000000000000kcxWMMMMMMMMMMMMMM
MN0kxxo::odl:d0000Olck000000000000000000000000000000000000O000dcOMMMMMMMMMMMMMMM
MMMMMMWKkddo;cO000Olck0000000000000000000000000000000000000O0OloNMMMMMMMMMMMMMMM
MMMMMMMMMWWOcoO000OkclO0000000000000000000000000000000000000Ool0MMMMMMMMMMMMMMMM
MMMMMMMMMMWxck0OOO0Olck00O00000000000000000000000000000O000Ool0WMMMMMMMMMMMMMMMM
MMMMMMMMMMM0loxolool:oO000000O000000000000000000000000000OxldKMMMMMMMMMMMMMMMMMM
MMMMMMMMMMMWKkxkKN0o;ck0000O000000000000000000000000OO0OxooONMMMMMMMMMMMMMMMMMMM
MMMMMMMMMMMMMMMMMMMWxck000O000OkkO000000000000000000Oxo:;xNMMMMMMMMMMMMMMMMMMMMM
MMMMMMMMMMMMMMMMMMMMOcx0000000kloO0000000000000Okxoollolc0MMMMMMMMMMMMMMMMMMMMMM
MMMMMMMMMMMMMMMMMMMMKld000O000d;codddddddddoooclooodkO0olKMMMMMMMMMMMMMMMMMMMMMM
MMMMMMMMMMMMMMMMMMMMNolO00000Oll00OOkkkkkkkO0OllO000O0OloNMMMMMMMMMMMMMMMMMMMMMM
MMMMMMMMMMMMMMMMMMMMWxck00000kckMMMMMMMMMMMMMMkcx00000kcxWMMMMMMMMMMMMMMMMMMMMMM
MMMMMMMMMMMMMMMMMMMMMkcx00000dcOMMMMMMMMMMMMMMKld00O00xckMMMMMMMMMMMMMMMMMMMMMMM
MMMMMMMMMMMMMMMMMMMMWdlk00000kckWMMMMMMMMMMMMM0ld0OOO0kldNMMMMMMMMMMMMMMMMMMMMMM
MMMMMMMMMMMMMMMMMMMMKlo0OkkO0OloNMMMMMMMMMMMMWxck0OOOO0dc0MMMMMMMMMMMMMMMMMMMMMM
MMMMMMMMMMMMMMMMMMMMXocdoollxxcdNMMMMMMMMMMMMMkcxOdlloxlc0MMMMMMMMMMMMMMMMMMMMMM
MMMMMMMMMMMMMMMMMMMMMKolxXXxc:oKMMMMMMMMMMMMMMKl:::x0dcckNMMMMMMMMMMMMMMMMMMMMMM
```

perses/app/cli.py
```diff
@@ -0,0 +1,159 @@
...
+        """
+
+
+def _check_openeye_license():
+    import openeye
+
+    assert openeye.oechem.OEChemIsLicensed(), "OpenEye license checks failed!"
+
+
+def _test_gpu(platform_name):
```
jchodera, line 39: This could just be `_test_platform`

perses/app/cli.py
```diff
@@ -0,0 +1,159 @@
...
+    from openmm.testInstallation import main
+
+    main()
```
jchodera, line 42: Suggest not calling an external `main()` without qualification:
```python
import openmm
openmm.testInstallation.main()
```

perses/app/cli.py
```diff
@@ -0,0 +1,159 @@
...
+    if platform_name:
+
+        test_system = HarmonicOscillator()
+        integrator = integrators.LangevinIntegrator(
+            temperature=298.0 * unit.kelvin,
+            collision_rate=1.0 / unit.picoseconds,
+            timestep=1.0 * unit.femtoseconds,
+        )
+        try:
+            platform = openmm.Platform.getPlatformByName(platform_name)
```
jchodera, line 54: I think you want to use [`openmmtools.utils.platform_supports_precision`](https://github.com/choderalab/openmmtools/blob/main/openmmtools/utils.py#L517) instead since this will not check whether the platform supports `mixed` precision:
```python
import openmmtools
result = openmmtools.utils.platform_supports_precision('mixed')
```
This will return `True` if `mixed` precision is supported.

You could alternatively just borrow some of the code from there instead.

perses/app/cli.py
```diff
@@ -0,0 +1,159 @@
...
+    """test"""
+    click.echo(click.style(percy, fg="bright_magenta"))
+    click.echo("📖\t Fetching simulation options ")
+    options = getSetupOptions(yaml)
+    click.echo("🖨️\t Printing options")
+    click.echo(options)
+    click.echo("🕵️\t Checking OpenEye license")
+    _check_openeye_license()
+    click.echo("✅\t OpenEye license good")
+    click.echo("🖥️⚡\t Making a test system to check if we can get a GPU")
```
jchodera, line 149: Suggest: `Checking whether requested compute platform is available`

jchodera (approved):
Looks good! Just some minor suggestions. Merge when addressed!

perses/app/cli.py
mikemhenry: I like it!

perses/app/cli.py
mikemhenry: Okay so that is much cleaner, but looks like I've got a bug in openmmtools to fix now:

```  
File "/home/mmh/Projects/perses/perses/app/cli.py", line 75, in _test_platform
    assert openmmtools.utils.platform_supports_precision(platform_name, 'mixed')
  File "/home/mmh/miniconda3/envs/perses/lib/python3.8/site-packages/openmmtools/utils.py", line 537, in platform_supports_precision
    platform = openmm.Platform.getPlatformByName(platform)
UnboundLocalError: local variable 'openmm' referenced before assignment
```

jchodera: Oh no! I don't understand how this didn't have test coverage. Can you be sure to add a test if you find you do indeed need to add the bugfix?

mikemhenry: Yes my plan is to figure out what tests are missing, add the tests, then fix the code. TDD in action!

jchodera: @ijpulidos @mikemhenry : Could this bug impact the new perses release too? Do we need to get the fix in before the new perses will work?

mikemhenry: Okay with some quick and dirty grep-ing, we don't use this utility in perses, and in openmmtools, the utility is also not used. I'm working on the fix now, but it looks like the test didn't ever try and use a string to set the platform name, but instead passed in an already instantiated openmm platform, so the branch here:

```
    if isinstance(platform, str):
        # Get the actual Platform object if the platform_name was specified
        platform = openmm.Platform.getPlatformByName(platform)
```
Doesn't have any test coverage.

ijpulidos: What am I missing here? We **are** importing openmm in the [first few lines of this code](https://github.com/choderalab/openmmtools/blob/7e6926f4719a331d06a66bc26be41d77779c3029/openmmtools/utils.py#L34-L38), aren't we? Why it cannot find the reference to openmm in this case?

mikemhenry: because this is weird but how python works :upside_down_face: 

```
>>> import sys
>>> def foo():
...  print(sys)
...  if True:
...   import sys
... 
>>> foo()
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 2, in foo
UnboundLocalError: local variable 'sys' referenced before assignment
>>>
```

mikemhenry: Importing openmm down here confused python
https://github.com/choderalab/openmmtools/blob/main/openmmtools/utils.py#L547-L550

ijpulidos: Right, thanks. Got it!

mikemhenry:
Once https://github.com/conda-forge/openmmtools-feedstock/pull/13 is merged, this should be good to go! Might just have to wait for the CDN to catch up

ijpulidos:
Hmm interesting, any ideas why is it taking so much more in tests now? Specially with [perses/tests/test_resume.py::test_resume_protein_mutation_no_checkpoint](https://github.com/choderalab/perses/runs/5326885279?check_suite_focus=true#step:9:329), compared to what it was spending before [here](https://github.com/choderalab/perses/runs/5099918214?check_suite_focus=true#step:9:331)?  I know some [testing concurrency changed](https://github.com/choderalab/perses/pull/934/files#diff-b268bb7dc66e8638921e223098a11eabb92b424875ec72a19d6145dc1efbe3f2R14-R17) in this PR, but not sure if that can explain it. I'm having a hard time trying to run tests locally.

mikemhenry:
Lets just make it an issue and address it later, I'd like to get 0.9.4 out the door today

PR #938, opened 2022-02-10
Use energy and sampler context caches in simulations

### LINKED ISSUE #900: Set reassign_velocities=True so that velocities are not zeroed when resuming a repex simulation
Related to https://github.com/choderalab/openmmtools/issues/531

We should go through the small molecule setup pipeline and make sure that `reassign_velocities=True` so that when we resume a simulation, the velocities are do not start at 0.

Example of line that needs to be changed: https://github.com/choderalab/perses/blob/master/perses/app/setup_relative_calculation.py#L642

### LINKED ISSUE #902: Fix default compute energies contexcache in Openmmtools
WIP - in `MultiStateSampler`

* Extend constructor to include `energies_context_cache=None` option.  If `None` create a ContextCache that mimics the global but isn't the same.
* Check cycling is not present.
* Run all unit tests on GPU.

### LINKED ISSUE #904: ContextCache memory issue with resuming simulations
cc https://github.com/choderalab/openmmtools/issues/536

### LINKED ISSUE #936: perses app that uses any MultiStateSampler (including ReplicaExchangeSampler and subclasses) should minimize before simulation
Right now, none of the perses apps that use`ReplicaExchangeSampler` and friends call `sampler.minmize()` before they start simulation. We should change this, since it is essential for ensuring stability of the resulting simulations.

Minimization should occur by default, but should be something we can override via the API or YAML.

Note that we should not re-minimize while resuming, however.

ijpulidos:
## Description
Making perses use the new way of specifying context cache from `openmmtools`. This means setting up the simulations to use the energy and sampler context caches, respectively.

## Motivation and context

* Avoid cycling issue by specifying context caches. 
* Fixes simulation cooling down on resume.
* Minimizes systems before equilibration and production, for repex and sams.

Resolves #902 
Resolves #904 
Resolves #936 
Resolves #900 

## Change log

```
* Using new streamlined way of dealing with context caches from openmmtools
* Avoid cool down of simulation on resume.
```

### DIFF
--- a/perses/app/setup_relative_calculation.py
+++ b/perses/app/setup_relative_calculation.py
@@ -676,7 +676,7 @@ def run_setup(setup_options, serialize_systems=True, build_samplers=True):
                         timestep=timestep,
                         collision_rate=1.0 / unit.picosecond,
                         n_steps=n_steps_per_move_application,
-                        reassign_velocities=False,
+                        reassign_velocities=True,
                         n_restart_attempts=20, constraint_tolerance=1e-06,
                         context_cache=cache.ContextCache(capacity=None, time_to_live=None)),
                         hybrid_factory=htf[phase], online_analysis_interval=setup_options['offline-freq'],
@@ -693,10 +693,10 @@ def run_setup(setup_options, serialize_systems=True, build_samplers=True):
                         timestep=timestep,
                         collision_rate=1.0 / unit.picosecond,
                         n_steps=n_steps_per_move_application,
-                        reassign_velocities=False,
+                        reassign_velocities=True,
                         n_restart_attempts=20, constraint_tolerance=1e-06,
                         context_cache=cache.ContextCache(capacity=None, time_to_live=None)),
-                        hybrid_factory=htf[phase] ,online_analysis_interval=setup_options['offline-freq'],
+                        hybrid_factory=htf[phase], online_analysis_interval=setup_options['offline-freq'],
                     )
                     hss[phase].setup(n_states=n_states, temperature=temperature, storage_file=reporter,
                                      lambda_protocol=lambda_protocol, endstates=endstates)
--- a/perses/app/setup_relative_calculation.py
+++ b/perses/app/setup_relative_calculation.py
@@ -343,6 +343,7 @@ def get_openmm_platform(platform_name=None):
 
     return platform
 
+
 def run_setup(setup_options, serialize_systems=True, build_samplers=True):
     """
     Run the setup pipeline and return the relevant setup objects based on a yaml input file.
@@ -669,7 +670,7 @@ def run_setup(setup_options, serialize_systems=True, build_samplers=True):
                 platform = get_openmm_platform(platform_name=None)
                 # Setup context caches for multistate samplers
                 energy_context_cache = cache.ContextCache(capacity=None, time_to_live=None, platform=platform)
-                propagation_context_cache = cache.ContextCache(capacity=None, time_to_live=None, platform=platform)
+                sampler_context_cache = cache.ContextCache(capacity=None, time_to_live=None, platform=platform)
 
                 if setup_options['fe_type'] == 'sams':
                     hss[phase] = HybridSAMSSampler(mcmc_moves=mcmc.LangevinSplittingDynamicsMove(
@@ -677,8 +678,7 @@ def run_setup(setup_options, serialize_systems=True, build_samplers=True):
                         collision_rate=1.0 / unit.picosecond,
                         n_steps=n_steps_per_move_application,
                         reassign_velocities=True,
-                        n_restart_attempts=20, constraint_tolerance=1e-06,
-                        context_cache=cache.ContextCache(capacity=None, time_to_live=None)),
+                        n_restart_attempts=20, constraint_tolerance=1e-06),
                         hybrid_factory=htf[phase], online_analysis_interval=setup_options['offline-freq'],
                         online_analysis_minimum_iterations=10, flatness_criteria=setup_options['flatness-criteria'],
                         gamma0=setup_options['gamma0']
@@ -687,22 +687,21 @@ def run_setup(setup_options, serialize_systems=True, build_samplers=True):
                                      storage_file=reporter, lambda_protocol=lambda_protocol, endstates=endstates)
                     # We need to specify contexts AFTER setup
                     hss[phase].energy_context_cache = energy_context_cache
-                    hss[phase].sampler_context_cache = propagation_context_cache
+                    hss[phase].sampler_context_cache = sampler_context_cache
                 elif setup_options['fe_type'] == 'repex':
                     hss[phase] = HybridRepexSampler(mcmc_moves=mcmc.LangevinSplittingDynamicsMove(
                         timestep=timestep,
                         collision_rate=1.0 / unit.picosecond,
                         n_steps=n_steps_per_move_application,
                         reassign_velocities=True,
-                        n_restart_attempts=20, constraint_tolerance=1e-06,
-                        context_cache=cache.ContextCache(capacity=None, time_to_live=None)),
+                        n_restart_attempts=20, constraint_tolerance=1e-06),
                         hybrid_factory=htf[phase], online_analysis_interval=setup_options['offline-freq'],
                     )
                     hss[phase].setup(n_states=n_states, temperature=temperature, storage_file=reporter,
                                      lambda_protocol=lambda_protocol, endstates=endstates)
                     # We need to specify contexts AFTER setup
                     hss[phase].energy_context_cache = energy_context_cache
-                    hss[phase].sampler_context_cache = propagation_context_cache
+                    hss[phase].sampler_context_cache = sampler_context_cache
             else:
                 _logger.info(f"omitting sampler construction")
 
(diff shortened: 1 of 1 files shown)


perses/app/setup_relative_calculation.py
```diff
@@ -599,6 +618,9 @@ def run_setup(setup_options, serialize_systems=True, build_samplers=True):
             #TODO expose more of these options in input
             if build_samplers:
 
+                # import module for dealing with ContextCache
+                from openmmtools.cache import ContextCache
```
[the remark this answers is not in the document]
ijpulidos: Not really, I thought we could avoid importing many dependencies but I guess we are already importing openmmtools, so no real reason to do that inside the function, I agree.

jchodera (commented):
Don't forget to also update the `openmmtools` version dependency in `devtools/` in the conda recipe!

perses/app/setup_relative_calculation.py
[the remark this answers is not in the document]
jchodera: I'd move this import to _just_ before we use `ContextCache` below

perses/app/setup_relative_calculation.py
```diff
@@ -906,6 +938,13 @@ def run(yaml_filename=None):
                 _logger.info(f"\t\tFinished phase {phase}")
 
 def _resume_run(setup_options):
+    from openmmtools.cache import ContextCache
+    # get platform
+    platform = get_openmm_platform(platform_name=None)
+    # Setup context caches for multistate samplers
+    energy_context_cache = ContextCache(capacity=None, time_to_live=None, platform=platform)
+    sampler_context_cache = ContextCache(capacity=None, time_to_live=None, platform=platform)
```
jchodera, line 946: This should be `propagation_context_cache`, not `sampler_context_cache`.

perses/app/setup_relative_calculation.py
```diff
@@ -921,6 +960,9 @@ def _resume_run(setup_options):
             total_steps = setup_options['n_cycles']
             run_so_far = simulation.iteration
             left_to_do = total_steps - run_so_far
+            # set context caches
+            simulation.sampler_context_cache = sampler_context_cache
```
jchodera, line 964: This should be `propagation_context_cache`, not `sampler_context_cache`.

perses/app/setup_relative_calculation.py
```diff
@@ -906,6 +938,13 @@ def run(yaml_filename=None):
                 _logger.info(f"\t\tFinished phase {phase}")
 
 def _resume_run(setup_options):
+    from openmmtools.cache import ContextCache
+    # get platform
+    platform = get_openmm_platform(platform_name=None)
```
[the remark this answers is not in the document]
ijpulidos: I will create the issue for the milestone for this, if we agree.

perses/app/setup_relative_calculation.py
ijpulidos: It is called `sampler_context_cache` on the openmmtools side, I used to call it `propagation_context_cache` but in some discussion this was changed. We could settle on one and make a bugfix release for `openmmtools` as well.

perses/app/setup_relative_calculation.py
```diff
@@ -312,6 +312,25 @@ def getSetupOptions(filename):
 
     return setup_options
 
+def get_openmm_platform(platform_name):
```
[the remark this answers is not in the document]
ijpulidos: Yes, we should do that.

perses/app/setup_relative_calculation.py
```diff
@@ -942,12 +996,16 @@ def _resume_run(setup_options):
             total_steps = setup_options['n_cycles']
             run_so_far = simulation.iteration
             left_to_do = total_steps - run_so_far
+            # set context caches
+            simulation.sampler_context_cache = propagation_context_cache
```
jchodera, line 1000: This should be `simulation.propagation_context_cache`

ijpulidos: It is called `sampler_context_cache` on the openmmtools side (we changed along the way and even if the behavior is consistent the naming should be improved). Check https://github.com/choderalab/perses/pull/938#discussion_r804111821 

I will change the `openmmtools` side to have the same consistent name `propagation_context_cache` and make a bugfix release (that would be `0.21.1`), once that's done we can merge this one and pin perses to that openmmtools release. Does that sound ok?

ijpulidos:
@jchodera @mikemhenry This one should be good to go. Please review it when you have the chance. Thanks!

I know the naming inconsistencies are still there, but as I said in the comments, these are from the `openmmtools` side of things and if we change them now we would break compatibility, unless we support now two names for the same attributes. This happened because along the way of the `openmmtools` refactor we changed the name in one of our discussions. That said, it should behave consistently (I have double checked), only that it might not have the best name for the attribute. Please let me know what you think about that, I guess we could also do something like `MultiStateSampler.old_name = MultiStateSampler.new_name`, which is a bit dirty.

perses/app/setup_relative_calculation.py
```diff
@@ -634,26 +665,44 @@ def run_setup(setup_options, serialize_systems=True, build_samplers=True):
...
-                                                   online_analysis_minimum_iterations=10,flatness_criteria=setup_options['flatness-criteria'],
-                                                   gamma0=setup_options['gamma0'])
-                    hss[phase].setup(n_states=n_states, n_replicas=n_replicas, temperature=temperature,storage_file=reporter,lambda_protocol=lambda_protocol,endstates=endstates)
+                    hss[phase] = HybridSAMSSampler(mcmc_moves=mcmc.LangevinSplittingDynamicsMove(
+                        timestep=timestep,
+                        collision_rate=1.0 / unit.picosecond,
+                        n_steps=n_steps_per_move_application,
+                        reassign_velocities=False,
+                        n_restart_attempts=20, constraint_tolerance=1e-06,
+                        context_cache=cache.ContextCache(capacity=None, time_to_live=None)),
```
jchodera, line 681: We no longer provide a `context_cache` argument to `MCMCMove`. Recall that we now [issue a warning if this happens](https://github.com/choderalab/openmmtools/blob/main/openmmtools/mcmc.py#L149-L152) since we now feed the `ContextCache` argument to `MCMCMove.apply()`.

perses/app/setup_relative_calculation.py
```diff
@@ -634,26 +665,44 @@ def run_setup(setup_options, serialize_systems=True, build_samplers=True):
...
-                                                                                        context_cache=cache.ContextCache(capacity=None, time_to_live=None)),
-                                                   hybrid_factory=htf[phase], online_analysis_interval=setup_options['offline-freq'],
-                                                   online_analysis_minimum_iterations=10,flatness_criteria=setup_options['flatness-criteria'],
-                                                   gamma0=setup_options['gamma0'])
-                    hss[phase].setup(n_states=n_states, n_replicas=n_replicas, temperature=temperature,storage_file=reporter,lambda_protocol=lambda_protocol,endstates=endstates)
+                    hss[phase] = HybridSAMSSampler(mcmc_moves=mcmc.LangevinSplittingDynamicsMove(
+                        timestep=timestep,
+                        collision_rate=1.0 / unit.picosecond,
+                        n_steps=n_steps_per_move_application,
+                        reassign_velocities=False,
```
jchodera, line 679: Let's set `reassign_velocities=True` to fix the issue with the simulation becoming too cold when resuming (#900).

perses/app/setup_relative_calculation.py
jchodera: Why did we use `propagation_context_cache` for the keyword argument and `sampler_context_cache` for the class attribute? That is just ... confusing.

Please do the following:
* Create a new release milestone 0.21.1 for openmmtools
* Create the issue to clean up the context cache API for multistate samplers.
* To avoid any chance for confusion, **please remove all keyword arguments dealing with the context caches** from the `openmmtools` `MultiStateSampler` methods. My initial inclination was that it was a mistake to put them in because they made the API confusing and untidy, and this is confirmed by even _us_ getting confused about what they should be named. The only way users will set these is via setting the class attributes after the object is created. 
* Tag me to review the PR
* Only once I've reviewed the PR, we can merge it and cut a new release
* Then we can come back to cleaning up the perses release

Please create a new release mils

mikemhenry: Milestone here:
https://github.com/choderalab/openmmtools/milestone/5

Issue here:
https://github.com/choderalab/openmmtools/issues/546

jchodera: Thanks!

ijpulidos:
@jchodera @mikemhenry This one should ready for review again. Note that it picked up the latest openmmtools release in the [CI tests](https://github.com/choderalab/perses/runs/5208120172?check_suite_focus=true#step:7:267). Thanks!

perses/app/setup_relative_calculation.py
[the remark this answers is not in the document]
ijpulidos: Created the issue https://github.com/choderalab/perses/issues/939 

At this point it is defaulting to `None` and picking the best platform with mixed precision. But it will be just adding the CLI/YAML option and make `get_open_platform` use that once we have it.

It is not a big change, but it just doesn't feel like a "bugfix". If we think we should change this in this bugfix release, let me know. Thanks!

jchodera (approved):
Looks great! Should be ready to merge!
Let's make sure to re-run the tyk2 benchmark to attach to the release notes.

PR #944, opened 2022-02-24
Fixes to visualization module

zhang-ivy:
## Description

Make fixes to the visualization module related to:
- Pymol frames are 1-indexed, but the module had been assuming it was 0-indexed
- Bug: `self._unique_old/new` should be `self._unique_old/new_selection`
- Renaming an object after loading isn't working, so instead, just load the objects as `old` and `new` directly
- Allow color of old residue to be different from the color of the new residue
- Change background protein transparency
- Remove "hide spheres" when we are using stick representation (shouldn't be necessary)
- Do not allow stick/sphere transparencies to fade to transparency 1.0 (completely transparent) -- make the max transparency 0.7, as there are weird visual artifacts when it is 1.0 (the stick/sphere looks white)
- Clarifications to comments/docstrings

## Motivation and context

Resolves #???

## Change log

```

```

### DIFF
--- a/perses/analysis/visualization.py
+++ b/perses/analysis/visualization.py
@@ -70,7 +70,7 @@ class Visualization(object):
     v.load()
     v.format(zoom_distance=3, rotate_x_angle=270, rotate_y_angle=180, rotate_z_angle=270)
     v.save_frames(outfile_prefix="output/frame")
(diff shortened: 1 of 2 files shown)


jchodera (approved):
Thanks so much for improving this, @zhang-ivy!

PR #953, opened 2022-03-04
Add more padding to fix issue #949

### LINKED ISSUE #949: E       openmm.OpenMMException: NonbondedForce: The cutoff distance cannot be greater than half the periodic box size.
Two days ago we started getting this error on nightly
https://github.com/choderalab/perses/runs/5367830989?check_suite_focus=true#step:9:293

`openmm.OpenMMException: NonbondedForce: The cutoff distance cannot be greater than half the periodic box size`

Which is causing these tests to fail:
```
FAILED perses/tests/test_relative.py::test_position_output - openmm.OpenMMExc...
FAILED perses/tests/test_relative.py::test_generate_endpoint_thermodynamic_states
FAILED perses/tests/test_relative.py::test_HybridTopologyFactory_energies - o...
FAILED perses/tests/test_relative.py::test_RepartitionedHybridTopologyFactory_energies
FAILED perses/tests/test_relative_point_mutation_setup.py::test_PointMutationExecutor
FAILED perses/tests/test_relative_point_mutation_setup.py::test_PointMutationExecutor_endstate_validation
FAILED perses/tests/test_rest.py::test_bookkeeping - openmm.OpenMMException: ...
FAILED perses/tests/test_rest.py::test_energy_scaling - openmm.OpenMMExceptio...
FAILED perses/tests/test_resume.py::test_resume_protein_mutation_no_checkpoint
FAILED perses/tests/test_topology_proposal.py::test_protein_counterion_topology_fix_positive
FAILED perses/tests/test_topology_proposal.py::test_protein_counterion_topology_fix_negitive
FAILED perses/tests/test_topology_proposal.py::test_protein_counterion_topology_fix_zero
FAILED perses/tests/test_examples.py::test_examples[/home/runner/work/perses/perses/examples/protein-neq-switching/run_example.py]
```

Which is a bit confusing since that line of code has been in openmm for 12 years:
https://github.com/openmm/openmm/blame/fd1cfdd60ccc883d4b811da96b823cb403182635/openmmapi/src/NonbondedForceImpl.cpp#L130

But looking at merged PRs and the dates, this may be the culprit:

https://github.com/openmm/openmm/pull/3480

Any ideas @jchodera 
It looks like this is enough to trigger the error:
`topology_proposal, old_positions, new_positions = utils.generate_solvated_hybrid_test_topology()`

mikemhenry:
## Description

Fix issue #949 

## Motivation and context

Due to a change in openmm, we need to increase the padding for our simulation boxes.

Resolves #949 

## How has this been tested?

Tested in CI

## Change log

```
Increase default padding in XXXX classes
```

I'm trying to increase the padding as little as possible in our tests/defaults to pass without increasing CI time significantly. So it is a bit slow/some trial and error to find that sweet spot.


jchodera:
Let's hold off on merging this until we decide on https://github.com/openmm/openmm/issues/3502

zhang-ivy:
Note from dev sync today: Mike says that some of the tests are still failing because the padding is still not large enough, but increasing the padding even more for all tests will slow the tests down. They are already taking > 45 min when beforehand, they took ~15 min.

mikemhenry:
So it looks like `setUseDispersionCorrection` is only a property of non-bonded forces (makes sense) and likewise it makes sense that `AttributeError: 'NonbondedForce' object has no attribute 'getNumTorsions'` which means that someone (on nightly) our forces are getting scrambled. 

I propose we make a new issue and merge this PR in to fix the issue with solvent padding. 

One open question, do we change the default padding in other parts of the code? For larger systems, the padding seems to work fine.

perses/tests/utils.py
```diff
@@ -522,7 +522,7 @@ def  generate_solvated_hybrid_test_topology(current_mol_name="naphthalene", prop
         hs = [atom for atom in modeller.topology.atoms() if atom.element.symbol in ['H'] and atom.residue.name not in ['MOL','OLD','NEW']]
         modeller.delete(hs)
         modeller.addHydrogens(forcefield=system_generator.forcefield)
-        modeller.addSolvent(system_generator.forcefield, model='tip3p', padding=9.0*unit.angstroms)
+        modeller.addSolvent(system_generator.forcefield, model='tip3p', padding=20.0*unit.angstroms)
```
ijpulidos, line 525: Why is it 20 A here instead of 11 A as it was with the other tests?

ijpulidos (commented):
I wonder if we should also change the default padding options in other places, for example [here](https://github.com/choderalab/perses/blob/59c77fbef45912d85afcf924d5386b33130b5fc1/perses/app/setup_relative_calculation.py#L83) or [here](https://github.com/choderalab/perses/blob/59c77fbef45912d85afcf924d5386b33130b5fc1/perses/app/relative_setup.py#L55) or [here](https://github.com/choderalab/perses/blob/59c77fbef45912d85afcf924d5386b33130b5fc1/perses/app/relative_setup.py#L817)

mikemhenry:
So since those systems are "real" they seem to work fine, since we run some of these examples through CI.

perses/tests/utils.py
mikemhenry: Good catch! I will double check that one!

zhang-ivy:
@mikemhenry : 
>Good news! I've bumped up the padding and the tests don't run really any slower 😎 I'm very happy to be wrong about the increased test times, which were due to some bad luck

Can you clarify why the test times were previously longer?

mikemhenry:
> @mikemhenry :
> 
> > Good news! I've bumped up the padding and the tests don't run really any slower sunglasses I'm very happy to be wrong about the increased test times, which were due to some bad luck
> 
> Can you clarify why the test times were previously longer?

I mention that here:
https://github.com/openmm/openmm/issues/3502#issuecomment-1083295956

I hadn't merged my fix to the multistate samplers which were causing the context cache to get a miss every time which slowed everything down.

zhang-ivy:
>I hadn't merged my fix to the multistate samplers which were causing the context cache to get a miss every time which slowed everything down.

@mikemhenry : I don't see the fix, could you point me to the PR?

mikemhenry:
> > I hadn't merged my fix to the multistate samplers which were causing the context cache to get a miss every time which slowed everything down.
> 
> @mikemhenry : I don't see the fix in the issue comment you linked, could you point me to the PR?

@zhang-ivy Here you go! https://github.com/choderalab/perses/pull/961/files

Because of this line https://github.com/choderalab/perses/blob/0.9.5/perses/samplers/multistate.py#L10
Importing that module caused weird behavior.

mikemhenry:
@ijpulidos I was able to reduce the padding in that one util from 20 A to 16 A but anything smaller fails.
Ready for re-reivew

zhang-ivy:
Noting that Peter has attempted to help resolve this with this PR: https://github.com/openmm/openmm/pull/3537 that John mentioned yesterday. 

I'm surprised that we need to increase padding at all, given this new PR..

ijpulidos:
So, from what I can see the problem is what Peter mentions [here](https://github.com/openmm/openmm/issues/3502#issuecomment-1068361109). Is it the case that Openmm is now making smaller boxes so we have the problem with the cutoff being greater than half of the box size. Then that's why increasing the padding works for these tests, we are just making larger boxes such that cutoff is smaller than half of its size. I think this is fine for the tests since they are actually fast tests (not a bottleneck), as far as I can see.

ijpulidos (approved):
This looks good, extra padding doesn't seem to affect the time for tests. Thanks!

zhang-ivy:
>So, from what I can see the problem is what Peter mentions here. Is it the case that Openmm is now making smaller boxes so we have the problem with the cutoff being greater than half of the box size. Then that's why increasing the padding works for these tests, we are just making larger boxes such that cutoff is smaller than half of its size. I think this is fine for the tests since they are actually fast tests (not a bottleneck), as far as I can see.

Yes, I agree with everything you said.. but in [this new PR](https://github.com/openmm/openmm/pull/3537) from peter. The description for the PR says: "First, it revises the method for selecting box size. When building a water box around a very small solute, it will never make the box width less than 2*padding."

My comment earlier today was referencing this revision. Peter's change should fix our problem, meaning we no longer need to increase the padding.

If there is no difference in the speed of the tests, then I guess it doesn't hurt to bump the padding, but in principle, we shouldn't have to, with Peter's revision.

ijpulidos:
@zhang-ivy Ah! I see what you mean. Yes, I agree. We probably just need the nightly builds to catch these changes then.

mikemhenry:
Last night's CI run is still throwing padding issues, and I really like to get our CI passing again. Given the low impact this has on our CI times, is this okay to merge @zhang-ivy @jchodera @ijpulidos ?

zhang-ivy:
@mikemhenry : I'm still not sure why we need to increase the padding. Happy to be overrridden on this decision, but see my comment here: https://github.com/choderalab/perses/pull/953#issuecomment-1089183580

jchodera:
@zhang-ivy is right that we _shouldn't_ need to change the padding. The [OpenMM nightlies](https://anaconda.org/omnia-dev/openmm/files) should be the latest `master` build, which includes https://github.com/openmm/openmm/pull/3537. 

The issue must be that https://github.com/openmm/openmm/pull/3537 is an incomplete fix. Perhaps we can open another PR starting from the perses `main`, figure out what tests are failing and why, and then give some feedback about how https://github.com/openmm/openmm/pull/3537 should be further refined?

mikemhenry:
I double checked https://github.com/choderalab/perses/runs/5872047384?check_suite_focus=true and we are running the latest nightly that was built last night, so https://github.com/openmm/openmm/pull/3537 is not a complete fix.

jchodera:
@mikemhenry Can you give me an example of a simple test that is failing due to the box size issues? I can dive into this in more detail and propose further fixes to the OpenMM feature.

mikemhenry:
```python

import numpy as np
from openmm import app, unit
from perses.tests.test_topology_proposal import (
    generate_atp, generate_dipeptide_top_pos_sys)

new_res = "ASP"
charge_diff = 1

# Make a vacuum system
atp, system_generator = generate_atp(phase="vacuum")

# Make a solvated system/topology/positions with modeller
modeller = app.Modeller(atp.topology, atp.positions)
modeller.addSolvent(
    system_generator.forcefield,
    model="tip3p",
    padding=9 * unit.angstroms,
    ionicStrength=0.15 * unit.molar,
)
solvated_topology = modeller.getTopology()
solvated_positions = modeller.getPositions()

atp.positions = unit.quantity.Quantity(
    value=np.array(
        [
            list(atom_pos)
            for atom_pos in solvated_positions.value_in_unit_system(unit.md_unit_system)
        ]
    ),
    unit=unit.nanometers,
)
atp.topology = solvated_topology

atp.system = system_generator.create_system(atp.topology)

# Make a topology proposal and generate new positions
top_proposal, new_pos, _, _ = generate_dipeptide_top_pos_sys(
    topology=atp.topology,
    new_res=new_res,
    system=atp.system,
    positions=atp.positions,
    system_generator=system_generator,
    conduct_geometry_prop=True,
    conduct_htf_prop=False,
    validate_energy_bookkeeping=True,
)
```

This fails with the currently nightly openmm build

jchodera:
Thanks, @mikemhenry! Filed a simpler version of this as https://github.com/openmm/openmm/issues/3502#issuecomment-1103890207

PR #955, opened 2022-03-05
Test PR to run CI

mikemhenry:
Going to see if this runs faster on GHA

### DIFF
--- a/foo
+++ b/foo
@@ -1,0 +1,42 @@
+* [new branch]        add-examples -> add-examples
+ * [new branch]        auto-example-test -> auto-example-test
+ * [new branch]        blest4 -> blest4
+ * [new branch]        counterion -> counterion
+ * [new branch]        feat/add-rc-openmm-to-ci -> feat/add-rc-openmm-to-ci
+ * [new branch]        feat/add-versioneer -> feat/add-versioneer
+ * [new branch]        feat/add_docker_support -> feat/add_docker_support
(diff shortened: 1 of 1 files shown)


mikemhenry:
Running this to see how long the `test_resume_protein_mutation_no_checkpoint` test takes, I'm getting really strange results locally so trying to see what happens on CI, but the tl;dr is that using commit 691d890 the test takes `1361.40s (0:22:41)` but using commit 10ec6d9 the test takes `70.63s (0:01:10)` and looking at the diff 10ec6d9...691d890 nothing really explains that! Locally this is easy to to test:

```
git checkout COMMIT_HASH
pytest -k test_resume_protein_mutation_no_checkpoint
```
Then you can checkout the other commit and see the difference. I've profiled the code and the difference is in how long openmm takes to run, which is even weirder. I've attached the profiles, but all they really tell me is openmm took longer to run, but why?
[fast.tar.gz](https://github.com/choderalab/perses/files/8189516/fast.tar.gz)
[slow.tar.gz](https://github.com/choderalab/perses/files/8189517/slow.tar.gz)

The profiles are .prof files and can be viewed with different tools, my fav is [snakeviz](https://jiffyclub.github.io/snakeviz/)

mikemhenry:
Next steps could be either adding or removing (or even using git bisect) to figure out which bit is actually making the difference.

PR #960, opened 2022-03-10
Feat/robust retry logic

### LINKED ISSUE #928: Improve robustness against OpenMMException (e.g. CUDA_ERROR_ILLEGAL_ADDRESS)
cc https://github.com/choderalab/perses/pull/924#issuecomment-1025060653

We should implement some level of protection against these failures. One option is to wrap the whole simulation in a try...except block. That might be the most conservative, safest option, since we don't know how many CUDA contexts need to be invalidated if this occurs.

mikemhenry:
## Description

When running MD, sometimes we get an error that is related to the hardware (GPU) and instead of just throwing an exception, we can try again.

## Motivation and context

Parts of the API are gross which is why I've copied and pasted the same bits everywhere, but used a function decorator where I could. 

Resolves #928 

## How has this been tested?

Tested on CI.

## Change log

```
Made simulation running more robust by re-trying the simulation when an OpenMMException is thrown
```


ijpulidos:
I think for this one it makes sense to write our own `retry` decorator and use that in the functions that need it, such as the multisampler `equilibrate()` or `minimize()` or similar.We can put it in `utils.py` and import it where we need in the code.

mikemhenry:
@ijpulidos I don't think that approach will work (at least until we refactor) since we probally want to control the retry paramters at function run time and not function defftion time.

mikemhenry:
We will take care of this in the refactor

PR #961, opened 2022-03-11
Improve CI speed

### LINKED ISSUE #945: Tests suddenly taking more time
Hmm interesting, any ideas why is it taking so much more in tests now? Specially with [perses/tests/test_resume.py::test_resume_protein_mutation_no_checkpoint](https://github.com/choderalab/perses/runs/5326885279?check_suite_focus=true#step:9:329), compared to what it was spending before [here](https://github.com/choderalab/perses/runs/5099918214?check_suite_focus=true#step:9:331).  I know some [testing concurrency changed](https://github.com/choderalab/perses/pull/934/files#diff-b268bb7dc66e8638921e223098a11eabb92b424875ec72a19d6145dc1efbe3f2R14-R17) in this PR, but not sure if that can explain it. I'm having a hard time trying to run tests locally.

_Originally posted by @ijpulidos in https://github.com/choderalab/perses/issues/934#issuecomment-1050419800_

mikemhenry:
## Motivation and context

Resolves #945 

## Change log

```

```

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -90,7 +90,7 @@ jobs:
         shell: bash -l {0}
         run: |
           export TRAVIS=true
-          pytest -v --cov-report xml --cov=perses --durations=0 -a "not advanced" -n auto -m "not gpu_needed" 
+          pytest -v --cov-report xml --cov=perses --durations=0 -a "not advanced" -n auto -m "not gpu_needed" perses/tests
 
       - name: Codecov
         if: ${{ github.repository == 'choderalab/perses'
--- a/perses/app/setup_relative_calculation.py
+++ b/perses/app/setup_relative_calculation.py
@@ -7,7 +7,6 @@ import simtk.unit as unit
 import logging
 from pathlib import Path
 
-from perses.samplers.multistate import HybridSAMSSampler, HybridRepexSampler
 from perses.annihilation.relative import HybridTopologyFactory
 from perses.app.relative_setup import RelativeFEPSetup
 from perses.annihilation.lambda_protocol import LambdaProtocol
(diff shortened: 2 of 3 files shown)


mikemhenry:
My goodness, how could this be what slows down an unrelated test? Going to investigate more, my hunch is that even when I only select one test, pytest has to read each test to find which ones exist, which would trigger the import of this module, which shouldn't have any side effects except importing other modules. Should be easy enough to narrow down...

mikemhenry:
Traced it down to this import `from perses.app.setup_relative_calculation import getSetupOptions` so now looking at that module to see what import is causing things to be so wack

mikemhenry:
Looks like it is then this import 
`from perses.samplers.multistate import HybridSAMSSampler, HybridRepexSampler`

mikemhenry:
Looking at `perses.samplers.multistate` every import is used in the module and the elsewhere so it was easy to comment things out. What did was move the imports of `HybridSAMSSampler` and `HybridRepexSampler` to exactly where they are needed and that speed the test back up to normal. We can talk about this Monday but the tl;dr is 

In `perses/tests/test_cli.py` we import `from perses.app.cli import cli`. That module has this import `from perses.app.setup_relative_calculation import getSetupOptions` which then takes us to `setup_relative_calculation` where we then import `HybridSAMSSampler` and `HybridRepexSampler`. Moving those imports into the functions that need them resulted in the `test_resume_protein_mutation_no_checkpoint` test running at normal speed. 


One tricky bit I found is that even when only running a subset of tests, pytest ends up running each test file, which will cause imports to happen. This makes sense, but took a bit to remember/figure out!

mikemhenry:
Need to add this as well 
https://github.com/choderalab/perses/pull/954

mikemhenry:
I think for this change to fix things, it must mean that, at least in a testing env, the imports of those classes cause the context cache to be messed up in a way that the cache acts like the dummy/uses the wrong platform, since I can't think of another reason openmm would be slow.

PR #962, opened 2022-03-11
Use AminoAcidAtomMapper and AminoAcidAtomMapping for protein mutation atom mappings

zhang-ivy:
## Description

Create `AminoAcidAtomMapper` and `AminoAcidAtomMapping` classes, based on `AtomMapper` and `AtomMapping`, which were written for small molecules. These classes use a rules (not 2D property or geometry)-based approach for generating protein mutation atom mappings. 

## Motivation and context

Resolves #???

## Change log

```

```


perses/rjmc/atom_mapping.py
```diff
@@ -1485,3 +1564,161 @@ def _assign_ring_id(oeobj, max_ring_size=10):
...
+
+                    except:
+                        no_direct_matches.append(atom)
+
+        # If an exact match doesn't exist for HB and CG atoms, try to find a non-exact match (i.e. disregarding the number in the atom name, e.g. CG = CG1)
+        for atom in no_direct_matches:
+            old_atom_index = atom.GetIdx()
+            old_atom_name = atom.GetName()
+
+            if old_atom_name in hb_atoms + cg_atoms:  # Do not try to find indirect matches for hg_atoms and cd_atoms
```
dominicrufa, line 1691: is there a reason for not finding these indirect matches?

dominicrufa: and can you put it in the code and comment it out?

perses/rjmc/topology_proposal.py
```diff
@@ -1115,18 +1116,18 @@ def _construct_atom_map(self,
...
+                            extra_sidechain_map=None):
         """
         Construct atom map (key: index to atom in new residue, value: index to atom in old residue) to supply as an argument to the TopologyProposal.
         
         By default, the atom map:
         1) Maps all backbone atoms (exception: for GLY, do not map HA2 and HA3)
-        2) Map CBs only (no other sidechain atoms)
-        
+        2) Map CBs, HBs, and CGs.
+        3) Also maps HG2s, HG3s, and CDs, but not CD1, CD2, HG, HG1, HG11, HG12, HG13, HG21, HG22, HG23 (as mapping the latter atoms may match dissimilar geometries)
```
dominicrufa, line 1126: this appears different from the logic that i thought we agreed upon? im not sure

perses/rjmc/atom_mapping.py
dominicrufa: nevermind; you clarified in the docs.

dominicrufa (dismissed):
i think the logic works; i wasn't able to find any edge cases with the exceptions you wrote in.

perses/rjmc/topology_proposal.py
zhang-ivy: How is it different than the logic we agreed upon?

zhang-ivy: I actually don't think we talked about this during our chat this week. Here, I am mapping up to CG atoms for most transformations. However, I also map CD and HG2 and HG3 if there are direct atom matches. There are only a few transformations where this is true -- transformations where the old and new residue are both in the set of {LYS, ARG, GLU, GLN}. I don't want to try to search for non-exact atom matches involving CD and HG atoms because there is a risk for mapping differing sidechain geometries to each other, e.g. I don't want to accidentally map CD from LYS onto the CD1 in TYR.

jchodera:
@zhang-ivy : Is this deprecated and should be closed, or does this still have to be updated and merged?
If it has to be updated and merged, what functionality requires it? Do we have this on the roadmap?

zhang-ivy:
@jchodera : Not sure. We may want to keep parts of this PR, as it allows for protein mutation atom mappings to be generated with the `AtomMapper` class, but we probably don't want to keep the default mappings as is, which is:
```
By default, the atom map:
        1) Maps all backbone atoms (exception: for GLY, do not map HA2 and HA3)        
        2) Map CBs, HBs, and CGs.
        3) Also maps HG2s, HG3s, and CDs, but not CD1, CD2, HG, HG1, HG11, HG12, HG13, HG21, HG22, HG23 (as mapping the latter atoms may match dissimilar geometries)
```

I'm currently generating atom maps using the [approach](https://github.com/choderalab/perses/blob/main/perses/rjmc/topology_proposal.py#L1060-L1189) in `main`, which does not use the `AtomMapper` class and by default maps up to the CBs (and does not map HBs). This seems to be working perfectly fine, but if we want the protein mutation code to use the `AtomMapper`, we should consider merging this PR at some point.

zhang-ivy:
Closing this since for protein mutations, we decided to map all backbone atoms and sidechain atoms up to the beta carbon

PR #967, opened 2022-03-15
Enable PointMutationExecutor input of solvated pdbs

zhang-ivy:
## Description

Allow the PointMutationExecutor to accept solvated PDBs (previously, it only accepted solute PDBs)

## Motivation and context

Resolves #???

## Change log

```

```

### DIFF
--- a/perses/app/relative_point_mutation_setup.py
+++ b/perses/app/relative_point_mutation_setup.py
@@ -118,7 +118,8 @@ class PointMutationExecutor(object):
             protein_filename : str
                 path to protein (to mutate); .pdb
                 Note: if there are nonstandard residues, the PDB should contain the standard residue name but the atoms/positions should correspond to the nonstandard residue. E.g. if I want to include HID, the PDB should contain HIS for the residue name, but the atoms should correspond to the atoms present in HID. You can use openmm.app.Modeller.addHydrogens() to generate a PDB like this. The same is true for the ligand_input, if its a PDB. 
-                Note: this can be the protein solute only or the solvated protein.
+                Note: this can be the protein solute only or the solvated protein. if its the former, solvate should be set to True.
+                if its the latter, solvate should be set to False.
             mutation_chain_id : str
                 name of the chain to be mutated
             mutation_residue_id : str
@@ -135,7 +136,8 @@ class PointMutationExecutor(object):
                 endstate validation cannot and will not be conducted.
             ligand_input : str, default None
                 path to ligand of interest (i.e. small molecule or protein)
-                Note: if this is not solvated, it should be the ligand alone (.sdf or .pdb). if solvated, this should be the protein + ligand (.pdb).
+                Note: if this is not solvated, it should be the ligand alone (.sdf or .pdb) and solvate should be set to True.
+                if this is solvated, this should be the protein-ligand complex (.pdb) and solvate should be set to False.
             ligand_index : int, default 0
                 which ligand to use
             allow_undefined_stereo_sdf : bool, default False
@@ -145,7 +147,8 @@ class PointMutationExecutor(object):
             demap_CBs : bool, default False
                 whether to remove CBs from the mapping
             solvate : bool, default True
-                whether to solvate the protein/complex
+                whether to solvate the protein/complex. If this is False, protein_filename and ligand_input should correspond
+                to the solvated protein PDB and solvated complex PDB, respectively.
             water_model : str, default 'tip3p'
                 solvent model to use for solvation
             ionic_strength : float * unit.molar, default 0.15 * unit.molar
--- a/perses/app/relative_point_mutation_setup.py
+++ b/perses/app/relative_point_mutation_setup.py
@@ -137,7 +137,8 @@ class PointMutationExecutor(object):
             ligand_input : str, default None
                 path to ligand of interest (i.e. small molecule or protein)
                 Note: if this is not solvated, it should be the ligand alone (.sdf or .pdb or .cif) and solvate should be set to True.
-                if this is solvated, this should be the protein-ligand complex (.pdb or .cif) and solvate should be set to False.
+                if this is solvated, this should be the protein-ligand complex (.pdb or .cif) -- with the protein to be mutated first and
+                the ligand second in the file -- and solvate should be set to False.
             ligand_index : int, default 0
                 which ligand to use
             allow_undefined_stereo_sdf : bool, default False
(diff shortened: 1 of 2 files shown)


perses/app/relative_point_mutation_setup.py
```diff
@@ -141,6 +144,8 @@ def __init__(self,
                 map of new to old sidechain atom indices to add to the default map (by default, we only map backbone atoms and CBs)
             demap_CBs : bool, default False
                 whether to remove CBs from the mapping
+            solvate : bool, default True
```
dominicrufa, line 147: can you clarify _when_ solvation happens. perses does solvate protein and protein:ligand systems. i think the difference is _when_ you are solvating, or rather, whether you have provided a pre-solvated pdb/structure file. perhaps this should change the name of the argument you are using. can you put into the docs that if `False`, it should be pre-solvated. Also, i was going to suggest adding a sanity check that will check if there are waters in the struct file/topology and raise an error if there are waters, but one has already specified `solvate=True`. Then again, if the pdb contains xtal waters that are important, that error would be erroneous. maybe adding a sanity exception error if there are lots of waters in the pdb structure.

perses/app/relative_point_mutation_setup.py
```diff
@@ -215,18 +220,22 @@ def __init__(self,
...
+                complex_positions[protein_n_atoms:, :] = ligand_positions
+
+                # Convert positions back to openmm vec3 objects
+                complex_positions_vec3 = []
+                for position in complex_positions:
+                    complex_positions_vec3.append(openmm.Vec3(*position.value_in_unit_system(unit.md_unit_system)))
+                complex_positions = unit.Quantity(value=complex_positions_vec3, unit=unit.nanometer)
+            else:
+                complex_topology = ligand_topology
+                complex_positions = ligand_positions
```
dominicrufa, line 238: the exception for pre-existing waters could go here i guess?

perses/app/relative_point_mutation_setup.py
```diff
@@ -238,11 +247,27 @@ def __init__(self,
...
+            positions_list.append(complex_positions)
+            box_dimensions_list.append(complex_box_dimensions)
+
+        for topology, positions, box_dimensions in zip(topology_list, positions_list, box_dimensions_list):
+            if solvate:
+                solvated_topology, solvated_positions = self._solvate(topology, positions, water_model, phase,
+                                                                      ionic_strength, box_dimensions)
+            else:
+                solvated_topology = topology
+                solvated_positions = unit.quantity.Quantity(value=np.array(
```
dominicrufa, line 266: there might be an issue here associated with the `solvate` function giving a tuple of posits rather than a list, but we'll see what the test says

perses/app/relative_point_mutation_setup.py
```diff
@@ -582,6 +607,5 @@ def _solvate(self,
 
         # Canonicalize the solvated positions: turn tuples into np.array
         solvated_positions = unit.quantity.Quantity(value=np.array([list(atom_pos) for atom_pos in solvated_positions.value_in_unit_system(unit.md_unit_system)]), unit=unit.nanometers)
-        solvated_system = self.system_generator.create_system(solvated_topology)
 
-        return solvated_topology, solvated_positions, solvated_system
+        return solvated_topology, solvated_positions
```
dominicrufa, line 611: changing the returnable can be problematic. can you return the solvated system, too?

perses/app/relative_point_mutation_setup.py
zhang-ivy: >can you put into the docs that if False, it should be pre-solvated.
yes

> Also, i was going to suggest adding a sanity check that will check if there are waters in the struct file/topology and raise an error if there are waters, but one has already specified solvate=True. Then again, if the pdb contains xtal waters that are important, that error would be erroneous.
Yes, I thought about throwing an error like this, but like you said, if there are crystal waters present, but the PDB is not actually solvated, the error would still be present, which we don't want.

perses/app/relative_point_mutation_setup.py
zhang-ivy: yes, that's the whole point of the `else` part of this logical branch. if the topology is already solvated, we're assuming its the full complex (not just the ligand), so we don't need to combine topologies anymore.

perses/app/relative_point_mutation_setup.py
zhang-ivy: the tests pass, so i don't think there is any problem with this.

perses/app/relative_point_mutation_setup.py
zhang-ivy: Can you elaborate on why you think changing the returned variables is problematic? This `_solvate()` function is only ever called in the `__init__` of PointMutationExecutor, and I made sure to properly adjust the code in the `__init__` to account for the change in returned variables. Therefore, I don't think its a problem.

I moved the code for creating the system to be outside of the _solvate function (see line 269). This way, if the users specifies a presolvated PDB as input, a system can still be generated without calling _solvate.

perses/tests/test_relative_point_mutation_setup.py
```diff
@@ -13,7 +13,7 @@ def test_PointMutationExecutor():
         ionic_strength=0.15 * unit.molar,
         flatten_torsions=True,
         flatten_exceptions=True,
-        conduct_endstate_validation=False,
+        conduct_endstate_validation=False
```
mikemhenry, line 16: Why remove the trailing comma here?

mikemhenry: and here https://github.com/choderalab/perses/pull/967/files#diff-f62f3ff16f52afacdf9163c3263cd309db5bf545f8d886c0eefd5720f8603917R35

mikemhenry, line 16:
```suggestion
        conduct_endstate_validation=False,
```

perses/tests/test_relative_point_mutation_setup.py
```diff
@@ -32,5 +32,31 @@ def test_PointMutationExecutor_endstate_validation():
         ionic_strength=0.15 * unit.molar,
         flatten_torsions=False,
         flatten_exceptions=False,
-        conduct_endstate_validation=True,
+        conduct_endstate_validation=True
```
mikemhenry, line 35:
```suggestion
        conduct_endstate_validation=True,
```

mikemhenry:
Oh my, I do not know how my review got posted so many times! @zhang-ivy the only small nit I have is to keep the trailing comma

mikemhenry:
> @mikemhenry : Ah thanks for the clarification -- I was wondering why you had left that in. I removed it because it seemed like you left it in by accident / syntactically its unnecessary to leave an extra comma when there are no arguments after. But if you'd prefer us to keep the comma in, feel free to add a suggested commit?

Done! Adding the comma makes the diffs cleaner/more obvious on what is being added to a function call/signature, but you are correct that they don't do anything functional AST wise!

zhang-ivy:
@mikemhenry : Thanks! Just messaged @dominicrufa  on slack and he said this is ok to merge -- please merge when you get a chance!

PR #969, opened 2022-03-16
Fix endstate energy validation test for flattened htfs

zhang-ivy:
## Description

`flattenedHybridTopologyFactory_energies` was not passing for transformations involving ring amino acids because `validate_energy_bookkeeping` was always set to `True` in the function. It should only be set to `True` if it doesn't involve the transformation does not involve a ring amino acid.

## Motivation and context

Resolves #???

## Change log

```

```

### DIFF
--- a/perses/tests/test_relative.py
+++ b/perses/tests/test_relative.py
@@ -855,6 +855,9 @@ def flattenedHybridTopologyFactory_energies(topology, chain, system, positions, 
 
         # Create geometry proposals
         for _ in range(5):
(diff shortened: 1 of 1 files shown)


dominicrufa (approved):
yes, this is the right treatment for ring amino acid bits; can you add a comment line on line 858 to  explain why this is happening for the sake of documentation? but i will approve

zhang-ivy:
@dominicrufa : Done! 

@mikemhenry @ijpulidos : This is ready to merge!

ijpulidos (approved):
Looks good!

Issue #966, opened 2022-03-15
Update example to use new (2.0) forcefield 


mikemhenry:
https://github.com/choderalab/perses/blob/main/examples/protein-ligand/cli/protein-ligand.yaml#L22

ijpulidos:
This was fixed by https://github.com/choderalab/perses/pull/971. Closing.

PR #971, opened 2022-03-18
Updates forcefield recommendation to use openff-2.0.0 instead of openff-1.0.0

mikemhenry:
fixes issue #966


ijpulidos (commented):
This looks great, thanks!

There's only one file missing for adding the change here https://github.com/choderalab/perses/blob/main/examples/kinase-neq-switching/run_example.py#L38

mikemhenry:
Good catch! I only looked for `openff-1.0.0`!

ijpulidos (approved):
This looks great, thanks!

mikemhenry:
Sorry for the review request, it was a mistake

PR #972, opened 2022-03-18
More CLI mocking

mikemhenry:
Still need to wire this up but now we have a first pass at overrides + a dummy real-time analysis yaml file created.

Really the only bit that is missing is how to do the analysis once you have ran the star map.


mikemhenry:
This is the spec document for this PR https://www.notion.so/choderalab/Integration-of-perses-into-Interline-infrastructure-4850c1884ff74cae86c5fd4c85b56164

ijpulidos:
@mikemhenry that will be finished soon, sorry for the delay. I will just add the test and docs that you suggested and merge it.

mikemhenry:
No worries at all! I still have some clean up to do on this PR re tests and overrides.

mikemhenry:
@ijpulidos if you could test the docker command in the readme file I added, that would be great!

perses/app/relative_setup.py
```diff
@@ -811,14 +811,14 @@ def _solvate_system(self, topology, positions, model='tip3p',phase='complex', bo
         #modeller.delete(hs)
         #modeller.addHydrogens(forcefield=self._system_generator.forcefield)
         _logger.info(f'box_dimensions: {box_dimensions}')
-        _logger.info(f'solvent padding: {self._padding._value}')
+        _logger.info(f'solvent padding: {self._padding}')
```
mikemhenry, line 814: I have no idea how our tests never caught this, but `self._padding` doesn't ahe a `_value` attr, ~2 year old bug?

ijpulidos:
@mikemhenry Just as a reminder from today's discussion. Check if we are able to see what's happening with the `openmmtools` logger not getting captured. This is not critical, so we can create an issue for this and tackle it in the future if needed.

FYI, the benchmark code does show the log by using the `run()` function from `setup_relative_calculation` [here](https://github.com/choderalab/perses/blob/59c77fbef45912d85afcf924d5386b33130b5fc1/benchmarks/run_benchmarks.py#L92), therefore it might be something in between that's stopping it from showing.

jchodera (approved):
This looks great!

Example YAML file conforms to current best practices.
Docs are nice and clear.

Thanks!

perses/app/setup_relative_calculation.py
```diff
@@ -739,7 +748,9 @@ def run(yaml_filename=None):
 
     _logger.info(f"Getting setup options from {yaml_filename}")
 
-    setup_options = getSetupOptions(yaml_filename)
+    # If a user doesn't pass in setup_options, generate them
+    if not setup_options:
+        setup_options = getSetupOptions(yaml_filename)
```
ijpulidos, line 753: I see an issue with this conditional, because you can end up shooting yourself in the foot rather easily when using Perses from the API, by calling the `run` function with any `setup_options` parameter. For example the following:

```python
from perses.app.setup_relative_calculation import getSetupOptions, run
run("my.yaml", {"phases": ["solvent", "complex"]})
```
This will end up in an error like `KeyError: 'trajectory_directory'` somewhere else in the code because not calling `getSetupOptions` will leave undefined many of the required options.

perses/app/cli.py
```diff
@@ -72,99 +84,30 @@ def _test_platform(platform_name):
...
     click.echo("🕵️\t Checking OpenEye license")
     _check_openeye_license()
     click.echo("✅\t OpenEye license good")
     click.echo("🖥️⚡\t Checking whether requested compute platform is available")
     _test_platform(platform_name)
-    click.echo("🖨️\t Writing out files")
-    trajectory_directory = options["trajectory_directory"]
-    _write_out_files(trajectory_directory, options)
+    click.echo("🏃\t Running simulation")
+    run(yaml_filename=yaml, setup_options=options)
```
ijpulidos, line 110: Shouldn't it be simpler and enough if we just add an `override_string` parameter to the `run` function itself? Because the latter still calls `getSetupOptions` and then that one would accept the `override_string` parameter and deal with it. That would mean we only have to call `run` here, and also it would solve potential issues in `run` mentioned in the comments for `setup_relative_calculation.py`, by always requiring to call `getSetupOptions`.

ijpulidos: Maybe `getSetupOptions` shouldn't be a public API point at all?

perses/app/setup_relative_calculation.py
ijpulidos: Changing this, by forcing to always call `getSetupOptions`, should also help with the issue commented in `cli.py` in this review.

perses/app/setup_relative_calculation.py
```diff
@@ -1023,6 +1034,42 @@ def _resume_run(setup_options):
...
+        # First lets see if we can make it a int:
+        try:
+            val = int(val)
+        except ValueError:
+            # Now try float
+            try:
+                val = float(val)
+            except ValueError:
+                # Just keep it a str
+                pass
```
ijpulidos, line 1067: Yeah, you are totally right, this is not pretty. I see a couple of issues here:
* Doing it this way could lead to undesirable behavior, think about the order of things being casted. If I pass a float here to get overwritten it will sucessfully be casted to an int, and I don't think we want to do this. 
* We are then limiting our CLI to be able to change only things that are either int, float or string. There are options that we might want to change that are booleans (such as `h_contraints`, `rmsd_restraint` or `render_atom_map`, to name a few).

To avoid having to deal with the logic ourselves, maybe using something like `pyyaml` (or maybe even `json`?) to actually parse the override string would be better, something like
```python
python_dict = yaml.safe_load(overrides)
```

examples/new-cli/README.md
```diff
@@ -0,0 +1,89 @@
...
+First, grab the dev image of perses that has the new CLI tool with `docker pull choderalab/perses:dev`.
+General docker instructions for using perses and docker can be found [here](https://github.com/choderalab/perses/tree/main/docker#readme).
+To run this example use the following docker command (to make it easier to read the command is split across multiple lines but this is not necessary).
+```bash
+docker run -it --rm --gpus device=0 --mount type=bind,source=$HOME/.OpenEye/,target=/openeye/,readonly \
+                                    --mount type=bind,source=$HOME/repos/perses/examples/,target=/mnt/ \
+                                    -w /mnt/new-cli choderalab/perses:dev bash ./run_star_map.sh
+```
+
+Of importance there are the paths to the OpenEye license file (in this example is `$HOME/.openeye/`), path to the examples directory in perses (`$HOME/Projects/perses/examples/`)
```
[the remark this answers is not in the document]
ijpulidos: The docker command runs locally! Great work!

ijpulidos (changes requested):
This is looking great. Thanks for the hard work. There are still a few issues I commented that I feel need to be discussed. Creating CLIs can be very complex!

examples/new-cli/README.md
```diff
@@ -0,0 +1,86 @@
...
+```yaml
+old_ligand_index: 0
+new_ligand_index: 1
+trajectory_directory: lig0to1
+```
+
+and you wanted to instead use ligand at index 2 for the old and ligand 3 for the new, you would run:
+
+```bash
+$ perses-cli --yaml my.yaml --override old_ligand_index:2 --override new_ligand_index:3 --override trajectory_directory:lig2to3
```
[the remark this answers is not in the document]
mikemhenry: I agree that in the future we can work out a way to use the full yaml namespace to specify more nested options

examples/new-cli/README.md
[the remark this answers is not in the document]
mikemhenry: I agree we should be consistent, this page here: https://github.com/choderalab/perses/tree/main/docker has a mix 
`type=bind,source=$HOME/.openeye/,target=/openeye/` and `type=bind,source=$HOME/.OpenEye/,target=/openeye/,readonly` I think since the opeye docs https://docs.eyesopen.com/toolkits/python/quickstart-python/license.html say `~USERNAME/.OpenEye` then we should change everything to match that. I will do that in this PR.

perses/app/setup_relative_calculation.py
mikemhenry: :100: agree but I think this is outside what this PR does. As I understand it, no one should use perses like this 

```
from perses.app.setup_relative_calculation import getSetupOptions, run
run("my.yaml", {"phases": ["solvent", "complex"]})
```

Either you make objects in the python API or use the CLI, but don't use the `setup_relative*.py` like that since they really are not setup to do that. I only did that because I needed a way to override options _after_ `getSetupOptions` does its thing.

perses/app/cli.py
mikemhenry: Really nothing in `setup_relative_calculation.py` should be public. I think I can re-work things so that we only have to call run. I used this approach since when this was just a demo, I never called `run` just `getSetupOptions`.

While I don't think we _need_ to do this, I think it will make things a bit more simple so I will make this modification.

perses/app/setup_relative_calculation.py
mikemhenry: So I agree that using a parser would be better, but then we have to have users either pass in properly formatted yaml/json from the cli OR we are still manipulating their input into yaml/json, we are still parsing it.

Passing in `true` or `false` shouldn't be a problem, I just tested it and things work fine. Let me know what doesn't work for you.

I thought I checked that `int(4.2)` would throw a value error. Ooops!  I pushed a fix though, first I check to see if `int(val)` and `float(val)`are the same object, that should fix the problem with ints getting truncated.

perses/app/cli.py
mikemhenry: Okay so I made the changes, but now there isn't an easy way to print the options + overrides before the simulation runs, is this change still worth it?

perses/app/setup_relative_calculation.py
ijpulidos: > Passing in `true` or `false` shouldn't be a problem, I just tested it and things work fine. Let me know what doesn't work for you.

Ah okay, yes, I just tested this and it seems that the yaml parser takes both `'True'` and `True` as the correct boolean option, so it shouldn't be a problem.

perses/app/cli.py
ijpulidos: Yes, I see and I agree! I was thinking about this because the benchmark code does use the `run()` function that way. Hence why it seemed simpler to me to try using just one function from the `setup_relative*.py`, even if we agree we shouldn't use any of them that way.

ijpulidos (approved):
Nice work, I think it is better this way!

mikemhenry:
Awesome! Now we just need to get the experimental data added to the SDF file and run a benchmark!

jchodera:
> Awesome! Now we just need to get the experimental data added to the SDF file and run a benchmark!

Apologies for the delay---the last few days were packed with meetings. Working on this now.

jchodera:
Something is systematically corrupting some of the YAML files generated by the CLI, producing this error when parsing:
```
  File "/lila/data/chodera/chodera/perses/perses/examples/new-cli/analyze-benchmark.py", line 327, in <module>
    perses_graph = get_perses_network_results(perses_results_basepath)
  File "/lila/data/chodera/chodera/perses/perses/examples/new-cli/analyze-benchmark.py", line 169, in get_perses_network_results
    perses_input = yaml.safe_load(infile)
  File "/lila/home/chodera/miniconda/envs/perses-dev/lib/python3.9/site-packages/yaml/__init__.py", line 125, in safe_load
    return load(stream, SafeLoader)
  File "/lila/home/chodera/miniconda/envs/perses-dev/lib/python3.9/site-packages/yaml/__init__.py", line 81, in load
    return loader.get_single_data()
  File "/lila/home/chodera/miniconda/envs/perses-dev/lib/python3.9/site-packages/yaml/constructor.py", line 49, in get_single_data
    node = self.get_single_node()
  File "/lila/home/chodera/miniconda/envs/perses-dev/lib/python3.9/site-packages/yaml/composer.py", line 36, in get_single_node
    document = self.compose_document()
  File "/lila/home/chodera/miniconda/envs/perses-dev/lib/python3.9/site-packages/yaml/composer.py", line 55, in compose_document
    node = self.compose_node(None, None)
  File "/lila/home/chodera/miniconda/envs/perses-dev/lib/python3.9/site-packages/yaml/composer.py", line 84, in compose_node
    node = self.compose_mapping_node(anchor)
  File "/lila/home/chodera/miniconda/envs/perses-dev/lib/python3.9/site-packages/yaml/composer.py", line 127, in compose_mapping_node
    while not self.check_event(MappingEndEvent):
  File "/lila/home/chodera/miniconda/envs/perses-dev/lib/python3.9/site-packages/yaml/parser.py", line 98, in check_event
    self.current_event = self.state()
  File "/lila/home/chodera/miniconda/envs/perses-dev/lib/python3.9/site-packages/yaml/parser.py", line 428, in parse_block_mapping_key
    if self.check_token(KeyToken):
  File "/lila/home/chodera/miniconda/envs/perses-dev/lib/python3.9/site-packages/yaml/scanner.py", line 115, in check_token
    while self.need_more_tokens():
  File "/lila/home/chodera/miniconda/envs/perses-dev/lib/python3.9/site-packages/yaml/scanner.py", line 152, in need_more_tokens
    self.stale_possible_simple_keys()
  File "/lila/home/chodera/miniconda/envs/perses-dev/lib/python3.9/site-packages/yaml/scanner.py", line 291, in stale_possible_simple_keys
    raise ScannerError("while scanning a simple key", key.mark,
yaml.scanner.ScannerError: while scanning a simple key
  in "/data/chodera/chodera/perses/perses/examples/new-cli/parsed-2022-05-08-17:24:50-template.yaml", line 51, column 1
could not find expected ':'
  in "/data/chodera/chodera/perses/perses/examples/new-cli/parsed-2022-05-08-17:24:50-template.yaml", line 52, column 1
```
Here's an example YAML file that is corrupted---note the last line:
```yaml
$ cat /data/chodera/chodera/perses/perses/examples/new-cli/parsed-2022-05-08-17:24:50-template.yaml
anneal_1,4s: false
atom_expr: 134217728
atom_expression:
- IntType
atom_selection: not water
bond_expession:
- DefaultBonds
bond_expr: null
checkpoint_interval: 50
complex_box_dimensions: null
fe_type: repex
forcefield_files:
- amber/ff14SB.xml
- amber/tip3p_standard.xml
- amber/tip3p_HFE_multivalent.xml
- amber/phosaa10.xml
given_geometries_tolerance: 0.4
ionic_strength: 0.15
ligand_file: ligands.sdf
map_strength: null
n_cycles: 10000
n_equilibration_iterations: 0
n_states: 12
n_steps_per_move_application: 250
neglect_angles: false
new_ligand_index: 10
nonbonded_method: PME
offline-freq: 10
old_ligand_index: 8
phases:
- solvent
- complex
pressure: 1
protein_pdb: ripk2_receptor.pdb
protocol-type: default
remove_constraints: false
render_atom_map: true
rmsd_restraint: false
run_type: null
small_molecule_forcefield: openff-2.0.0
small_molecule_parameters_cache: null
softcore_v2: false
solvent_box_dimensions: null
solvent_padding: 9.0
spectators: null
temperature: 300
timestep: 4
trajectory_directory: 10ns_lig8to10
trajectory_prefix: out
use_given_geometries: true
e
```
This has now happened twice: Once to @mikemhenry, once to me. 

Any ideas what may be happening?

jchodera:
Here's the issue:

The only way the CLI makes the generated YAML filenames unique is by use of the datestamp with second granularity. 

As a result:
(1) A race condition exists where a file will become corrupted when being written by two different instances running at the same time
(2) When launching many different simulations at the same time, we end up duplicating simulation work or losing track of simulations we are running because the YAML file is overwritten or shared by multiple simulations that are launched.

We absolutely have to fix this before deploying it.

jchodera:
As a terrible hack, I've added a UUID to make the filenames unique, but I think we should actually standardize the filenames and store them in the perses experiment directory (#998).

ijpulidos:
We could just use `setup_options['trajectory_directory']` [here](https://github.com/choderalab/perses/blob/aca1ea2af4acb3744527f577cca6c27dd402b106/perses/app/setup_relative_calculation.py#L750) when we generate the yaml file. I think that should be clean enough. We do have to consider the changes that we might end up doing in https://github.com/choderalab/perses/issues/997

PR #976, opened 2022-03-31
grab forces by name instead of hard-coded index

mikemhenry:
## Description

While working on #949 I found that we were grabbing forces by index, and it seems some changes to openmm changed the order of these forces, causing errors. See https://github.com/choderalab/perses/pull/953/files#issuecomment-1083962468

## Motivation and context

Resolves #???

## Change log

```

```


perses/tests/test_relative.py
```diff
@@ -823,16 +823,18 @@ def flattenedHybridTopologyFactory_energies(topology, chain, system, positions,
         # Make list of off atoms that should have flattened torsions/exceptions
         off_atoms = topology_proposal.unique_new_atoms if endstate == 0 else topology_proposal.unique_old_atoms
         system = topology_proposal.old_system if endstate == 0 else topology_proposal.new_system
+        # TODO will openmm ever allow duplicate names here?
+        force_names = {force.__class__.__name__ : index for index, force in enumerate(system.getForces())}
```
jchodera, line 827: FYI: There's now a `force.getName()` method that can be used instead, though this is a recent addition.
The name can even be set by users if nonstandard names are desired.

jchodera (approved):
Looks good, thanks!

zhang-ivy:
@mikemhenry : Can you clarify what changes in openmm caused the change in order of the forces?

ijpulidos (commented):
This looks good, I'm just worried that there's no guarantee that the forces names will be different. According to official doc they would default to using the class name if the name is not specified. When was this changed upstream? I tried looking but couldn't find a direct cause for it.

mikemhenry:
> @mikemhenry : Can you clarify what changes in openmm caused the change in order of the forces?

No idea what changed, but it seems clear the order can change either from where we set them or grab them so getting by name seems safer



> This looks good, I'm just worried that there's no guarantee that the forces names will be different. According to official doc they would default to using the class name if the name is not specified. When was this changed upstream? I tried looking but couldn't find a direct cause for it.

I did print all the names and there are no duplicates, given this is a test, it seems it would be up to whoever is re-writing the test and adding duplicate forces names to not do that.

mikemhenry:
RE: Our dev meeting, I'm going to see what prints out on CI to see what the order is for OpenMM 7.7 and Nightly

mikemhenry:
Nightly the order is:
```
{'HarmonicBondForce': 0, 'PeriodicTorsionForce': 1, 'NonbondedForce': 2, 'HarmonicAngleForce': 3}
```

On openMM 7.7.0

```
{'HarmonicBondForce': 0, 'HarmonicAngleForce': 1, 'PeriodicTorsionForce': 2, 'NonbondedForce': 3}
```

I have no idea why it changes.

zhang-ivy:
@mikemhenry : Hmm very strange. Is it worth opening an issue in openmm about this to see what may have changed in the nightly? 
Either way, I think you should be good to merge this PR, since the test is written to check the vanilla hybrid system, which as its written now, only has one `PeriodicTorsionForce` and one `NonbondedForce`, so I think this PR is fine for the purposes of fixing this test.

mikemhenry:
Looks like I had to fix another hard-coded force index. Not sure if I should fix all of them or just the ones that are giving us problems.

zhang-ivy:
@mikemhenry : Maybe would be best to fix all of them?

jchodera:
Let's fix anything that used a hard-coded force index.

zhang-ivy:
The nightlys are still failing here. 
```
________________ test_PointMutationExecutor_endstate_validation ________________
[gw1] linux -- Python 3.9.12 /usr/share/miniconda/envs/test/bin/python

    def test_PointMutationExecutor_endstate_validation():
        from pkg_resources import resource_filename
        from simtk import unit
    
        from perses.app.relative_point_mutation_setup import PointMutationExecutor
    
        pdb_filename = resource_filename("perses", "data/ala_vacuum.pdb")
>       PointMutationExecutor(
            pdb_filename,
            "1",
            "2",
            "ASP",
            ionic_strength=0.15 * unit.molar,
            flatten_torsions=False,
            flatten_exceptions=False,
            generate_unmodified_hybrid_topology_factory=True,
            generate_repartitioned_hybrid_topology_factory=True,
            generate_rest_capable_hybrid_topology_factory=True
        )
```

mikemhenry:
@zhang-ivy That is a different issue https://github.com/choderalab/perses/issues/993 (which I would love ideas on how to fix/what you think might be causing it, there is a TODO that looks like it could be the solution, but I wonder if there is some regression causing us to get different energies now)
So I think this one is good to go? @ijpulidos @zhang-ivy

ijpulidos:
@mikemhenry I agree, it is a different issue. This one should be good to go.

zhang-ivy:
Ah thanks! I didn’t notice that issue before

On May 12, 2022, at 15:51, Mike Henry ***@***.***> wrote:

﻿

Merged #976<https://github.com/choderalab/perses/pull/976> into main.

PR #977, opened 2022-04-05
Avoid changing global context cache behavior on modules imports

### LINKED ISSUE #968: Clean up global context cache use
I recently was using this code snippet to run one of my energy validation tests:

```python
    from perses.tests.utils import validate_endstate_energies_point
   
    outdir = "/data/chodera/zhangi/perses_benchmark/repex/33/0/"

    for endstate in [0, 1]:
        with open(os.path.join(outdir, f"0_apo.pickle"), "rb") as f:
            htf = pickle.load(f)
            validate_endstate_energies_point(htf, endstate=endstate, minimize=True)
```

However, I'm getting this error:
```
Input In [2], in run_rest_factory_test(name)
     45 with open(os.path.join(outdir, f"0_{phase}.pickle"), "rb") as f:
     46     htf = pickle.load(f)
---> 48 validate_endstate_energies_point(htf, endstate=endstate, minimize=True)

File ~/miniconda3/envs/openmm-dev/lib/python3.9/site-packages/perses-0.9.5+94.g0fb13173.dirty-py3.9.egg/perses/tests/utils.py:950, in validate_endstate_energies_point(htf, endstate, minimize)
    948     hybrid_positions = sampler_state.positions
    949 context_hybrid.setPositions(hybrid_positions)
--> 950 components_hybrid = compute_potential_components(context_hybrid, beta=beta)
    952 # Get energy components of original system
    953 thermostate_other = states.ThermodynamicState(system=system, temperature=temperature)

File ~/miniconda3/envs/openmm-dev/lib/python3.9/site-packages/perses-0.9.5+94.g0fb13173.dirty-py3.9.egg/perses/tests/utils.py:307, in compute_potential_components(context, beta, platform)
    304 # Make a deep copy of the system.
    305 import copy
--> 307 from perses.dispersed.utils import configure_platform
    308 platform = configure_platform(platform.getName(), fallback_platform_name='Reference', precision='double')
    310 system = context.getSystem()

File ~/miniconda3/envs/openmm-dev/lib/python3.9/site-packages/perses-0.9.5+94.g0fb13173.dirty-py3.9.egg/perses/dispersed/utils.py:90, in <module>
     87     return platform
     89 #########
---> 90 cache.global_context_cache.platform = configure_platform(utils.get_fastest_platform().getName())
     91 #########
     92 
     93 #smc functions
     94 def compute_survival_rate(sMC_particle_ancestries):

File ~/miniconda3/envs/openmm-dev/lib/python3.9/site-packages/openmmtools-0.21.1+14.g025fc04-py3.9.egg/openmmtools/cache.py:335, in ContextCache.platform(self, new_platform)
    332 @platform.setter
    333 def platform(self, new_platform):
    334     if len(self._lru) > 0:
--> 335         raise RuntimeError('Cannot change platform of a non-empty ContextCache')
    336     if new_platform is None:
    337         self._platform_properties = None

RuntimeError: Cannot change platform of a non-empty ContextCache
```

This is because [this line](https://github.com/choderalab/perses/blob/main/perses/dispersed/utils.py#L90) tries to set the platform of the global context cache, but was already set when [this line](https://github.com/choderalab/perses/blob/rest-over-protocol/perses/tests/utils.py#L867) was called, because `feptasks.py` has `import openmmtools.cache as cache`, which creates a global context cache [here](https://github.com/choderalab/openmmtools/blob/main/openmmtools/cache.py#L791).

If I comment out [this line](https://github.com/choderalab/perses/blob/main/perses/dispersed/utils.py#L90), the problem goes away. 

I think we need to go through the all the files in this repo and make sure that global context caches are being used properly.

ijpulidos:
## Description

We are now avoiding changing the global context cache behavior when importing/using modules.

## Motivation and context

Resolves #968

## How has this been tested?

Tested manually with snippet found in #968 

## Change log

```
Global context cache behavior is now unchanged when importing modules.
```

### DIFF
--- a/perses/dispersed/utils.py
+++ b/perses/dispersed/utils.py
@@ -2,7 +2,6 @@ import simtk.openmm as openmm
 import os
 import copy
 
-from openmmtools import cache
 import openmmtools.mcmc as mcmc
 import openmmtools.integrators as integrators
 import openmmtools.states as states
@@ -20,7 +19,6 @@ from perses.annihilation.lambda_protocol import LambdaProtocol
 import dask.distributed as distributed
 from scipy.special import logsumexp
 import openmmtools.cache as cache
-from openmmtools import utils
 
 # Instantiate logger
 logging.basicConfig(level = logging.NOTSET)
--- a/perses/dispersed/utils.py
+++ b/perses/dispersed/utils.py
@@ -86,11 +86,8 @@ def configure_platform(platform_name='Reference', fallback_platform_name='CPU', 
     print(f"conducting subsequent work with the following platform: {platform.getName()}")
     return platform
 
-#########
-cache.global_context_cache.platform = configure_platform(utils.get_fastest_platform().getName())
-#########
 
-#smc functions
+# smc functions
 def compute_survival_rate(sMC_particle_ancestries):
     """
     compute the time-series survival rate as a function of resamples
--- a/perses/dispersed/utils.py
+++ b/perses/dispersed/utils.py
@@ -768,7 +768,7 @@ class LocallyOptimalAnnealing():
                 if rethermalize:
                     self.context.setVelocitiesToTemperature(self.thermodynamic_state.temperature) #rethermalize
                 if noneq_trajectory_filename is not None:
-                    self.save_configuration(idx, sampler_state, context)
+                    self.save_configuration(idx, sampler_state)
                 if return_timer:
                     timer[idx] = time.time() - start_timer
             except Exception as e:
@@ -884,7 +884,7 @@ class LocallyOptimalAnnealing():
         self.thermodynamic_state.apply_to_context(self.context)
 
 
-    def save_configuration(self, iteration, sampler_state, context):
+    def save_configuration(self, iteration, sampler_state):
         """
         pass a conditional save function
 
--- a/perses/dispersed/feptasks.py
+++ b/perses/dispersed/feptasks.py
@@ -24,7 +24,6 @@ logging.basicConfig(level = logging.NOTSET)
 _logger = logging.getLogger("feptasks")
 _logger.setLevel(logging.INFO)
 
-#cache.global_context_cache.platform = openmm.Platform.getPlatformByName('Reference') #this is just a local version
 EquilibriumFEPTask = namedtuple('EquilibriumInput', ['sampler_state', 'inputs', 'outputs'])
 NonequilibriumFEPTask = namedtuple('NonequilibriumFEPTask', ['particle', 'inputs'])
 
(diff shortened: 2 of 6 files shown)


jchodera:
There are a few other commented-out appearances in the codebase we can remove as well:
* [`feptasks.py:27`](https://github.com/choderalab/perses/blob/c81fa5d6a9bfbea6727d596640949c4cd221160c/perses/dispersed/feptasks.py#L27)
* [`tests/test_relative.py:28-31`](https://github.com/choderalab/perses/blob/3bd663c2514e8878d416b77f065db29836499642/perses/tests/test_relative.py#L28-L31), since we don't need to use the `Reference` platform here
* [`dispersed/smc.py:25`](https://github.com/choderalab/perses/blob/c81fa5d6a9bfbea6727d596640949c4cd221160c/perses/dispersed/smc.py#L25)
* [`samplers/samplers.py:24`](https://github.com/choderalab/perses/blob/c81fa5d6a9bfbea6727d596640949c4cd221160c/perses/samplers/samplers.py#L24)

ijpulidos:
@jchodera Oh yeah, good catch, somehow I missed the last two when looking for matches. Thanks!

perses/dispersed/utils.py
```diff
@@ -20,7 +19,6 @@
 import dask.distributed as distributed
 from scipy.special import logsumexp
 import openmmtools.cache as cache
```
mikemhenry, line 21: This import is safe now with the changes in openmmtools, right? or will importing this create a global context cache?

mikemhenry:
Is this okay? (it won't let me make a comment on it)
https://github.com/choderalab/perses/pull/977/files#diff-88ae267cea8ae4381fe65491208bff038f987c5ac99031e3ce9d44c375abac9eL109

mikemhenry:
This too:
https://github.com/choderalab/perses/pull/977/files#diff-09a4eea0080619a2bd6101be97e93a97d8f0c248d25cefc79e84b7df2a55506dR140

mikemhenry (approved):
There were a few more places that showed up when I ran 
`rg cache` that look sus but they might be okay, I've pointed them out :point_up_2: but LGTM

perses/dispersed/utils.py
jchodera: Importing will create a global context cache, but it should be a singleton, I believe.

jchodera (approved):
Looks much better, thanks!

jchodera:
@mikemhenry : I can't tell which lines you're pointing to.

mikemhenry:
> @mikemhenry : I can't tell which lines you're pointing to.

Sorry I didn't realize those links didn't keep what I had highlighted -- but they were the same as the cache import which you pointed out here https://github.com/choderalab/perses/pull/977#discussion_r844686290

Issue #978, opened 2022-04-07
Adapt `create_endstates()` function for new REST HTF and real time output analysis


ijpulidos:
We need to adapt the `create_endstates()` function for the new rest factory in #848

ijpulidos:
We need not only to fix the function for the new REST htf but also for it to be able to work with the real time analysis output, more information in PR #979. Changing title accordingly.

PR #979, opened 2022-04-08
Fix real time mbar disabling appending uncorrelated endstates

### LINKED ISSUE #916: real-time analysis output
To have a way to know how is our simulation performing in "real time", perses can generate a computer- and human-readable file (YAML?) with at least the following information:

* **progress**: current iteration, max iterations, percent progress
* **free energy estimates**: DDG, dDDG, number of uncorrelated samples
* **performance**: ns/day, wall clock time per iteration, ETA for completion

We could use/revive what's in the [`MultiStateSampler` in openmmtools](https://github.com/choderalab/openmmtools/blob/48a3f35f06442ca252a5589dc3ebd52e681ce30b/openmmtools/multistate/multistatesampler.py#L1393-L1472).

ijpulidos:
## Description

In order to have the real time analysis YAML output with estimates for MBAR calculations and performance of the simulation, we had to disable uncorrelated end states creation.

Minor clean up: No minimization needed for `HybridCompatibilityMixin`.

## Motivation and context

Resolves #916 

## How has this been tested?

tested locally. Waiting for benchmarks runs for real-case performance testing.

## Change log

```
Output for real time mbar calculations and simulation performance in YAML format.
```

### DIFF
--- a/perses/app/setup_relative_calculation.py
+++ b/perses/app/setup_relative_calculation.py
@@ -660,7 +660,8 @@ def run_setup(setup_options, serialize_systems=True, build_samplers=True):
                 if phase == 'vacuum':
                     endstates = False
                 else:
-                    endstates = True
+                    # TODO: Make this True when dispersed.utils.create_endstates is fixed
+                    endstates = False
 
                 if setup_options['fe_type'] == 'fah':
                     _logger.info('SETUP FOR FAH DONE')
--- a/perses/samplers/multistate.py
+++ b/perses/samplers/multistate.py
@@ -27,12 +27,10 @@ class HybridCompatibilityMixin(object):
         self._hybrid_factory = hybrid_factory
         super(HybridCompatibilityMixin, self).__init__(*args, **kwargs)
 
-    def setup(self, n_states, temperature, storage_file, minimisation_steps=100,
+    # TODO: Should this overload the create() method from parent instead of breing setup()?
+    def setup(self, n_states, temperature, storage_file,
               n_replicas=None, lambda_schedule=None,
               lambda_protocol=LambdaProtocol(), endstates=True):
-
-
-        from perses.dispersed import feptasks
 
         hybrid_system = self._factory.hybrid_system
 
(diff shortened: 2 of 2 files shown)


ijpulidos:
Minimization is needed so far, but we will have to look into this in more detail in the future.

perses/app/setup_relative_calculation.py
```diff
@@ -660,7 +660,8 @@ def run_setup(setup_options, serialize_systems=True, build_samplers=True):
                 if phase == 'vacuum':
                     endstates = False
                 else:
-                    endstates = True
+                    # TODO: Make this True when dispersed.utils.create_endstates is fixed
+                    endstates = False
```
jchodera, line 664: Do we want to output something to the logger so the user knows this too? We can say "Disabling creation of endstates with expanded cutoffs in order to support new alchemical factories."

perses/samplers/multistate.py
```diff
@@ -29,11 +27,11 @@ def __init__(self, *args, hybrid_factory=None, **kwargs):
         self._hybrid_factory = hybrid_factory
         super(HybridCompatibilityMixin, self).__init__(*args, **kwargs)
 
-    def setup(self, n_states, temperature, storage_file, minimisation_steps=100,
```
jchodera, line 32: Did you say you had to _keep_ `minimisation_steps`, or is this just unused?

jchodera (approved):
Can you also link in the issue you created where we need to restore this once we've figured out how to generalize it?

Otherwise, looks good to me for now!

ijpulidos:
CC https://github.com/choderalab/perses/issues/978 for the creation of endstates issue. We have to consider re-enabling this once we refactor the `create_endstates` function.

perses/samplers/multistate.py
ijpulidos: You are right, this was not being used, but I think it doesn't hurt to have it back,  because it could actually break existing code if we don't.

ijpulidos: Yes, you are right. It wasn't being used but we should add it back since it _could_ break currently existing code.

perses/app/setup_relative_calculation.py
ijpulidos: Yes, I agree.

jchodera:
Looks good, thanks!

PR #981, opened 2022-04-11
Benchmark DeltaG plots now shift data to experimental DeltaG mean.

jchodera:
## Description

Currently, the DeltaG calculated vs experiment comparison plots subtract the mean from both axes, leading to nonsensical experimental axes:
espaloma-plot_absolute-mean
This PR adds back the experimental mean free energy to both axes so that the experimental range is correct:
image

cc: https://github.com/openforcefield/openff-arsenic/issues/42

## Motivation and context

Resolves what was originally suspected to be an issue with `openff-arsenic` but was just an incorrect usage:
https://github.com/openforcefield/openff-arsenic/issues/42
 
## How has this been tested?

The attached plot has been generated.

## Change log

```
Benchmarking now produces a calculated versus experimental DeltaG plot that correctly centers both axes on the mean experimental binding free energy.
```

### DIFF
--- a/benchmarks/benchmark_analysis.py
+++ b/benchmarks/benchmark_analysis.py
@@ -211,12 +211,12 @@ plotting.plot_DDGs(fe.graph,
                    figsize=5,
                    filename='./plot_relative.png'
                    )
-# Absolute plot
+# Absolute plot, with experimental data shifted to correct mean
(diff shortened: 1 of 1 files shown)


ijpulidos (approved):
Thanks for catching this. Looks great!

Issue #975, opened 2022-03-31
GPU Tests Failing


mikemhenry:
I need to investigate this further but I thought @zhang-ivy might have an idea on what is going on here:
https://github.com/choderalab/perses/runs/5732242075?check_suite_focus=true

```
        if chain_id not in chain_ids_in_topology:
>           raise Exception(f"Specified chain id {chain_id} not found in the wildtype topology. choices are {chain_ids_in_topology}")
E           Exception: Specified chain id 1 not found in the wildtype topology. choices are ['D', 'D', '3', '4']
```

These tests only run on the GPU, so I'm not surprised that things are working on our GHA CI. We probably should figure out a way to run a short version of these tests on GHA so we can catch them in the PR that causes the regression before we merge them in. 

It looks like https://github.com/choderalab/perses/pull/967 introduced this bug.

zhang-ivy:
@mikemhenry : 
- First, I'd like to point out that these lines can be removed: https://github.com/choderalab/perses/blob/main/perses/tests/test_GPU.py#L48-L51
If the test is only running neq switching on apo barstar, we don't need to include the barnase PDB file as `ligand_input`.
However, I'm not sure that this will fix the error. 
- I can't tell whether the error is coming from generating the htf for the apo phase or the complex phase, but I think the error might be coming from this change: https://github.com/choderalab/perses/pull/967/files#diff-b8f1b8770beeae3c016d936d6e8276134c5c92836b2fea2adb739e0a14747234L186
I made this change because I was getting an error when inputing a solvated PDB and creating the openmm topology from the mdtraj topology, so I thought it would be better to use the openmm topology directly. However, this change may cause differences in the chain IDs, so one thing to try is to change that line to: `protein_topology = protein_md_topology.to_openmm() if solvated else protein_topology`

ijpulidos:
Yes, I think @zhang-ivy is right about the problem being caused by the change in chanIDs when going back and for from/to mdtraj. I'll take a look into this, this might also hint that we need a test for the `PointMutationExecutor` that deals with different chains ids/numbers.

zhang-ivy:
@ijpulidos : Have you had a chance to look into this?

ijpulidos:
> @ijpulidos : Have you had a chance to look into this?

Mot really. Not much other than what I already said up there. As far as I can see, using `mdtraj` ignores/overrides the chain IDs, whereas if we use the `openmm` topology it should use whatever the PDB file specifies as the chain ID.

[...]

PR #985, opened 2022-04-22
remove dead lines in test + keep topology in openmm if we can

mikemhenry:
## Description

Fix issue with GPU tests where we didn't handle solvated topologies correctly.

## Motivation and context

See https://github.com/choderalab/perses/issues/975

Resolves 975

## How has this been tested?

@ijpulidos can you test this locally on a GPU? Right now my laptop graphics drivers are borked 

## Change log

```

```


zhang-ivy:
Looks good. In the future, when designing the API, we will probably want to make sure both cases use the mdtraj topology OR both use the openmm topology (like mentioned in case 2 [here](https://github.com/choderalab/perses/issues/975#issuecomment-1101705948))

zhang-ivy (approved):

PR #986, opened 2022-04-22
update to run gpu tests on openMM 7.7

mikemhenry:
## Description

We should run our GPU tests using the newest openMM version we support.

## Motivation and context

We were running our GPU tests on openMM 7.6 but we should use 7.7 as it is the newest version we support. 

Resolves #???

## Change log

```

```


.github/workflows/self-hosted-gpu-test.yml
```diff
@@ -66,7 +66,7 @@ jobs:
       - name: Refine test env
         shell: bash -l {0}
         run: |
-          mamba install -y cudatoolkit==11.0.3 openmm==7.6
+          mamba install -y cudatoolkit==11.0.3 openmm==7.7
```
ijpulidos, line 69: I wonder if not pinning any version should be enough to always get the latest supported version.

ijpulidos (approved):
Looks good to me!

.github/workflows/self-hosted-gpu-test.yml
mikemhenry: In theory it should, I like to use this sometimes as a canary though, like if something happens to the conda-forge recipe that makes 7.7 not able to install, and so we grab 7.6, I'd rather get an error saying the env couldn't solve then just test on 7.6. Thoughts? If we got some aws credits, I'd do a more proper matrix which would grab versions we want + nightly.

ijpulidos: Yeah, if we have the credits I guess a matrix would be better. I was also thinking that we probably want to match what we have in our CI test environment [here](https://github.com/choderalab/perses/blob/main/devtools/conda-envs/test_env.yaml#L20), I think that would actually make more sense to correlate GHA CI tests with the AWS GPU tests, if needed.

mikemhenry: I think you're right! I'll specify the version like that, good idea!

jchodera:
@mikemhenry : We're back to a ton of these errors---do you need to fix the buffer issue in more tests?
```
E                   new_positions, logp_proposal = geometry_engine.propose(topology_proposal, pos, beta,
E                 File "/usr/share/miniconda/envs/test/lib/python3.8/site-packages/perses/rjmc/geometry.py", line 237, in propose
E                   logp_proposal, new_positions, rjmc_info, atoms_with_positions_reduced_potential, final_context_reduced_potential, neglected_angle_terms, omitted_terms = self._logp_propose(top_proposal, current_positions, beta, direction='forward', validate_energy_bookkeeping = validate_energy_bookkeeping)
E                 File "/usr/share/miniconda/envs/test/lib/python3.8/site-packages/perses/rjmc/geometry.py", line 641, in _logp_propose
E                   final_context = openmm.Context(final_system, final_system_integrator, platform)
E                 File "/usr/share/miniconda/envs/test/lib/python3.8/site-packages/openmm/openmm.py", line 16230, in __init__
E                   _openmm.Context_swiginit(self, _openmm.new_Context(*args))
E               openmm.OpenMMException: NonbondedForce: The cutoff distance cannot be greater than half the periodic box size.

perses/tests/test_examples.py:34: Exception
----------------------------- Captured stdout call -----------------------------
```

mikemhenry:
Let me merge in `main` to this PR and see, I think those errors are fixed now

PR #988, opened 2022-04-27
Specifying hybrid topology factory (htf) from input file

### LINKED ISSUE #899: API point to specify the Hybrid Topology Factory
We need a way to specify the Factory we want for Hybrid Topology. From API and YAML/CLI.

ijpulidos:
## Description

Created a common API point to handle the Hybrid Topology Factory creation. It can handle the 'vanilla' `HybridTopologyFactory` and the new `RESTCapableHybridTopologyFactory`. 

The new parameter for the input `yaml` file is `hybrid_topology_factory` and it expects the full name of the class as a value. It raises an error if you provide an invalid name.

## Motivation and context

Resolves #899

## How has this been tested?

Tested locally by running the protein-ligand example from cli.

## Change log

```

```

### DIFF
--- a/perses/app/setup_relative_calculation.py
+++ b/perses/app/setup_relative_calculation.py
@@ -7,7 +7,7 @@ import simtk.unit as unit
 import logging
 from pathlib import Path
 
-from perses.annihilation.relative import HybridTopologyFactory
+from perses.annihilation.relative import HybridTopologyFactory, RESTCapableHybridTopologyFactory
 from perses.app.relative_setup import RelativeFEPSetup
 from perses.annihilation.lambda_protocol import LambdaProtocol
 
@@ -306,6 +306,12 @@ def getSetupOptions(filename):
     if not 'rmsd_restraint' in setup_options:
         setup_options['rmsd_restraint'] = False
 
+    # Handling htf input parameter
+    if 'hybrid_topology_factory' not in setup_options:
+        default_htf_class_name = "HybridTopologyFactory"
+        setup_options['hybrid_topology_factory'] = default_htf_class_name
+        _logger.info(f"\t 'hybrid_topology_factory' not specified: default to {default_htf_class_name}")
+
     os.makedirs(trajectory_directory, exist_ok=True)
 
 
@@ -562,33 +568,24 @@ def run_setup(setup_options, serialize_systems=True, build_samplers=True):
         else:
             _internal_parallelism = None
 
-
         ne_fep = dict()
         for phase in phases:
             _logger.info(f"\t\tphase: {phase}")
-            hybrid_factory = HybridTopologyFactory(top_prop['%s_topology_proposal' % phase],
-                                               top_prop['%s_old_positions' % phase],
-                                               top_prop['%s_new_positions' % phase],
-                                               neglected_new_angle_terms = top_prop[f"{phase}_forward_neglected_angles"],
-                                               neglected_old_angle_terms = top_prop[f"{phase}_reverse_neglected_angles"],
-                                               softcore_LJ_v2 = setup_options['softcore_v2'],
-                                               interpolate_old_and_new_14s = setup_options['anneal_1,4s'],
-                                               rmsd_restraint=setup_options['rmsd_restraint'],
-                                               )
+            hybrid_factory = _generate_htf(phase, top_prop, setup_options)
 
             if build_samplers:
-                ne_fep[phase] = SequentialMonteCarlo(factory = hybrid_factory,
-                                                     lambda_protocol = setup_options['lambda_protocol'],
-                                                     temperature = temperature,
-                                                     trajectory_directory = trajectory_directory,
-                                                     trajectory_prefix = f"{trajectory_prefix}_{phase}",
-                                                     atom_selection = atom_selection,
-                                                     timestep = timestep,
-                                                     eq_splitting_string = eq_splitting,
-                                                     neq_splitting_string = neq_splitting,
-                                                     collision_rate = setup_options['ncmc_collision_rate_ps'],
-                                                     ncmc_save_interval = ncmc_save_interval,
-                                                     internal_parallelism = _internal_parallelism)
+                ne_fep[phase] = SequentialMonteCarlo(factory=hybrid_factory,
+                                                     lambda_protocol=setup_options['lambda_protocol'],
+                                                     temperature=temperature,
+                                                     trajectory_directory=trajectory_directory,
+                                                     trajectory_prefix=f"{trajectory_prefix}_{phase}",
+                                                     atom_selection=atom_selection,
+                                                     timestep=timestep,
+                                                     eq_splitting_string=eq_splitting,
+                                                     neq_splitting_string=neq_splitting,
+                                                     collision_rate=setup_options['ncmc_collision_rate_ps'],
+                                                     ncmc_save_interval=ncmc_save_interval,
+                                                     internal_parallelism=_internal_parallelism)
 
         print("Nonequilibrium switching driver class constructed")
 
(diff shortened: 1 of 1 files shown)


ijpulidos:
@zhang-ivy Can you help me figure out why I am getting the following error? This happens when I try to run the `protein-ligand` benchmark code specifying the `RESTCapableHybridTopologyFactory`.

```python
  File "/home/user/miniconda3/envs/perses-dev/bin/perses-relative", line 33, in <module>
    sys.exit(load_entry_point('perses', 'console_scripts', 'perses-relative')())
  File "/home/user/workdir/repos/perses/perses/app/setup_relative_calculation.py", line 800, in run
    setup_dict = run_setup(setup_options)
  File "/home/user/workdir/repos/perses/perses/app/setup_relative_calculation.py", line 614, in run_setup
    zero_state_error, one_state_error = validate_endstate_energies(_top_prop, _htf, _forward_added_valence_energy, _reverse_subtracted_valence_energy, beta = 1.0/(kB*temperature), ENERGY_THRESHOLD = ENERGY_THRESHOLD)#, trajectory_directory=f'{xml_directory}{phase}')
  File "/home/user/workdir/repos/perses/perses/tests/utils.py", line 796, in validate_endstate_energies
    nonalch_zero, nonalch_one, alch_zero, alch_one = generate_endpoint_thermodynamic_states(hybrid_system, top_proposal, repartitioned_endstate)
  File "/home/user/workdir/repos/perses/perses/tests/utils.py", line 387, in generate_endpoint_thermodynamic_states
    check_system(system)
  File "/home/user/workdir/repos/perses/perses/tests/utils.py", line 346, in check_system
    force = forces['PeriodicTorsionForce']
KeyError: 'PeriodicTorsionForce'
```

As the error says it is expecting that key [here](https://github.com/choderalab/perses/blob/732dd87e4dfe7e5b5a95b7e304357037798c99b6/perses/tests/utils.py#L346), but the `forces` dictionary doesn't have it.

perses/app/setup_relative_calculation.py
```diff
@@ -1024,5 +1013,30 @@ def _resume_run(setup_options):
...
+    try:
+        htf = factory(topology_proposal_dictionary[f'{phase}_topology_proposal'],
+                      topology_proposal_dictionary[f'{phase}_old_positions'],
+                      topology_proposal_dictionary[f'{phase}_new_positions'],
+                      neglected_new_angle_terms=topology_proposal_dictionary[f"{phase}_forward_neglected_angles"],
+                      neglected_old_angle_terms=topology_proposal_dictionary[f"{phase}_reverse_neglected_angles"],
+                      softcore_LJ_v2=setup_options['softcore_v2'],
+                      interpolate_old_and_new_14s=setup_options['anneal_1,4s'],
+                      rmsd_restraint=setup_options['rmsd_restraint']
+                      )
```
zhang-ivy, line 1034: @ijpulidos : I think you are missing a bunch of `HybridTopologyFactory` and `RESTCapableHybridTopologyFactory` arguments: https://github.com/choderalab/perses/blob/main/perses/app/relative_point_mutation_setup.py#L444-L460.

zhang-ivy: Most of the ones in the code I linked are for the `HybridTopologyFactory`. The only ones that are used for the `RESTCapableHybridTopologyFactory` are: `rest_radius=rest_radius, w_scale=w_scale` -- you'll want to make sure the yaml has options for specifying these.

ijpulidos: Yes, you are correct.

ijpulidos:
@zhang-ivy You are correct, but I'm using what used to be in the `setup_relative_calculation.py` script originally [here](https://github.com/choderalab/perses/blob/c9fa0c6702da536e32983dccf188d2171bee073f/perses/app/setup_relative_calculation.py#L569-L577) and [here](https://github.com/choderalab/perses/blob/c9fa0c6702da536e32983dccf188d2171bee073f/perses/app/setup_relative_calculation.py#L607-L615). I agree, they are not specifying all of the parameters, using many of the defaults. But I don't want to change the logic here, unless this has been missing since the beginning and it's something we want to change.

ijpulidos:
I had to manually add the new option/key to the `fah_generator.py` in order for tests to pass. This is less than ideal but I think we want this if we want to move forward with this PR without having to do a big refactor that will eventually be overridden when we change things to the new `Settings` object from the OpenFE folks.

ijpulidos:
> I had to manually add the new option/key to the `fah_generator.py` in order for tests to pass. This is less than ideal but I think we want this if we want to move forward with this PR without having to do a big refactor that will eventually be overridden when we change things to the new `Settings` object from the OpenFE folks.

Related https://github.com/choderalab/perses/issues/995

perses/app/setup_relative_calculation.py
```diff
@@ -1071,5 +1060,76 @@ def _process_overrides(overrides, yaml_options):
...
+        reverse_substracted_valence_energy = topology_proposal_dict[f"{phase}_subtracted_valence_energy"]
+        zero_state_error, one_state_error = validate_endstate_energies(topology_proposal,
+                                                                       current_htf,
+                                                                       forward_added_valence_energy,
+                                                                       reverse_substracted_valence_energy,
+                                                                       **kwargs)
+        _logger.info(f"\t\terror in zero state: {zero_state_error}")
+        _logger.info(f"\t\terror in one state: {one_state_error}")
+    elif isinstance(current_htf, RESTCapableHybridTopologyFactory):
+        for endstate in [0, 1]:
```
zhang-ivy, line 1130: When validating the `RESTCapableHybridTopologyFactory`, we need to run `validate_endstate_energies_point` twice -- once at each endstate. We don't want to run the test twice on the same htf though, since the test involves changing hybrid system parameters. I recommend making a copy of the htf within this for loop, like I do [here](https://github.com/choderalab/perses/blob/main/perses/app/relative_point_mutation_setup.py#L440)

perses/app/setup_relative_calculation.py
```diff
@@ -1071,5 +1060,76 @@ def _process_overrides(overrides, yaml_options):
...
+            _logger.info("'rest_radius' not specified. Using default value.")
+        try:
+            rest_specific_options.update({'w_scale': setup_options['w_scale']})
+        except KeyError:
+            _logger.info("'w_scale' not specified. Using default value.")
+
+        # update htf_setup_dictionary with new parameters
+        htf_setup_dict.update(rest_specific_options)
+    else:
+        raise ValueError(f"Unsupported Hybrid Topology Factory. Check 'hybrid_topology_factory' name in input file.")
```
zhang-ivy, line 1094: I think a more useful error message here would be something like f"You specified an unsupported factory type: {factory_name}. Currently, the supported factories are: HybridTopologyFactory and RESTCapableHybridTopologyFactory."

perses/app/setup_relative_calculation.py
ijpulidos: As discussed, it's probably better to make the copy inside the `validate_endstate_energies_point`. This will help with not having to remember the function alters the htf and also should be better in terms of memory management. We would need to change this as well for `validate_endstate_energies_md` and make the corresponding changes in `relative_point_mutation_setup.py` and `test_relative.py`. I'll do those, it should be an easy fix.

zhang-ivy:
One more thing that is no longer necessary due to your suggestion to make a copy of `htf` in the `validate_endstate_energies_point/md()` functions is saving the htf as a pickle [here](https://github.com/choderalab/perses/blob/main/perses/tests/test_relative.py#L1009-L1023)

All the pickle code can be removed and we can just call:
```python
for endstate in [0, 1]:
    validate_endstate_energies_point(htf, endstate=endstate, minimize=True)
```

The pickle code is only in there because I wanted to run the validate function on fresh htfs each time (aka the same reason why I was making copies of the htfs).

zhang-ivy (approved):

PR #991, opened 2022-05-03
Fix typo in test_resume.py

### LINKED ISSUE #990: Vacuum phase typo in test_resume
This [line](https://github.com/choderalab/perses/blob/main/perses/tests/test_resume.py#L250) has a typo in 'vacuum', which means that the simulation is actually being run in solvent. Should be an easy fix.

zhang-ivy:
## Description

Fixes typo that causes the test to be run in solvent instead of vacuum.

## Motivation and context

Resolves #990 

## Change log

```

```

### DIFF
--- a/perses/app/relative_point_mutation_setup.py
+++ b/perses/app/relative_point_mutation_setup.py
@@ -362,7 +362,10 @@ class PointMutationExecutor(object):
                                                         validate_energy_bookkeeping=validate_bool)
 
             # Check for charge change...
-            self._handle_charge_changes(topology_proposal, new_positions)
+            if phase != 'vacuum':
+                self._handle_charge_changes(topology_proposal, new_positions)
+            else:
+                _logger.info("Skipping counterion because phase is vacuum.")
 
             if generate_unmodified_hybrid_topology_factory:
                 repartitioned_endstate = None
--- a/perses/app/relative_point_mutation_setup.py
+++ b/perses/app/relative_point_mutation_setup.py
@@ -23,6 +23,7 @@ temperature = 300 * unit.kelvin
 kT = kB * temperature
 beta = 1.0/kT
 ring_amino_acids = ['TYR', 'PHE', 'TRP', 'PRO', 'HIS', 'HID', 'HIE', 'HIP']
+KNOWN_PHASES = ['vacuum', 'complex', 'solvent']
 
 # Set up logger
 import logging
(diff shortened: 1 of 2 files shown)


ijpulidos:
I see that the tests are failing because it detects a charge difference and tries to use the `get_water_indices` from the `PointMutationEngine` [here](https://github.com/choderalab/perses/blob/main/perses/app/relative_point_mutation_setup.py#L368). I guess we shouldn't be changing from `ALA` to `ASP` if we don't want to deal with charges differences. Maybe changing to other similar residues with hydrophobic side chains should be enough (i.e. `VAL`), especially if all we are testing here is that the simulation can be resumed. Thoughts?

zhang-ivy:
I think we want to make sure that the counterion isn't introduced for charge changing transformations in vacuum. I'll fix this.

mikemhenry:
> I think we want to make sure that the counterion isn't introduced for charge changing transformations in vacuum. I'll fix this.

Yes that is a good point and something I didn't consider in the OG PR.

mikemhenry:
@zhang-ivy I hate to open this can of worms on this PR, but do we need to make the same change for small molecules as well? I know @dominicrufa recently refactored this a bit and I'm not sure if he thought of that either.

zhang-ivy:
@mikemhenry : Nope, I double checked and the counterion behavior is correct for small molecules!

jchodera (approved):
Thanks!

perses/app/relative_point_mutation_setup.py
```diff
@@ -362,7 +362,10 @@ def __init__(self,
                                                         validate_energy_bookkeeping=validate_bool)
 
             # Check for charge change...
-            self._handle_charge_changes(topology_proposal, new_positions)
+            if phase != 'vacuum':
```
jchodera, line 365: Can you add a sanity check to this method that checks to make sure that `phase` is valid?
```python
KNOWN_PHASES = ['vacuum', 'complex', 'solvent']
if not phase in KNOWN_PHASES:
    raise ValueError(f"phase '{phase}' unknown; must be one of {KNOWN_PHASES}")
```

zhang-ivy: Done!

zhang-ivy:
Ah, I need to update the PointMutationExecutor to not introduce a counterion if phase=vacuum was specified. We should keep the test as is.

On May 3, 2022, at 19:49, Iván Pulido ***@***.***> wrote:

﻿

I see that the tests are failing because it detects a charge difference and tries to use the get_water_indices from the PointMutationEngine here<https://github.com/choderalab/perses/blob/main/perses/app/relative_point_mutation_setup.py#L368>. I guess we shouldn't be changing from ALA to ASP if we don't want to deal with charges differences. Maybe changing to other similar residues with hydrophobic side chains should be enough (i.e. VAL), especially if all we are testing here is that the simulation can be resumed. Thoughts?

PR #992, opened 2022-05-03
Update HybridCompatibilityMixin to handle RESTCapableHybridTopologyFactory and add repex self-consistency tests

### LINKED ISSUE #984: Add `HybridCompatilityMixin` subclass for use with `RESTCapableHybridTopologyFactory`
We'll want to subclass  `HybridCompatilityMixin` (used with `HybridTopologyFactory`) to work for running repex with `RESTCapableHybridTopologyFactory`. 

Also need to add a test for running dipeptide repex, which will be specified as a GPU unit test.

zhang-ivy:
## Description

- Updated `HybridCompatilityMixin` to automatically detect which type of factory (`HybridTopologyFactory` or `RESTCapableHybridTopologyFactory`) was inputted and create a sampler depending on the factory type. 
- Added self-consistency tests for dipeptide systems (running repex with the new `RESTCapableHybridTopologyFactory`):
   - Neutral mutation: ALA->THR vs THR->ALA
   - Charge changing: ARG->ALA->ARG vs LYS->ALA->LYS (see docstring of test for explanation on why these particular transformations are necessary)

TODO:
- I put the tests in `test_samplers.py` -- not sure if this is the right place -- may want to move it somewhere else and/or make it an example?
- Not sure if the tests in `test_samplers.py` are even working -- should we delete everything else in here?
- `n_iterations` should be bumped up to 1000 and it should be run on a gpu (@mikemhenry : Can you help with this?)

## Motivation and context

Resolves #984 

## Change log

```

```


mikemhenry:
Could we run a few iterations on the CPU but skip the validation checks, and run the 1000 iterations on the GPU? An advantage of just doing a few iterations on the CPU is it will help find API/other run time errors faster

zhang-ivy:
Yes that sounds good. Sorry I wasn’t clear, that’s actually what I meant.

On May 4, 2022, at 03:51, Mike Henry ***@***.***> wrote:

﻿

Could we run a few iterations on the CPU but skip the validation checks, and run the 1000 iterations on the GPU? An advantage of just doing a few iterations on the CPU is it will help find API/other run time errors faster

perses/tests/test_samplers.py
```diff
@@ -154,10 +154,187 @@ def run_samplers(testsystem_names, niterations=5):
...
+    from openmm import unit
+
+    from perses.app.relative_point_mutation_setup import PointMutationExecutor
+    from perses.dispersed.utils import configure_platform
+    from perses.samplers.multistate import HybridRepexSampler
+
+    from openmmtools.multistate import MultiStateReporter, MultiStateSamplerAnalyzer
+    from openmmtools import cache, utils, mcmc
+    platform = configure_platform(utils.get_fastest_platform().getName())
+    cache.global_context_cache.platform = platform
```
ijpulidos, line 176: We shouldn't try to change the `global_context_cache` platform this way, since already by default it should be using the fastest available platform. Also, as far as I see we are specifying local contexts in lines 221 and 222, so the `global_context_cache` is not used.  I think the tests are failing because the `global_context_cache` is imported and being used in other parts of the code already. Removing line 176 should fix this.
```suggestion
    platform = configure_platform(utils.get_fastest_platform().getName())
```

perses/tests/test_samplers.py
```diff
@@ -154,10 +154,187 @@ def run_samplers(testsystem_names, niterations=5):
...
+    from openmm import unit, app
+
+    from perses.app.relative_point_mutation_setup import PointMutationExecutor
+    from perses.dispersed.utils import configure_platform
+    from perses.samplers.multistate import HybridRepexSampler
+
+    from openmmtools.multistate import MultiStateReporter, MultiStateSamplerAnalyzer
+    from openmmtools import cache, utils, mcmc
+    platform = configure_platform(utils.get_fastest_platform().getName())
+    cache.global_context_cache.platform = platform
```
ijpulidos, line 264: Same as above.
```suggestion
    platform = configure_platform(utils.get_fastest_platform().getName())
```

perses/tests/test_samplers.py
```diff
@@ -154,10 +154,187 @@ def run_samplers(testsystem_names, niterations=5):
...
-if __name__=="__main__":
-    for t in test_hybrid_scheme():
-        t()
+# if __name__=="__main__":
+    # for t in test_hybrid_scheme():
+    #     t()
 #    for t in test_samplers():
 #        print(t.description)
 #        t()
+#     run_RESTCapableHybridTopologyFactory_repex()
```
ijpulidos, line 340: We should probably remove these lines and also the `test_hybrid_scheme()` commented function, if we are not using them. In any case the comments are already in the history in case we ever need them again, I guess.
```suggestion
```

ijpulidos (changes requested):
Looks good! Just a few comments.

On the other hand, in my opinion these tests should be examples, and we are already testing the examples in `test_examples.py`, there's probably a bit of tweaking to be done in order to get this run in the new GPU CI, since the examples predates it. These tests are not really testing "units" but rather complete workflows/simulations. I can make the changes for this to be converted as examples, we also need examples with repex.

jchodera:
Can this make it into the 0.10.0 release, or will it have to wait until 0.11.0?

zhang-ivy:
@jchodera : This can probably make it into `0.10.0`, we just need to decide whether we want to keep the tests I wrote as tests or convert them to examples. Was thinking we could discuss this tomorrow at the dev sync. 

After we make a decision, we might need @mikemhenry to help specify which tests should be run on cpu vs gpu.

zhang-ivy:
I should also note that the new tests I added in this PR are the reason why the GHA workflow has failures. This is because the tests are only running 2 cycles of repex, which is surely not enough to get the free energies to be equal and opposite. We'll need to move this test to the gpu to get the free energies to match.

zhang-ivy:
Notes from dev sync:
- We need to determine the smallest number of replicas and iterations necessary to get the repex examples to converge.
- Make this an example (not a test) -- which will be marked as gpu only and will run this once a week on aws (or lilac??) to catch potential regressions.

ijpulidos:
I keep getting the following error traceback when running the counterion mutation test using 4 or more iterations (2 work):
```python
  File "/lila/data/chodera/pulidoi/sandbox/perses-992/test_REST_counterion_mutation.py", line 79, in <module>
    f_ij, df_ij = analyzer.get_free_energy()
  File "/home/pulidoi/miniconda3/envs/perses-dev/lib/python3.9/site-packages/openmmtools/multistate/multistateanalyzer.py", line 1932, in get_free_energy
    self._compute_free_energy()
  File "/home/pulidoi/miniconda3/envs/perses-dev/lib/python3.9/site-packages/openmmtools/multistate/multistateanalyzer.py", line 1892, in _compute_free_energy
    (Deltaf_ij, dDeltaf_ij) = self.mbar.getFreeEnergyDifferences()
  File "/home/pulidoi/miniconda3/envs/perses-dev/lib/python3.9/site-packages/pymbar/mbar.py", line 541, in getFreeEnergyDifferences
    Theta_ij = self._computeAsymptoticCovarianceMatrix(
  File "/home/pulidoi/miniconda3/envs/perses-dev/lib/python3.9/site-packages/pymbar/mbar.py", line 1679, in _computeAsymptoticCovarianceMatrix
    check_w_normalized(W, N_k)
  File "/home/pulidoi/miniconda3/envs/perses-dev/lib/python3.9/site-packages/pymbar/utils.py", line 358, in check_w_normalized
    raise ParameterError(
pymbar.utils.ParameterError: Warning: Should have \sum_n W_nk = 1.  Actual column sum for state 0 was 0.635215. 12 other columns have similar problems
```

zhang-ivy:
Ah yes that’s an issue with your version of pymbar. Make sure you install from master.

On May 24, 2022, at 20:04, Iván Pulido ***@***.***> wrote:

﻿

I keep getting the following error traceback when running the counterion mutation test using 4 or more iterations (2 work):

  File "/lila/data/chodera/pulidoi/sandbox/perses-992/test_REST_counterion_mutation.py", line 79, in <module>
    f_ij, df_ij = analyzer.get_free_energy()
  File "/home/pulidoi/miniconda3/envs/perses-dev/lib/python3.9/site-packages/openmmtools/multistate/multistateanalyzer.py", line 1932, in get_free_energy
    self._compute_free_energy()
  File "/home/pulidoi/miniconda3/envs/perses-dev/lib/python3.9/site-packages/openmmtools/multistate/multistateanalyzer.py", line 1892, in _compute_free_energy
    (Deltaf_ij, dDeltaf_ij) = self.mbar.getFreeEnergyDifferences()
  File "/home/pulidoi/miniconda3/envs/perses-dev/lib/python3.9/site-packages/pymbar/mbar.py", line 541, in getFreeEnergyDifferences
    Theta_ij = self._computeAsymptoticCovarianceMatrix(
  File "/home/pulidoi/miniconda3/envs/perses-dev/lib/python3.9/site-packages/pymbar/mbar.py", line 1679, in _computeAsymptoticCovarianceMatrix
    check_w_normalized(W, N_k)
  File "/home/pulidoi/miniconda3/envs/perses-dev/lib/python3.9/site-packages/pymbar/utils.py", line 358, in check_w_normalized
    raise ParameterError(
pymbar.utils.ParameterError: Warning: Should have \sum_n W_nk = 1.  Actual column sum for state 0 was 0.635215. 12 other columns have similar problems

zhang-ivy:
@ijpulidos : This is the pymbar issue related to your error: https://github.com/choderalab/pymbar/issues/419

zhang-ivy:
@ijpulidos : Actually, which version are you using? 3.0.6+ should be fine, 3.0.5 is broken.

ijpulidos:
@zhang-ivy yes, I was using 3.0.5, I can confirm 3.0.6+ works fine. Thanks!

I checked the behavior of the free energy with the number of iterations. I don't see any tendency or convergence. The results are as follows
```
# neutral mutation
n_iterations, energy_diff
2,      4.18895
4,      0.990343
8,      -0.00456955
16,     -5.50846
32,     -0.54498
64,      -1.93432
128,     0.827766
256,     0.176885
512,     0.0994963  # took ~3 hours

# counterion mutation
n_iterations, energy_diff
2,      4.79188
4,      6.45721
8,      0.626254
16,     1.82293
32,     -1.89355
64,     1.7545
128,    -2.88888   # took ~1.5 hours
```
None of them converged (they all resulted in an `AssertionError` as per the test). The walltime for the longest ones are specified.

zhang-ivy:
@ijpulidos : I think the `n_iterations` is too small,  can you try n_iterations=250, 500, 1000, 1500, 2000? You basically already did the first 2 for the neutral mutation, so need to repeat those.

ijpulidos:
@zhang-ivy Sure I can try, I'm just wondering if we are okay with tests running for more than 6 hours in the GPU CI. I guess it's not much of an issue if these are running overnight or so.

mikemhenry:
Yah lets first figure out what the minimal amount of time we can spend is, then we can figure out what we want to do with that. It may be something we check before a release or once a month or something. A p2.xl is something like 90 cents an hour, so even it it takes 10 hours, that isn't munch per month or week if we decide we need it.

ijpulidos:
The most recent results are as follows (I'm included previous results as well just in case we spot some pattern or something).
```
# neutral mutation
n_iterations, energy_diff
2,      4.18895
4,      0.990343
8,      -0.00456955
16,     -5.50846
32,     -0.54498
64,      -1.93432
128,     0.827766
256,     0.176885
512,     0.0994963  # took ~3 hours
1024,   -0.199971
1512,   -0.026376  # took ~13 hours in A40

# counterion mutation
n_iterations, energy_diff
2,      4.79188
4,      6.45721
8,      0.626254
16,     1.82293
32,     -1.89355
64,     1.7545
128,    -2.88888   # took ~1.5 hours
512,    1.87339
1024,   -1.04253
1512,   -1.37579
2048,   0.446562
4096,   -0.247799
5000,   -0.436782  # took ~17 hours in A100
```
These are still not converged in terms of `np.isclose` which, as far as I can see, defaults to checking around 5 significant figures. I wonder if we are expecting too much and whether we should change our convergence criteria. As far as I can see, it also depends on the system, the counterion one is one order of magnitude further apart compared to the neutral system. Thoughts?

ijpulidos:
> As far as I can see, it also depends on the system, the counterion one is one order of magnitude further apart compared to the neutral system. Thoughts?

Actually, nvm this part, if we check the significant figures in the actual values (not the difference) it is similar for both systems. For example:
```
# neutral system
ALA-THR is -51.84165478283243. THR-ALA is -51.81527874133261
# counterion system
ARG-ALA-ARG is -277.7522884618811. LYS-ALA-LYS is -277.31550615053516
```

zhang-ivy:
>I wonder if we are expecting too much and whether we should change our convergence criteria. 

Yes, I think the convergence criteria is too stringent. Let's just check whether the difference in the forward and reverse free energies is < 1 kT, or perhaps 0.5 kT if we want to be a bit more stringent.

ijpulidos:
Ok, so I think we leave these as tests (now I think they are actually tests not examples, sorry for my confusion before). I'll be pushing the changes soon.

@mikemhenry I'm wondering if we use a new `pytest` mark for these kind of tests that could potentially take hours/days maybe having yet another CI workflow that run these specially marked tests every once in a while or so? Ideas on how could we accomplish this are appreciated.

mikemhenry:
@ijpulidos How about we merge this in and see how long it takes on AWS? How long does it take if we only care about <1kT?

ijpulidos:
> @ijpulidos How about we merge this in and see how long it takes on AWS? How long does it take if we only care about <1kT?

So there are two tests, the first one might run in a reasonable time, the second one I have the suspicion that it is going to take more than 24 hours to run on the current AWS instance we are using. I guess we can just try and see, I'm going to make the changes for this to be run (it's now explicitly skipped with pytest). What's a good limit for this? My assumption is that anything below 24 hours should be okay since we are running the GPU CI every 24 hours as well, but I don't know if there are other limits (github/aws limiting it).

mikemhenry:
On a self hosted runner I think it's like a 30 day timeout 🤣 let's do it an see what happens

zhang-ivy:
@ijpulidos : I just reviewed your changes and I don't think we should call the file `test_topology_factories.py` -- we aren't actually testing the factory here -- we do that in `test_relative.py` -- I think we should leave it as `test_samplers.py`, since we are checking to make sure the sampler `HybridCompatabilityMixin` is working properly.

ijpulidos:
@zhang-ivy Yes, I agree. I think it makes sense to have a `test_repex.py` module or similar with all the repex tests (these ones at least, maybe we'll migrate/create others with time). Similarly with others like `sams` and `neq-switching`, etc. I'll make the changes.

mikemhenry (approved):
LGTM

ijpulidos (approved):

PR #1007, opened 2022-05-17
Specifying energy components by name, using dictionaries.

### LINKED ISSUE #993: `validate_endstate_energies_point` failing on openmm nightly
`test_PointMutationExecutor_endstate_validation` and a few other tests are failing now on the [nightly tests for openmm](https://github.com/choderalab/perses/runs/6282350509?check_suite_focus=true):

```
             print(f"{components_other[i][0]} -- og: {components_other[i][1]}, hybrid: {components_hybrid[i][1]}")
>           assert np.isclose(components_other[i][1], components_hybrid[i][1])
E           AssertionError

perses/tests/utils.py:964: AssertionError
```

(a few tests are also failing with a force index error, but that is fixed by PR #976 )

ijpulidos:
## Description

The function `compute_potential_components` is now using dictionaries to look for components by name, avoiding changes in the order by nightly and upcoming `openmm` versions.

## Motivation and context

Resolves #993 

## How has this been tested?

Tested locally.

## Change log

```
Energy components are now accessed by name, using dictionaries in `compute_potential_components`.
```


ijpulidos:
I noticed that if we don't force the same `platform` in the tests we get accuracy problems. That's why I had to change some of the tests to use the specified platform. We need a better and streamlined way of accomplishing this. I already raised an issue #1008

perses/tests/utils.py
```diff
@@ -959,14 +958,22 @@ def validate_endstate_energies_point(htf, endstate=0, minimize=False):
...
     # Check that each of the valence force energies are concordant
     # TODO: Instead of checking with np.isclose(), check whether the ratio of differences is less than a specified energy threshold (like in validate_endstate_energies())
-    for i in range(3):
-        print(f"{components_other[i][0]} -- og: {components_other[i][1]}, hybrid: {components_hybrid[i][1]}")
-        assert np.isclose(components_other[i][1], components_hybrid[i][1])
-
-    # Check that the nonbonded force energies are concordant
+    # Build map between other and rest force keys - keys are for other, values are for rest
+    bonded_keys_other_to_hybrid = {'HarmonicBondForce': 'CustomBondForce', 'HarmonicAngleForce': 'CustomAngleForce',
+                                   'PeriodicTorsionForce': 'CustomTorsionForce'}
```
ijpulidos, line 963: While accessing the force/energy components by name should make it more robust and it doesn't expect any specific order with the `getForces()` method in `openmm`, I have the feeling this is still a bit prone to errors, especially if we start adding multiple `CustomBondForce` objects, which would not be included in the highlighted map. Our current tests/HTFs do not have this, but it is something that may happen in the future.

ijpulidos: On way is to actually store both the name (from `force.getName()`) and the class name (from `force.__class__.__name__`). But then this means that the part where it validates the energy needs to specify the indices if it wants the force name or the object's class name, which can get confusing. There's "no free lunch".

ijpulidos:
Could this be the PR that helps us solve the different problems we were facing with Openmm nightly? @mikemhenry @zhang-ivy When you have the chance, can you take a look into this one? I believe we'd like to have something that solves the nightly tests we keep encountering before trying to keep merging more PRs. Thanks!

mikemhenry:
Awesome! I'll look at tomorrow but it looks like this will get nightly passing again!

zhang-ivy:
Just double checked that the hybrid systems all have unique force names and wanted to note that here. This is necessary because in this PR, @ijpulidos uses `force.getName()` in `compute_potential_components`.

For the `HybridTopologyFactory`:
['MonteCarloBarostat', 'CustomBondForce', 'HarmonicBondForce', 'CustomAngleForce', 'HarmonicAngleForce', 'CustomTorsionForce', 'PeriodicTorsionForce', 'NonbondedForce', 'CustomNonbondedForce', 'CustomBondForce']

For the `RepartitionedHybridTopologyFactory`:
['MonteCarloBarostat', 'HarmonicBondForce', 'HarmonicAngleForce', 'PeriodicTorsionForce', 'NonbondedForce']

For the `RESTCapableHybridTopologyFactory':
['MonteCarloBarostat', 'CustomBondForce', 'CustomAngleForce', 'CustomTorsionForce', 'CustomNonbondedForce_electrostatics', 'CustomNonbondedForce_sterics', 'CustomBondForce_exceptions', 'NonbondedForce_reciprocal', 'NonbondedForce_sterics']

perses/tests/test_geometry_engine.py
```diff
@@ -1122,6 +1122,7 @@ def _guessFileFormat(file, filename):
     file.seek(0)
     return 'pdb'
 
+# TODO: This function is not being used by anything. Remove?
```
mikemhenry, line 1125: Remove it!

mikemhenry (approved):
LGTM, I like @zhang-ivy suggestion to add this to the doc string

"Note that this function assumes that the RESTCapableHybridTopologyFactory hybrid system contains the following forces:['MonteCarloBarostat', 'CustomBondForce', 'CustomAngleForce', 'CustomTorsionForce', 'CustomNonbondedForce_electrostatics', 'CustomNonbondedForce_sterics', 'CustomBondForce_exceptions', 'NonbondedForce_reciprocal', 'NonbondedForce_sterics']. It may fail if there have been changes to the forces or force names, so proceed with caution"

perses/tests/utils.py
ijpulidos: I like that test, but it still will run into problem if we have non-standard names for `Bond` forces, because the only way I could figure out to easily check the bonded energies is using specific/hard-coded names. I agree this might not be a big issue. Will be making the changes and push them to this PR. Thanks!

zhang-ivy: Hmm, how about we just assert that the force names are what we expect them to be before or after creating the dictionary that maps the forces? That way, the test will catch instances where the force names changed and will not even attempt to validate the endstate energies if that is the case.

Something like: 
```python
expected_hybrid_system_force_names = ['CustomBondForce', 'CustomAngleForce', 'CustomTorsionForce', 'CustomNonbondedForce_electrostatics', 'CustomNonbondedForce_sterics', 'CustomBondForce_exceptions', 'NonbondedForce_reciprocal', 'NonbondedForce_sterics']
hybrid_system_force_names = [force.getName() for force in hybrid_system.getForces()]
try: # Remove barostat, if present. If the system is in vacuum phase, it won't have one of these, so we shouldn't require that its present
    hybrid_system_force_names.remove('MonteCarloBarostat')

assert sorted(hybrid_system_force_names) == expected_hybrid_force_names, f"Force names in the RESTCapableHybridTopologyFactory hybrid system do not match the expected force names. Expected: {expected_force_names}, Got: {hybrid_system_force_names}"
```
You could probably do something similar for the `other_system` forces as well.

ijpulidos: Yes, that may work, but I wanted to avoid doing that, because I don't really have access to the system at that point but to the context, so I would first have to get the system and then the forces. Which is okay, just inconvenient, I guess.

ijpulidos: Oh wait, we already have the system and a `force_dict`! I think this might work. Will make the changes and let you know.

ijpulidos: I thought about it and decided to just add the comment, it would add a bit of complexity to do this and I think it is basically analogous to what we already have and are already testing. Maybe we should try to make a test that checks for unique names in the HTFs, but I am gonna leave that for another set of changes, since it seems a bit out of the scope for this one.

perses/tests/test_geometry_engine.py
zhang-ivy: Is there a similar test somewhere else that does the same thing? 

I would be careful about removing stale code like this -- there are a lot of tests like this in perses (e.g. all of the code in `test_samplers.py` where the test is broken and/or not being run anywhere, but we do want to make sure there is test coverage for that particular feature and so at some point, we'll need to fix the tests. Naybe this is a larger topic to discuss at the next perses dev sync, but for now I would probably just leave the test in there.

ijpulidos:
@zhang-ivy Oh, thanks for noticing this, I didn't realize at the time. Now I'm a bit concerned on why the tests are passing, we need to review them and make sure forces with the same name get added. Ill work on this and make another PR. Thanks again!

mikemhenry:
If we can set them to have different names. that would be ideal.

zhang-ivy:
Yep, just opened #1022

PR #1020, opened 2022-05-25
Speed up functions for retrieving old/new positions from hybrid positions

### LINKED ISSUE #1005: Speed up functions for retrieving old/new positions from hybrid positions
I've noticed that the `HybridTopologyFactory` functions [for retrieving old/new positions](https://github.com/choderalab/perses/blob/main/perses/annihilation/relative.py#L2140-L2184) from the hybrid positions are very slow. 

I've found that these are much faster:

```python
def new_positions(htf, hybrid_positions):
    n_atoms_new = htf._topology_proposal.n_atoms_new
    hybrid_indices = [htf._new_to_hybrid_map[idx] for idx in range(n_atoms_new)]
    return hybrid_positions[hybrid_indices, :]

def old_positions(htf, hybrid_positions):
    n_atoms_old = htf._topology_proposal.n_atoms_old
    hybrid_indices = [htf._old_to_hybrid_map[idx] for idx in range(n_atoms_old)]
    return hybrid_positions[hybrid_indices, :]

```

We should 1) double check that the functions I wrote do the same thing as the original functions and 2) integrate my functions with the existing ones.

zhang-ivy:
## Motivation and context

Resolves #1005 

## How has this been tested?

The existing test `test_relative.py::test_position_output()` passes.

I added a timer to this test and here are the times with the old approach vs new approach (this PR):
- with old approach:
	- old_positions(): 0.04454398155212402
	- new_positions(): 0.04473400115966797
- with new approach
	- old_positions(): 0.001668691635131836
	- new_positions(): 0.001196146011352539

## Change log

```

```

### DIFF
--- a/perses/annihilation/relative.py
+++ b/perses/annihilation/relative.py
@@ -2155,9 +2155,8 @@ class HybridTopologyFactory(object):
         # making sure hybrid positions are simtk.unit.Quantity objects
         if not isinstance(hybrid_positions, unit.Quantity):
             hybrid_positions = unit.Quantity(hybrid_positions, unit=unit.nanometer)
-        old_positions = unit.Quantity(np.zeros([n_atoms_old, 3]), unit=unit.nanometer)
-        for idx in range(n_atoms_old):
-            old_positions[idx, :] = hybrid_positions[idx, :]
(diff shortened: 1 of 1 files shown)


mikemhenry (approved):
LGTM

ijpulidos (approved):
Looks good!

Issue #1011, opened 2022-05-18
Test for unique force/components names in Hybrid topology factories


ijpulidos:
From working on #988 it was suggested that we could benefit from having a test that checks that the names in the force (or components) are unique, in the hybrid system for hybrid topology factories. This is needed because we are now accessing energy components by name instead of index, which makes it more robust, but we now require unique names, evidently.

PR #1022, opened 2022-05-27
Rename second CustomBondForce in HybridTopologyFactory system and update docstrings

zhang-ivy:
## Description

- Renames second `CustomBondForce` in `HybridTopologyFactory` system to `CustomBondForce_exceptions`
- Updates docstring in `HybridTopologyFactory` to contain the list of forces present in the hybrid system
- Fixes docstring for RESTTopologyFactory to accurately reflect the forces in the system 

## Motivation and context

This change is necessary because https://github.com/choderalab/perses/pull/1007 changes `compute_potential_components()` to use a dictionary where the keys are force names and values are energies.

## Change log

```

```

### DIFF
--- a/perses/annihilation/relative.py
+++ b/perses/annihilation/relative.py
@@ -1835,7 +1835,7 @@ class HybridTopologyFactory(object):
 
 
         nonbonded_exceptions_force = openmm.CustomBondForce(old_new_nonbonded_exceptions)
-        name = nonbonded_exceptions_force.__class__.__name__ + '_exceptions'        
+        name = f"{nonbonded_exceptions_force.__class__.__name__}_exceptions"        
         nonbonded_exceptions_force.setName(name)
         self._hybrid_system.addForce(nonbonded_exceptions_force)
         _logger.debug(f"\thandle_old_new_exceptions: {nonbonded_exceptions_force} added to hybrid system")
--- a/perses/annihilation/relative.py
+++ b/perses/annihilation/relative.py
@@ -1823,6 +1823,8 @@ class HybridTopologyFactory(object):
 
 
         nonbonded_exceptions_force = openmm.CustomBondForce(old_new_nonbonded_exceptions)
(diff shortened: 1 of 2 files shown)


ijpulidos:
We also need to change the `validate_endstate_energies_point` because I changed it assuming the bonded forces wouldn't have different names and now it is just broken because we now not only have different names but also different number of components, as far as I can see. 

@zhang-ivy I can contribute code fixing this part once the other parts are completed.

ijpulidos:
It also makes me realize that we need more testing for this part, because tests should've failed with the previous PR (#1007 ), so we might one to deal with https://github.com/choderalab/perses/issues/1011 at the same time.

ijpulidos:
> We also need to change the `validate_endstate_energies_point` because I changed it assuming the bonded forces wouldn't have different names and now it is just broken because we now not only have different names but also different number of components, as far as I can see.

@zhang-ivy from our discussion, thanks for pointing out my confusion, we probably don't **need** to change that function, but we might think of better ways of accomplishing what it does without hardcoding forces/components names. Hopefully we can come up with an easy and more robust way, if not, then it's not too bad to leave it as it is.

zhang-ivy:
@mikemhenry : Also, I think the test failure you linked above is old -- the gpu tests are currently passing.

perses/annihilation/relative.py
```diff
@@ -1823,6 +1835,8 @@ def handle_old_new_exceptions(self):
 
 
         nonbonded_exceptions_force = openmm.CustomBondForce(old_new_nonbonded_exceptions)
+        name = nonbonded_exceptions_force.__class__.__name__ + '_exceptions'
```
mikemhenry, line 1838:
```suggestion
        name = f"{nonbonded_exceptions_force.__class__.__name__}_exceptions"        
```

mikemhenry (changes requested):
Just one suggestion!

mikemhenry (approved):
LGTM

ijpulidos (approved):
Looks good!

Issue #810, opened 2021-06-21
Allow users to set temperature


mikemhenry:
We have a few locations where the default temperature is hard coded and cannot be set by a user.  Refactor so that temperature is set in higher level objects and is correctly passed around. 

Example:

```
--- a/perses/app/relative_hydration.py
+++ b/perses/app/relative_hydration.py
@@ -5,11 +5,8 @@ import numpy as np
 from perses.tests.utils import generate_solvated_hybrid_test_topology
 
 kB = unit.BOLTZMANN_CONSTANT_kB * unit.AVOGADRO_CONSTANT_NA
-temperature = 300.0 * unit.kelvin
-kT = kB * temperature
-beta = 1.0/kT
 
-def simulate_hybrid(hybrid_system,functions, lambda_value, positions, nsteps=500, timestep=1.0*unit.femtoseconds, temperature=temperature, collision_rate=5.0/unit.picoseconds):
+def simulate_hybrid(hybrid_system,functions, lambda_value, positions, nsteps=500, timestep=1.0*unit.femtoseconds, temperature=300.0 * unit.kelvin, collision_rate=5.0/unit.picoseconds):
     platform = openmm.Platform.getPlatformByName("CUDA")
     integrator = openmm.LangevinIntegrator(temperature, collision_rate, timestep)
     context = openmm.Context(hybrid_system, integrator, platform)
```

PR #1023, opened 2022-05-31
Introduce new function for creating unsampled endstates

### LINKED ISSUE #978: Adapt `create_endstates()` function for new REST HTF and real time output analysis

zhang-ivy:
## Description

- Introduces a new `create_endstates()` function called `create_endstates_from_real_systems()` that generates unsampled endstates for the `HybridTopologyFactory` and `RESTCapableHybridTopologyFactory` using each's real systems (old and new systems). 
    - Note that `create_endstates()` only worked for `HybridTopologyFactory`
- Adds necessary tests for the above feature
- Updates `validate_endstate_energies_md()` to accept a `step_freq` argument
    - Note that this is necessary because previously, if we specified `n_steps=10`, the function would save snapshots every 1 ps, but we want it to save every snapshot.

~Note that one of the lines in `tests/test_relative.py::run_unsampled_endstate_energies()' can be removed once we merge this PR: https://github.com/choderalab/perses/pull/1022~
^ I've removed the line now that #1022 has been merged

## Motivation and context

Resolves #978 

## How has this been tested?

Tested in `tests/test_relative.py::test_unsampled_endstate_energies{_GPU}`

## Change log

```

```


perses/dispersed/utils.py
```diff
@@ -20,6 +20,10 @@
 from scipy.special import logsumexp
 import openmmtools.cache as cache
 
+temperature = 300.0 * unit.kelvin
+kT = kB * temperature
+beta = 1.0/kT
```
mikemhenry, line 25: Why hard-code the temperature? I think this should either be set by the user, or pulled from the htf with introspection if possible. Else, we will run into the issue where someone wants to run at 298K or 310K.

perses/tests/test_relative.py
```diff
@@ -1047,3 +1047,112 @@ def test_RESTCapableHybridTopologyFactory_energies_GPU():
...
+                                                 'THR',
+                                                 atp.system,
+                                                 atp.positions,
+                                                 system_generator,
+                                                 conduct_htf_prop=True,
+                                                 generate_rest_capable_hybrid_topology_factory=generate_rest_factory,
+                                                 validate_endstate_energy=False)
+            htfs.append(htf)
+
+    elif test_name == 'barstar':
```
mikemhenry, line 1091: What about the case where the test name is neither `barstar` or `ala-dipeptide`? Since right now if the test name is niether, it will just keep going after the if, elif block.

perses/tests/test_relative.py
```diff
@@ -1047,3 +1047,112 @@ def test_RESTCapableHybridTopologyFactory_energies_GPU():
...
+    Uses run_unsampled_endstate_energies() to run energy validation for the unsampled endstates generated for
+    RESTCapableHybridTopologyFactory and HybridTopologyFactory.
+
+    Test system: alanine dipeptide in solvent.
+    """
+
+    # Run point energy validation test
+    run_unsampled_endstate_energies('ala-dipeptide', use_point_energies=True)
+
+@pytest.mark.gpu_needed
```
mikemhenry, line 1143: @ijpulidos Have you benchmarked how long this takes?

mikemhenry (changes requested):

perses/dispersed/utils.py
zhang-ivy: We hard the temperature in many perses files, e.g. `relative_point_mutation_setup.py`, 'test_relative.py`, etc

We already have an issue open regarding going through the repo and fixing this: https://github.com/choderalab/perses/issues/810, which I think is something we should address in the long term.

perses/tests/test_relative.py
zhang-ivy: He hasn't, but when I ran these in a jupyter notebook with an interactive gpu (probably an lt node) on lilac, the `validate_unsampled_endstates_md()` take 7 minutes for barstar.

perses/tests/test_relative.py
zhang-ivy: I addressed this concern in the most recent commit.

perses/dispersed/utils.py
mikemhenry: Temps are okay in things like tests, I had thought we ran into an issue where we had the temp hard coded and it caused some issues with energy validation, but you're right we never did fix it. It just means that we can't run anything at temp other than 300K. So this is fine for now.

mikemhenry (approved):

mikemhenry (approved):
LGTM

ijpulidos (approved):
Looks good! We still need to figure out why it doesn't play well when computing the real time statistics in openmmtools [`MultiStateSampler._offline_analysis`](https://github.com/choderalab/openmmtools/blob/83dd2be11cf30aadf998369301072d02dccdb1cf/openmmtools/multistate/multistatesampler.py#L1495) .

I tried enabling `endstates=True` [here](https://github.com/choderalab/perses/blob/e3f65eafaef294e68ba5fb939da62115f8094194/perses/app/setup_relative_calculation.py#L661), without success. We are missing the exact error traceback  (logging issues), but it should be related to https://github.com/choderalab/perses/pull/979

zhang-ivy:
>I tried enabling endstates=True here, without success

Can you elaborate on what you mean by `without success`? Does the simulation run ok but we can't analyze it using mbar? or does it error when you try to run the simulation?

ijpulidos:
@zhang-ivy The realtime analysis file doesn't get created and this happened before because we were getting an error with matrix dimensions when estimating free energies using `pymbar` in the `_offline_analysis` function. That's why we decided to disabled the `endstates` flag from the script.

ijpulidos:
@zhang-ivy I was just playing with this and finally managed to recover the logging capabilities. So, the realtime analysis file is generated in [`MultiStateSampler._offline_analysis` method, here](https://github.com/choderalab/openmmtools/blob/83dd2be11cf30aadf998369301072d02dccdb1cf/openmmtools/multistate/multistatesampler.py#L1571-L1581). The problem is that we are catching the exception in the same method [here](https://github.com/choderalab/openmmtools/blob/83dd2be11cf30aadf998369301072d02dccdb1cf/openmmtools/multistate/multistatesampler.py#L1532) and the error is not thrown unless it fails enough times (6 times) as per the conditional in line 1553, making it a bit tricky to test.

I managed to get the error and it is a `ParameterError` with the message `initial_f_k must be a 5-dimensional np array`. Which is raised [here on the pymbar side](https://github.com/choderalab/pymbar/blob/7ff37e95c92527c6d9e64453ecf892986edb21c6/pymbar/mbar.py#L276-L278).

zhang-ivy:
>The problem is that we are catching the exception in the same method here and the error is not thrown unless it fails enough times (6 times) as per the conditional in line 1553, making it a bit tricky to test.

To make it easier to debug, I would just remove the 6 times thing for now and just make it throw the error the first time it occurs.

zhang-ivy:
The pymbar error message tells us that the problem originates from [this line](https://github.com/choderalab/openmmtools/blob/83dd2be11cf30aadf998369301072d02dccdb1cf/openmmtools/multistate/multistatesampler.py#L1519), which creates a MultiStateSamplerAnalyzer with `analysis_kwargs={'initial_f_k': self._last_mbar_f_k}` -- I think that `self._last_mbar_f_k` is `None` the first time that `_offline_analysis` is called, so I would expect that the pymbar error message does not occur during the the first iteration. I'm guessing `self._last_mbar_f_k` is computed incorrectly after the first call to `_offline_analysis`, which is what's causing the error.

perses/samplers/multistate.py
```diff
@@ -6,10 +6,9 @@
 
 from openmmtools.multistate import sams, replicaexchange
 from openmmtools.states import CompoundThermodynamicState, SamplerState, ThermodynamicState
+from perses.dispersed.utils import create_endstates, create_endstates_from_real_systems
```
ijpulidos, line 9: Non-blocker. We can remove `create_endstates` here. Just to make it cleaner.
```suggestion
from perses.dispersed.utils import create_endstates_from_real_systems
```

zhang-ivy:
Notes from my chat with @ijpulidos today: 

The `initial_f_k must be a 5-dimensional np array` error comes from [this line](https://github.com/choderalab/openmmtools/blob/83dd2be11cf30aadf998369301072d02dccdb1cf/openmmtools/multistate/multistatesampler.py#L1603).
Questions:
- Is online analysis just a crude estimate of the free energy?
- If it is, should we just not include the unsampled endstates in the loop of this function? 
- If so, we just need a fix for the line I linked above, we don't need to alter the loop at line 1607.

Other problems we found:
- self.online_analysis_interval controls the offline analysis frequency -- we should change this to be called self.offline_analysis_interval
- add docstring that explains difference between online and offline analysis
- locality -- what is this?

These are all openmmtools issues, and they are not blockers for this PR, the perses 0.10.0 release, or the opentools 0.21.4 release.

PR #1027, opened 2022-05-31
CLI can now configure a platform to be used for the simulations

### LINKED ISSUE #939: The `platform_name` should be settable via the YAML and CLI
The `platform_name` should be settable via the YAML and CLI, with `None` as default.

_Originally posted by @jchodera in https://github.com/choderalab/perses/pull/938#discussion_r804081593_

ijpulidos:
## Description

We can now specify the platform to be used for computations through CLI and YAML file in a consistent way. 

## Motivation and context

Up to this point the CLI and YAML input file were not sharing the `platform` option. The CLI only tested a given platform but wasn't "forcing" it in the simulation.

Resolves #939 

## How has this been tested?

Tested locally with CLI examples

## Change log

```
Command line interface (cli) can now set a platform for the multistate sampler computations.
```

### DIFF
--- a/perses/app/cli.py
+++ b/perses/app/cli.py
@@ -1,11 +1,10 @@
 # New cli for testing
 import logging
 import os
-from pathlib import Path
+import openmm.testInstallation
 
 import click
 import openmmtools.utils
-from perses.app.setup_relative_calculation import getSetupOptions, run
 
 # Setting logging level config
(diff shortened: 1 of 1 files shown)


mikemhenry:
does the platform when added to the yaml via the override work?

ijpulidos:
> does the platform when added to the yaml via the override work?

Yes, just tested this locally and it is working correctly.

mikemhenry (approved):

PR #1029, opened 2022-06-01
Changing docstring (avoid double hyphens and asterisk symbols).

### LINKED ISSUE #1028: Fix documentation building issue in RTD
We inadvertently generated an issue with the documentation build by merging PR #1022. I think it might be related to using asterisks and double hyphens [here](https://github.com/choderalab/perses/blob/9c9902e537a9462684f742bb04ea1c928764af66/perses/annihilation/relative.py#L32-L41).

ijpulidos:
## Description

Fixes documentation build issues with read the docs.

## Motivation and context

Resolves #1028 

## How has this been tested?

Testing with documentation build checks in this PR.

## Change log

```

```

### DIFF
--- a/perses/annihilation/relative.py
+++ b/perses/annihilation/relative.py
@@ -29,16 +29,16 @@ class HybridTopologyFactory(object):
         on and are alchemically unmodified.
 
     Here are the forces in the hybrid system:
-    - CustomBondForce -- handles bonds involving `core_atoms` (these are interpolated)
-    - HarmonicBondForce -- handles bonds involving `environment_atoms`, `unique_old_atoms` and `unique_new_atoms` (these are never scaled)
-    - CustomAngleForce -- handles angles involving `core_atoms` (these are interpolated)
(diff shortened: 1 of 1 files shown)


mikemhenry:
Hmmm, I do know that error typically means that it has some issue with that file, and that the error just isn't descriptive

zhang-ivy:
It looks like the problem went away with Mike's most recent commits! Not sure if this is a good permanent fix, but if it is, i think we'll want to roll back Ivan's commits that change the docstring, as they are unnecessary

ijpulidos:
@zhang-ivy Yeah, I guess the best move here is to discard this PR and just leave the issue open (#1028 ). And eventually properly fix it. I'm going to close it, but feel free to re-open it if needed.

ijpulidos:
I changed the doc environment file directly in the `main` branch, since this is a quick fix that doesn't require a PR/review, in my opinion.

PR #1030, opened 2022-06-01
Add option to control whether water is alchemically changed into counterion for charge changes

### LINKED ISSUE #1004: Add option to control whether water is alchemically changed into counterion for charge changes
To help debug issues with charge changes, it would be useful to add an option to control whether we alchemically change a water to a counterion during charge changes.

This is handled [here](https://github.com/choderalab/perses/blob/main/perses/app/relative_setup.py#L421-L422) for the complex phase and [here](https://github.com/choderalab/perses/blob/main/perses/app/relative_setup.py#L475-L476) for the solvent phase.

By default, we should probably have this on, but we should be able to disable this in the YAML file or setup options for debugging.

We could call the option something obvious, like `transmute_waters_into_ions_for_charge_changes` for now.

zhang-ivy:
## Description

This PR allows the user to disable the introduction of counterion(s) for small molecule and protein mutation transformations.

## Motivation and context

Resolves #1004 

## How has this been tested?

@ijpulidos : Could you make sure that specifying `transform_waters_into_ions_for_charge_changes : False` in the yaml actually controls the introduction of the counterion? If the changes I made are right, you should see a logger message that says `Skipping counterion`.

Note that I added a test (for protein mutations, but not small molecules) that checks that this option is working properly. You could adapt this test for a small molecule example, or it might be sufficient to just double check that you see the logger statement when you include the `transform_waters_into_ions_for_charge_changes ` option in the yaml.

## Change log

```
Add option for disabling the introduction of a counterion for charge changing transformations
```


zhang-ivy:
Note that this PR should not be merged until #1024 has been merged, as I made some changes to the protein mutation API. Those changes will need to be carefully merged with these ones.

perses/tests/test_relative_point_mutation_setup.py
```diff
@@ -60,4 +83,35 @@ def test_PointMutationExecutor_solvated():
...
+        conduct_endstate_validation=False,
+        transform_waters_into_ions_for_charge_changes=False
+    )
+    htf = solvent_delivery.get_apo_htf()
+
+    # If there is no counterion, there should be no water atoms in the core atom class
+    solvent_atoms = set(htf.hybrid_topology.select('water'))
+    assert len(solvent_atoms.intersection(htf._atom_classes['core_atoms'])) == 0, "There are water atoms in the core atom " \
+                                                                                  "class, which may mean that a counterion is being introduced"
+
```
ijpulidos, line 117: This test is a good one, but I wonder if we also want a test that directly computes and validates the charge of the mutation with and without the `transform_waters_into_ions_for_charge_changes` flag. Would this be easy to accomplish with our current data structures and methods?

ijpulidos (changes requested):
Looks good, great work. Just a couple of things:
* We need to use the option read from the YAML file.
* Do we want a test that directly computes the charges?

perses/tests/test_relative_point_mutation_setup.py
zhang-ivy: I can't think of an easy way to do this, but maybe you could open an issue for this and we can discuss at the next dev sync?

perses/tests/test_relative_point_mutation_setup.py
```diff
@@ -15,9 +21,21 @@ def test_PointMutationExecutor():
         flatten_exceptions=True,
         conduct_endstate_validation=False,
     )
+    htf = solvent_delivery.get_apo_htf()
+
+    # If there is a counterion, there should be water atoms in the core atom class
+    solvent_atoms = set(htf.hybrid_topology.select('water'))
+    assert len(solvent_atoms.intersection(htf._atom_classes['core_atoms'])) != 0, "There are no water atoms in the core atom " \
+                                                                           "class, which may mean that the counterion was not introduced"
```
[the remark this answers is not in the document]
zhang-ivy: Hmm, but won't that just double the test time for no reason? 

If we have one test that checks to see if `PointMutationExecutor` is initialized without errors and a second test that creates a `PointMutationExecutor` and then checks that the atom classes are correct, that seems like a waste of effort to me.

ijpulidos (approved):
LGTM!

ijpulidos:
By the way, just confirmed locally that the YAML file is working correctly.

ijpulidos:
@zhang-ivy Could you please add what would need to appear in the changelog for these changes? Thanks!

zhang-ivy:
@ijpulidos : Done!

PR #1032, opened 2022-06-06
Recovering logging capabilities

### LINKED ISSUE #1018: Ensure log level is working on new CLI
Make sure that users can set the log level + maybe make the default `INFO`

ijpulidos:
## Description

Adding the import line from `perses.multisate` module recovers the logging. I don't really understand why is this the case, but I could trace it back to this change, even if the imported classes are not directly used in the code.

This isn't really a fix, just a quick solution/workaround I could come up with, to the annoying inconvenience of not having logs, especially annoying for debugging purposes.

## Motivation and context

Resolves #1018 

## How has this been tested?

Tested locally with both `perses-relative` and `perses-cli` entry points.

## Change log

```
Recovering logging capabilities by adding `perses.multistate` import to the root entry point.
```


mikemhenry:
Looks like that that hack didn't re-introduce the performance regression we saw, so it looks like the cache cleanup worked.

mikemhenry (approved):
LGTM! Tested it and it worked!

Issue #1041, opened 2022-06-13
Examine source of large error bars in 0.10.0 release


jchodera:
The [benchmark data for 0.10.0](https://github.com/choderalab/perses/releases/tag/0.10.0) has a few compounds with very large error bars.

![image](https://user-images.githubusercontent.com/3656088/173436774-b950ca79-7dcc-44e3-8e89-8740638b5804.png)

We should take a look at these and see if there is something weird with the atom mappings, or other discernible issue.

zhang-ivy:
I've gone through the release notes and tried to pick out potential reasons why the error has increased for some mutations:
- Bug generated by changes upstream in the openmm package -- default method for calculating solvent padding changed, which resulted in smaller boxes. Fixed by adding more padding to the simulation box, it is now 1.1 nm. Openmm issue https://github.com/openmm/openmm/issues/3502. Perses issue https://github.com/choderalab/perses/issues/949 (https://github.com/choderalab/perses/pull/953)
- Use names (instead of indices) for fetching the force components of a system, avoiding issues with force reordering upstream in openmm. Issue https://github.com/choderalab/perses/issues/993 (https://github.com/choderalab/perses/pull/976 and https://github.com/choderalab/perses/pull/1007)
- Use of unique names for force components in HybridTopologyFactory (https://github.com/choderalab/perses/pull/1022).
- Introduce unsampled_endstates boolean option in input YAML file, which enables/disables creation of unsampled endstates with long-range sterics correction. Issue https://github.com/choderalab/perses/issues/1033 (https://github.com/choderalab/perses/pull/1037).

zhang-ivy:
Ah great find! Could you point me to the code you're using to compute the free energies (and errors) for the above plots? 

And also could you retrieve the `.nc` files for one transformation (both phases, with and without unsampled endstates)? We may need take a closer look at how the free energies (and errors) are computed.

zhang-ivy:
LInk to the code used for computing free energies: https://github.com/choderalab/perses/blob/main/perses/analysis/load_simulations.py

Issue #1013, opened 2022-05-19
General tests cleanup and increase coverage


ijpulidos:
While working on #1007 it was noted that there is code in the tests that never gets executed but they might be a good starting point for new and useful tests. We need to work on a cleanup of the tests and identify the parts of the code base that are not covered by tests.

ijpulidos:
For the `PointMutationExecutor` check https://github.com/choderalab/perses/pull/1030#discussion_r900267666 . This can be accomplish by using pytest fixtures and avoiding redundant code and efforts. In general, we want that in many of our tests.

ijpulidos:
We can parametrize the different combinations in `test_name` and `use_point_energies` for `run_unsampled_endstates_energies` in `test_unsampled_endstate_energies_GPU` test.

PR #1050, opened 2022-06-20
Fix create_endstates_from_real_systems

zhang-ivy:
## Description

This PR fixes `create_endstates_from_real_systems()`, which was not creating the unsampled endstates properly. The global parameters in the valence forces (of the unsampled endstate systems generated by `create_endstates_from_real_systems()`) were not being set according to the appropriate endstate. They were always defaulting to the lambda = 0 endstate, meaning that the unsampled endstate at lambda = 1 had a hybrid system with valence terms set for the lambda = 0 endstate. I've introduced a fix that sets the default values for global parameters in the valence forces according to the right endstate. The fix is based on what is done in `create_endstates()` [here](https://github.com/choderalab/perses/blob/d0bec4d822592e06679036aa6345fd2a3a20102f/perses/dispersed/utils.py#L501-L506). 

This bug was not caught by the existing tests likely because the valence terms are pretty similar for the lambda = 0 and 1 systems in ala dipeptide and barstar. Therefore, I've included one of the tyk2 transformations with large error bars (from which was originally discovered this bug) as a test -- this additional test case should cover transformations where the valence terms are quite different at the lambda = 0 vs 1 endstates. 

## Motivation and context

Resolves https://github.com/choderalab/perses/issues/1041

## How has this been tested?

Added tyk2 test in `test_relative.py` -- the transformation tested here is one that had large error bars as noted in #1041

## Change log

```
Bug fix in the `create_endstates_from_real_systems()` -- fixed by setting the global parameters for valence forces to the appropriate endstate. Also added tyk2 transformation test.
```

### DIFF
--- a/perses/dispersed/utils.py
+++ b/perses/dispersed/utils.py
@@ -556,6 +556,24 @@ def create_endstates_from_real_systems(htf, for_testing=False):
         for index in sorted(indices_to_remove, reverse=True):
             hybrid_system.removeForce(index)
 
+            # Set defaults for global parameters depending on the factory
+            htf_class = htf.__class__.__name__
+            for force_index, force in enumerate(list(hybrid_system.getForces())):
+                if hasattr(force, 'getNumGlobalParameters'): # Only custom forces will have global parameters to set
+                    for parameter_index in range(force.getNumGlobalParameters()):
+                        global_parameter_name = force.getGlobalParameterName(parameter_index)
+                        if global_parameter_name[0:7] == 'lambda_':
+                            if htf_class == 'HybridTopologyFactory':
+                                force.setGlobalParameterDefaultValue(parameter_index, lambda_val)
+                            elif htf_class == 'RESTCapableHybridTopologyFactory':
+                                if 'old' in global_parameter_name:
+                                    force.setGlobalParameterDefaultValue(parameter_index, 1 - lambda_val)
+                                elif 'new' in global_parameter_name:
+                                    force.setGlobalParameterDefaultValue(parameter_index, lambda_val)
+                            else:
+                                raise Exception(
+                                    f"{htf_class} is not supported. Supported factories: HybridTopologyFactory, RESTCapableHybridTopologyFactory")
+
         # Create NonbondedForce
         nonbonded_force = openmm.NonbondedForce()
         hybrid_system.addForce(nonbonded_force)
--- a/perses/tests/test_relative.py
+++ b/perses/tests/test_relative.py
@@ -1110,6 +1110,59 @@ def run_unsampled_endstate_energies(test_name, use_point_energies=True, use_md_e
         vanilla_htf = solvent_delivery.get_apo_htf()
         htfs = [htf, vanilla_htf]
 
+    elif test_name == 'tyk2':
+        import tempfile
+        from perses.utils.url_utils import retrieve_file_url
+        from perses.annihilation.relative import HybridTopologyFactory, RESTCapableHybridTopologyFactory
+        from perses.app.relative_setup import RelativeFEPSetup
+
+        def concatenate_files(input_files, output_file):
+            """
+            Concatenate files given in input_files iterator into output_file.
+            """
+            with open(output_file, 'w') as outfile:
+                for filename in input_files:
+                    with open(filename) as infile:
+                        for line in infile:
+                            outfile.write(line)
+
+        with tempfile.TemporaryDirectory() as temp_dir:
+            # Fetch ligands sdf files and concatenate them in one
+            base_repo_url = "https://github.com/openforcefield/protein-ligand-benchmark"
+            ligand_files = []
+            for ligand in ['lig_ejm_42', 'lig_ejm_54']:
+                ligand_url = f"{base_repo_url}/raw/main/data/2020-02-07_tyk2/02_ligands/{ligand}/crd/{ligand}.sdf"
+                ligand_file = retrieve_file_url(ligand_url)
+                ligand_files.append(ligand_file)
+            concatenate_files(ligand_files, os.path.join(temp_dir, 'ligands.sdf'))
+            ligands_filename = os.path.join(temp_dir, 'ligands.sdf')
+
+            # Retrieve host PDB
+            pdb_url = f"{base_repo_url}/raw/main/data/2020-02-07_tyk2/01_protein/crd/protein.pdb"
+            host_pdb = retrieve_file_url(pdb_url)
+
+            # Generate topology proposal, old/new positions
+            fe_setup = RelativeFEPSetup(
+                ligand_input=ligands_filename,
+                protein_pdb_filename=host_pdb,
+                old_ligand_index=0,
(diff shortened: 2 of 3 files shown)


zhang-ivy:
@ijpulidos : Could you try re-running a couple of the problematic tyk2 benchmark transformations (with large error bars) with this branch? I just want to confirm that the error bars are no longer large before we merge this and get the bug fix release out.

ijpulidos:
@zhang-ivy That bash error is just codecov failing (it does that sometimes). But there's another error when testing energies for REST system here https://github.com/choderalab/perses/runs/6971007912?check_suite_focus=true#step:9:440

zhang-ivy:
@ijpulidos : That error does pop up quite a bit, I have it on my to do list to figure out if there's a way to prevent that test from failing so often. For now, I just re-ran both the failed tests again.

perses/tests/test_relative.py
```diff
@@ -1110,6 +1110,59 @@ def run_unsampled_endstate_energies(test_name, use_point_energies=True, use_md_e
...
+
+        def concatenate_files(input_files, output_file):
+            """
+            Concatenate files given in input_files iterator into output_file.
+            """
+            with open(output_file, 'w') as outfile:
+                for filename in input_files:
+                    with open(filename) as infile:
+                        for line in infile:
+                            outfile.write(line)
```
ijpulidos, line 1127: I see this is the same `concatenate_files` from the `run_benchmarks.py` code. Seems like we want this to be in `perses.utils.concatenate_files` and avoid redundant code between both. We would just import it both here and in the benchmarks script.

ijpulidos: Actually, reading now the whole test, I think what we do really want it to transform the benchmarks code into an actual module in `perses.benchmarks` or similar. This can be left for a future release (next 0.11.0 release sounds like a good option for it). I'll raise an issue with this for the 0.11.0 milestone.

perses/tests/test_relative.py
```diff
@@ -1157,5 +1210,8 @@ def test_unsampled_endstate_energies_GPU():
     # Alanine dipeptide in solvent -- Run MD energy validation test
     run_unsampled_endstate_energies('ala-dipeptide', use_point_energies=False, use_md_energies=True)
 
-    # Apo barstar -- Run point and MD energy validation test
+    # Apo barstar -- Run point and MD energy validation tests
     run_unsampled_endstate_energies('barstar', use_point_energies=True, use_md_energies=True)
+
+    # Tyk2 -- Run point and MD energy validation tests
+    run_unsampled_endstate_energies('tyk2', use_point_energies=True, use_md_energies=True)
```
ijpulidos, line 1217: Just a comment, nothing blocking. Sounds like a good test to parametrize in the future, I'll add that to our [test cleanup issue](https://github.com/choderalab/perses/issues/1013).

perses/dispersed/utils.py
```diff
@@ -556,6 +556,24 @@ def create_endstates_from_real_systems(htf, for_testing=False):
...
+                    for parameter_index in range(force.getNumGlobalParameters()):
+                        global_parameter_name = force.getGlobalParameterName(parameter_index)
+                        if global_parameter_name[0:7] == 'lambda_':
+                            if htf_class == 'HybridTopologyFactory':
+                                force.setGlobalParameterDefaultValue(parameter_index, lambda_val)
+                            elif htf_class == 'RESTCapableHybridTopologyFactory':
+                                if 'old' in global_parameter_name:
+                                    force.setGlobalParameterDefaultValue(parameter_index, 1 - lambda_val)
+                                elif 'new' in global_parameter_name:
+                                    force.setGlobalParameterDefaultValue(parameter_index, lambda_val)
```
ijpulidos, line 572: Just wondering if this would be something that the actual HTF objects should do on their side, rather than having all the logic and conditionals here. Or is this something that's only accessible or only makes sense when creating the endstates? (Non-blocking)

perses/tests/test_relative.py
zhang-ivy: Yes, I agree, raising an issue sounds good -- thanks!

perses/dispersed/utils.py
zhang-ivy: Yes, `setGlobalParameterDefaultValue()` can be called on the HTF side. 

Let me walk through what the code above does and try to write a helper function that would simplify the code. We are creating two hybrid systems: one for the lambda = 0 endstate and one for the lambda = 1 endstate. Therefore, the hybrid systems should have different default values (for lambda = 0, the default values will be 0 and for lambda = 1, the default values will be 1). However, the challenge is that the global parameters for the different factories work differently. 
- For the `HybridTopologyFactory`, we want `lambda_*` to be 0 for lambda = 0 and 1 for lambda = 1. 
- However, for `RESTCapableHybridTopologyFactory`, we decided that:
    - for lambda = 0,  `lambda_*_old` should be 1 and `lambda_*_new` should be 0. 
    - for lambda = 1, `lambda_*_old` should be 0 and `lambda_*_new` should be 1.

Here is are suggested helper functions that we could add to the factories to simplify the logic above.
```python
# Helper function for HybridTopologyFactory
def set_lambdas_for_endstate(self, endstate):
    for force_index, force in enumerate(list(hybrid_system.getForces())):
        if hasattr(force, 'getNumGlobalParameters'):
            for parameter_index in range(force.getNumGlobalParameters()):
                global_parameter_name = force.getGlobalParameterName(parameter_index)
                if global_parameter_name[0:7] == 'lambda_':
                    force.setGlobalParameterDefaultValue(parameter_index, endstate)

# Helper function for RESTCapableHybridTopologyFactory
def set_lambdas_for_endstate(self, endstate):
    for force_index, force in enumerate(list(hybrid_system.getForces())):
        if hasattr(force, 'getNumGlobalParameters'):
            for parameter_index in range(force.getNumGlobalParameters()):
                global_parameter_name = force.getGlobalParameterName(parameter_index)
                if global_parameter_name[0:7] == 'lambda_':
                    if 'old' in global_parameter_name:
                        force.setGlobalParameterDefaultValue(parameter_index, 1 - endstate)
                    elif 'new' in global_parameter_name:
                        force.setGlobalParameterDefaultValue(parameter_index, endstate)
```
which would simplify the above code to:
```python
# Set defaults for global parameters depending on the factory
htf_class = htf.__class__.__name__
htf.set_lambdas_for_endstate(lambda_val)
```

ijpulidos: I have been thinking about this and I think the best approach is probably just to leave it like this and refactor to a more factory-like pattern for the next 0.11.0 release, hopefully without breaking any API compatibility. I can already see many things that can be improved in the whole structure of the classes and methods, but I don't want to make such big changes right now. This is good enough for the current state.

ijpulidos:
I run the benchmarks again with the changes in this branch and we now recovered the low error bars. Funnily enough, that also means we have "recover" that outlier on the right for the relative FE calculations.
 
plot_relative
plot_absolute

PR #1052, opened 2022-06-23
Add timestamp and ligands name to output parsed yaml file

### LINKED ISSUE #998: Have CLI place generated YAML file into the perses calculation directory
Currently, the CLI generates a YAML file with the updated overridden fields, but it places it into a file like
```
parsed-2022-04-18-20:51:55-my.yaml
```
rather than _actually putting it in the perses calculation output directory where it would be useful_.

We should fix this so that we can pull useful information from it, like _what calculation was actually performed to generate all the files in this directory_.

ijpulidos:
## Description

Adds timestamp and ligands names (old and new) metadata to the generated perses yaml file.

Wrote a new function that encapsulates the generation of the parsed YAML file and adds the mentioned metadata.

## Motivation and context

Resolves #998 

## How has this been tested?
Created automated unit test for yaml generation in `test_relative_setup.py`. Also tested locally with running the `examples/protein-ligand/cli` example.

## Change log

```
Perses output yaml file now adds timestamp and ligands names information (for old and new ligands).
```

### DIFF
--- a/perses/app/setup_relative_calculation.py
+++ b/perses/app/setup_relative_calculation.py
@@ -1166,8 +1166,8 @@ def _generate_parsed_yaml(setup_options, input_yaml_file_path):
     yaml_parse_name = f"perses-{time}-{yaml_name}"
     # Add timestamp information
     setup_options["timestamp"] = time
-    # Read input sdf file and save into list
-    ligands_list = Molecule.from_file(setup_options['ligand_file'])
+    # Read input sdf file and save into list -- We don't check stereochemistry
+    ligands_list = Molecule.from_file(setup_options['ligand_file'], allow_undefined_stereo=True)
     # Get names according to indices in parsed setup options
     setup_options['old_ligand_name'] = ligands_list[setup_options['old_ligand_index']].name
     setup_options['new_ligand_name'] = ligands_list[setup_options['new_ligand_index']].name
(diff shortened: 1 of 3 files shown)


perses/app/setup_relative_calculation.py
```diff
@@ -1144,5 +1139,46 @@ def _validate_endstate_energies_for_htf(hybrid_topology_factory_dict: dict, topo
...
+    from openff.toolkit.topology import Molecule
+    # The parsed yaml file will live in the experiment directory to avoid race conditions with other experiments
+    yaml_path = Path(setup_options['trajectory_directory'])
+    yaml_name = Path(input_yaml_file_path).name  # extract name from input/template yaml file.
+    time = datetime.datetime.now().strftime("%Y-%m-%d-%H:%M:%S")
+    yaml_parse_name = f"perses-{time}-{yaml_name}"
+    # Add timestamp information
+    setup_options["timestamp"] = time
+    # Read input sdf file and save into list -- We don't check stereochemistry
+    ligands_list = Molecule.from_file(setup_options['ligand_file'], allow_undefined_stereo=True)
```
ijpulidos, line 1170: I had to explicitly tell `openff-toolkit` to allow undefined stereochemistry for the test in `test_cli_resume_repex` to pass. Apparently we have undefined stereochemistry in the ligands for that test. I think this is okay, but I'm making the comment just in case this is something we have to solve in the test system.

ijpulidos: Not that I'm not touching this test, I just changed the `setup_relative_calculation.py` to extract names from the ligands file using the `openff-toolkit`.

zhang-ivy:
@ijpulidos : LGTM!

zhang-ivy (approved):

PR #1054, opened 2022-06-24
Fix protein mutation repex consistency tests

### LINKED ISSUE #1044: Fix GPU tests for repex internal consistency (protein mutations)
The [repex internal consistency tests](https://github.com/choderalab/perses/blob/main/perses/tests/test_repex.py) for terminally blocked amino acid mutations are taking a long time and failing. We need to figure out the minimum number of replicas/iterations necessary to achieve convergence.

I wrote these tests to check 1) for a neutral mutation (ALA->THR), forward and reverse DGs are equal and opposite, 2) for charge changing mutations, the DGs for ARG->ALA->ARG and LYS->ALA->LYS are equal.

Experiment details:
- using the lifting factory
- n_replicas=36
- phase=solvent
- without unsampled endstates
- with LRC on for real-real interactions

Here are the discrepancies (in kT) we are seeing (@ijpulidos helped me run these):

### 500 iterations
replicate 1: -0.637689
replicate 2: -0.18484
replicate 3: 0.631614
### 1000 iterations
replicate 1: 0.254568
replicate 2: 0.210292
replicate 3: -0.0215541

### 2000 iterations
replicate 1: -0.280622
replicate 2: -0.115679
replicate 3: -0.290202
### 3000 iterations
replicate 1: -0.366365
replicate 2: 0.31904
replicate 3: 0.362364
### 4000 iterations
replicate 1: -0.30283
replicate 2: 0.279554
replicate 3: -0.378

Related to: 
https://github.com/choderalab/perses/pull/1031 and https://github.com/choderalab/perses/pull/992

zhang-ivy:
## Description

This PR introduces the following changes:

For speed up:
- Decrease the number of steps per iteration from 250 to 50 
- Set `reassign_velocities` to be False 

For improved convergence:
- Set `n_iterations` = 3000 for the neutral mutations and = 3000 for the charge changing mutations
- Set `n_replicas=36` for the charge changing test
- Set `minimisation_steps` = 0, which means that the minimization will run until the energy is below a tolerance (instead of running for 100 steps, which is the default in `HybridRepexSampler`)
- Change the convergence criteria to check if DDG < 6 * dDDG

Other:
~- Change the solvent padding to be 1.7 nm (which is necessary for the nightly dev builds of openmm). **Note this may make the tests slower if we are testing against OpenMM <= 7.7**~ EDIT: We are currently testing against openmm 7.7. We should not bump up the solvent padding until we switch the GPU tests to test against openmm => 7.8 
- Clean up unfinished comment and reporter file name

## Motivation and context

This PR changes some of the parameters in the protein mutation repex consistency tests to ensure convergence (DDG < 6 * dDDG) and speed up the tests.

Resolves #1044 

## How has this been tested?

Tested on lilac.

## Change log

```
Fix protein mutation repex internal consistency tests to ensure convergence and speed up the tests. So far only testing neutral transformations.
```


zhang-ivy:
I think we'll want to skip the charge changing test for now -- @ijpulidos do you know how to do this?

perses/tests/test_repex.py
```diff
@@ -81,10 +80,9 @@ def test_RESTCapableHybridTopologyFactory_repex_neutral_mutation():
...
-        forward_data = data['ala-thr']['free_energy']
-        reverse_data = data['thr-ala']['free_energy'] * -1
-        assert np.isclose([forward_data], [reverse_data], atol=ENERGY_DIFF_TOLERANCE_KT), \
-            f"ALA-THR is {forward_data}. THR-ALA is {reverse_data}."
+        DDG = abs(data['ala-thr']['free_energy'] - data['thr-ala']['free_energy'] * -1)
+        dDDG = np.sqrt(data['ala-thr']['error'] ** 2 + data['thr-ala']['error'] ** 2)
+        assert DDG < 6 * dDDG, f"DDG ({DDG}) is greater than 6 * dDDG ({6  * dDDG})"
 
 
 @pytest.mark.gpu_needed
```
ijpulidos, line 88: We can skip this test by making the following change.
```suggestion
@pytest.mark.skip(reason="Currently taking too long in CI.")
@pytest.mark.gpu_needed
```

perses/tests/test_repex.py
```diff
@@ -39,6 +37,7 @@ def test_RESTCapableHybridTopologyFactory_repex_neutral_mutation():
                 "1",
                 "2",
                 mutant_name.upper(),
+                padding=1.7 * unit.nanometers,
```
ijpulidos, line 40: Why is that we need to increase the default padding here? Thinking that would make the test to take more time to run.

ijpulidos (changes requested):
Looks good, just some suggested changes to skip the charge change test and a question with the change of the padding.

perses/tests/test_repex.py
zhang-ivy: I decreased it back down and left a TODO to increase it once we move to testing against openmm >= 7.8

ijpulidos (approved):
Looks good! We have to double check how much the tests will take on our GPU CI and that the charge changes test is not getting run.

ijpulidos:
Just wanted to say, that according to our latest [GPU CI workflow](https://github.com/choderalab/perses/actions/runs/2566051449), we are testing correctly just the neutral mutation. It is passing and taking ~5.5 hours.

jchodera:
This is odd. I was wondering why it took so long, so I ran the test on lilac on a GTX 1080 Ti, using
```
$ pytest -vvv --log-cli-level=DEBUG --log-cli-format="%(asctime)s %(levelname)s %(message)s" perses/tests/test_repex.py::test_RESTCapableHybridTopologyFactory_repex_neutral_mutation
```
to print a timestamp for each step.

It's been stuck at `Generating unsampled endstates` for a full hour!
```
00:11:03 WARNING Warning: The openmmtools.multistate API is experimental and may change in future releases
00:11:03 WARNING Warning: The openmmtools.multistate API is experimental and may change in future releases
00:11:05 INFO n_replicas not defined, setting to match n_states, 12
00:11:17 INFO Generating unsampled endstates.
```
@zhang-ivy : Is it possible this is actually taking an hour? Any idea why?

jchodera:
Hm---I think this is a red herring. If I disable the endstates in the test, it goes silent after the previous line:
```
01:24:42 WARNING Warning: The openmmtools.multistate API is experimental and may change in future releases
01:24:42 WARNING Warning: The openmmtools.multistate API is experimental and may change in future releases
01:24:43 INFO n_replicas not defined, setting to match n_states, 12
```
Some of our code must be disabling or changing the logging so I'm not getting any profess information even when requesting it via `--log-cli-level=DEBUG`. Any ideas why @ijpulidos or @mikemhenry?

jchodera:
For small molecules in solvent, I thought I had been seeing ~30 min
execution times before.

Instead of looking at the overall time (which is not very informative), I
really want to see the breakdown of where time is being spent once we get
past the first iteration.

zhang-ivy:
@jchodera : I posted the logs (for the last 2 iterations) of a charge changing mutation (ALA->ARG) as well as a neutral mutation (THR->ALA) [here](https://github.com/choderalab/perses/issues/1058#issuecomment-1167907464)

>For small molecules in solvent, I thought I had been seeing ~30 min
execution times before.

Hmm, I don't think we've finished writing the tests for small molecule repex consistency yet, so I'm not sure where that would be from, but maybe @dominicrufa  knows which data you're referring to?

jchodera:
I had been referring to timings of small molecules in solvent from my own
previous experiments.

Two things worry me from the logs:
1. The energy matrix computation is way too long at over 3s. This should be
less than 1s. We need to look into why this is happening. Are long range
dispersion corrections slowing things down?
2. The replica propagation time is nearly 7s. This seems way too long for a
small solvent box. Do you have this calculation staged on disk somewhere?

zhang-ivy:
@jchodera : For the neutral (THR->ALA) mutation, which uses n_replicas=12, propagating replicas takes ~2 seconds and computing energy matrix takes < 1 second. However, running 5000 iterations still takes ~3.5 hours, which you mentioned is much longer than expected. Is 2 seconds still much longer than you'd expect for the replica propagation time? If not, then I'm not sure what else would be making these so slow. 
All the scripts/output files for this are located here: `/data/chodera/zhangi/perses_benchmark/repex/38/38/with-perses-paper-test/`, although I just accidentally deleted the log file :( 

For the charge changing (ALA->ARG) mutation, which uses uses n_replicas=36, propagating replicas and computing the energy matrix take longer, as you mentioned above. I think the reason is because we are using so many more replicas.
All the scripts/output files are located here: `/data/chodera/zhangi/perses_benchmark/repex/38/39/`

Also since the solvation procedure has changed in openmm, i just want to double check that i'm using the appropriate amount of solvent. I'm using padding = 1.7 nm. Here is solvated capped ALA:
Screen Shot 2022-07-06 at 10 42 06 AM

jchodera:
I spoke with @dominicrufa about implementing a `LangevinMiddleIntegrator` variant of `LangevinDynamicsMove`, which would give us better configuration-space accuracy. This uses BAOAB, and should be very easy to put into the upcoming `openmmtools` release that @mikemhenry will need to cut.

In general, I think we want to set `reassign_velocities=False` once we have fixed the velocity resume issue. This will reduce correlation times.

The other slowdowns---generating the htf in the same script, as well as the difference between hybrid propagation and repex---are things we might want to look into as well.

zhang-ivy:
@jchodera :  A new `LangevinDynamicsMove` with `LangevinMiddleIntegrator` sounds great, and fixing the velocity resume issue would also be very helpful, but I'm wondering how essential it is for me to wait for these improvements to be able to start generating data for the barnase:barstar paper. I'll set up a time to chat about this (and other software roadblocks I've been experiencing) next week

jchodera:
We should use a BAOAB-based  integrator, which means you cannot use `LangevinDynamicsMove` since it is based on OBABO.

It may be easier just to modify `LangevinDynamicsMove` to use `openmm.LangevinMiddleIntegrator`, since there's no reason to specifically use `openmm.LangevinIntetegrator`. This is a one-line change [here](https://github.com/choderalab/openmmtools/blob/main/openmmtools/mcmc.py#L1145) we should be able to get in right away.

PR #1059, opened 2022-06-24
Bugfix Release 0.10.1

ijpulidos:
## Description

Changelog for bugfix release 0.10.1.

Please check `changelog.rst` for detailed information.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -8,6 +8,25 @@ This section lists features and improvements of note in each release.
 
 The full release history can be viewed `at the GitHub perses releases page <https://github.com/choderalab/perses/releases>`_.
 
+0.10.1 - Release
+----------------
+
+Bugfix release.
+
+Bugfixes
+^^^^^^^^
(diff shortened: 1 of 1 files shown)


zhang-ivy (approved):

zhang-ivy:
@ijpulidos : Looks good to me!

jchodera (approved):
Looks good!

After this release, let's make sure each PR adds its own release note text before it is merged!

ijpulidos:
> After this release, let's make sure each PR adds its own release note text before it is merged!

The correspnding PRs have the release notes, we are not forcing to change the `changelog.rst` file to avoid possible merge conflicts. But we might want to reconsider this, since these should be really easy to solve if they happen at all.

jchodera:
> The corresponding PRs have the release notes, we are not forcing to change the changelog.rst file to avoid possible merge conflicts. But we might want to reconsider this, since these should be really easy to solve if they happen at all.

I think we want each PR to add the new release note. After each release, we can add the headers for each section of the release notes to avoid collisions between PRs.

Issue #1061, opened 2022-06-27
Add host-guest charge changing repex consistency test


zhang-ivy:
Test system: CB7 (or BCD):A1 to A2

Make sure to use cache argument to `SystemGenerator` for all charge changing tests, e.g.:
```python
system_generator = SystemGenerator(forcefields=['amber/ff14SB.xml', 'amber/tip3p_standard.xml'], small_molecule_forcefield='gaff-2.11', forcefield_kwargs=forcefield_kwargs, cache='db.json')

```

zhang-ivy:
@jchodera : I'm having trouble parametrizing the A1 guest because the system_generator doesn't recognize the atom type Ho. Note that I get the same problem when using openff 2.0.0 as well as gaff 2.11.

Here's a stripped down example to reproduce my problem:

```python
from openmoltools import forcefield_generators
from perses.utils.openeye import createOEMolFromSDF, extractPositionsFromOEMol
from openeye import oechem 

from openmmforcefields.generators import SystemGenerator
from openff.toolkit.topology import Molecule
import openmm
from openmm import app, unit

# Load A1
ligand_oemol_old = createOEMolFromSDF("a1.mol2")
ligand_topology_old = forcefield_generators.generateTopologyFromOEMol(ligand_oemol_old)
ligand_positions_old = extractPositionsFromOEMol(ligand_oemol_old)

# Create system generator
forcefield_files = ["amber14/tip3p.xml"]
small_molecule_forcefield = "gaff-2.11"
# small_molecule_forcefield = "openff-2.0.0"
forcefield_kwargs = {'removeCMMotion': False, 'ewaldErrorTolerance': 0.00025, 'constraints' : app.HBonds, 'rigidWater': True, 'hydrogenMass' : 3 * unit.amu}
barostat = openmm.MonteCarloBarostat(1 * unit.atmospheres, 300 * unit.kelvin, 50)
molecules = [ Molecule.from_openeye(oemol,allow_undefined_stereo=True) for oemol in [ligand_oemol_old] ]
system_generator = SystemGenerator(forcefields=forcefield_files, barostat=barostat, 
                                   forcefield_kwargs=forcefield_kwargs,
                                   small_molecule_forcefield=small_molecule_forcefield, 
                                   molecules=molecules,
                                   periodic_forcefield_kwargs = {'nonbondedMethod': app.PME})

# Solvate
modeller = app.Modeller(ligand_topology_old, ligand_positions_old)
[...]

Issue #1058, opened 2022-06-24
Speed up protein mutation charge changing repex consistency test


zhang-ivy:
This test is currently very slow, so we need to speed up. 

See: https://github.com/choderalab/perses/issues/1044#issuecomment-1165575916

zhang-ivy:
I just analyzed the dipeptide charge changing test data: https://github.com/choderalab/perses/issues/1044#issuecomment-1167913833

zhang-ivy:
@ijpulidos : This should speed the test up a lot:
```python
move = mcmc.LangevinDynamicsMove(timestep=4.0 * unit.femtoseconds,
                                                                       collision_rate=1.0 / unit.picosecond,
                                                                       n_steps=50,
                                                                       reassign_velocities=False,
                                                                       n_restart_attempts=20)
```

Note that we want to wait until openmmtools 0.21.5 is released so that we can use the updated `LangevinDynamicsMove` with the `LangevinMiddleIntegrator`

ijpulidos:
By using `openmm-nightly` and `LangevinDynamicsMove` we are getting almost a 4x speedup for the neutral test:

![neutral_77-LangSplitDynMove_vs_nightly-LangDynMove](https://user-images.githubusercontent.com/2949729/185011365-566dfc9a-501d-45ce-9f80-55604415d0d5.png)

This means that for the neutral mutation test, for 1000 iterations, we are talking about ~35 minutes in a 1080Ti GPU (1 s per iteration, 1000 iterations and 2 replicates -- forward and backward --). Which should fit our GPU CI workflow.

For the charge mutation (explicit solvent) the speedup is not that high, we are getting about a 30% increase in performance

![charge_77-LangSplitDynMove_vs_nightly-LangDynMove](https://user-images.githubusercontent.com/2949729/185014788-f8ffaf94-6144-4025-b62c-7427c005a6bd.png)

that would mean the tests would take about ~9 hours for 1000 iterations on a 1080Ti GPU (8 s per iteration, 1000 iterations and 4 replicates as per the test goes)

zhang-ivy:
[...]

PR #1065, opened 2022-06-30
[WIP] Add small molecule repex consistency tests

### LINKED ISSUE #959: Add some A->B and B->A consistency tests to our GPU CI
https://github.com/choderalab/perses/issues/958#issuecomment-1062394797

zhang-ivy:
## Description

Adds neutral and charge changing small molecule tests (the latter one has not been added yet). 

## Motivation and context

Resolves #959 and #1061

## Change log

```

```


zhang-ivy:
@ijpulidos : I've pushed the small molecule neutral and charge tests. The charge test isn't done yet, but I was thinking you could take over from here? The charge test is currently yielding the error described [here](https://github.com/choderalab/perses/issues/1061#issuecomment-1172931985). Once we figure out how to fix that error, you may have to debug a bit more to make sure the test is running without any more errors and then check that its actually passing.

zhang-ivy:
@ijpulidos : Update: I've fixed the parametrization error in the charge test that I mentioned above

ijpulidos:
We need to switch to the new LangevinMiddelIntegrator once we cut a new release of openmmtools.

zhang-ivy:
To speed up this test, see https://github.com/choderalab/perses/issues/1058#issuecomment-1188299488

ijpulidos:
To use the change in `openmmtools` 0.21.5 that uses `LangevinMiddleIntegrator` we would have to change the test to use `LangevinDynamicsMove` instead of `LangevinSplittingDynamicsMove`, but I see that almost all tests (except one) use the latter. How do we want to approach this? Just changing the move would be enough to test what we want to test?

ijpulidos:
> How do we want to approach this? Just changing the move in this test would be enough to test what we want to test?

From discussions with @zhang-ivy , I think we want to change all of the tests to use `LangevinDynamicsMove` which in turn would use the `LangevinMiddleIntegrator` changes in latest openmmtools.

ijpulidos:
As discussed in our dev sync, I saw the following results

comparison_iteration_times

According to log output it would be taking more than 4 hours for completion, which seems like too much compared to what @zhang-ivy has obtained in other tests.

The script to generate the data is attached. I basically just copied the `test_RESTCapableHybridTopologyFactory_repex_charge_transformation` and made it a script. To run you can use `python test_repex_charge.py 2> stderr_out.txt` for example, this would write the stderr output to the specified file. You can generate the plots (assuming you have two log files you want to compare), using the attached `make_plots.py` barebones script (modify as needed).

We should also try to run this using the nightly versions of Openmm to see further performance boosts.

[test_repex_charge.tar.gz](https://github.com/choderalab/perses/files/9339814/test_repex_charge.tar.gz)

perses/tests/test_repex.py
```diff
@@ -54,16 +54,15 @@ def test_RESTCapableHybridTopologyFactory_repex_neutral_mutation():
             # Set up repex simulation
             reporter_file = os.path.join(temp_dir, f"{wt_name}-{mutant_name}.nc")
             reporter = MultiStateReporter(reporter_file, checkpoint_interval=10)
```
jchodera, line 56: We can set the `checkpoint_interval` to be 100 here. There's no need to checkpoint this frequently.

perses/tests/test_repex.py
```diff
@@ -157,16 +155,15 @@ def test_RESTCapableHybridTopologyFactory_repex_charge_mutation():
                 # Set up repex simulation
                 reporter_file = os.path.join(temp_dir, f"{wt_name}-{mutant_name}.nc")
                 reporter = MultiStateReporter(reporter_file, checkpoint_interval=10)
```
jchodera, line 157: We can set the `checkpoint_interval` to 100 here.

perses/tests/test_repex.py
```diff
@@ -157,16 +155,15 @@ def test_RESTCapableHybridTopologyFactory_repex_charge_mutation():
...
+                                                                              collision_rate=1.0 / unit.picosecond,
+                                                                              n_steps=50,
+                                                                              reassign_velocities=False,
+                                                                              n_restart_attempts=20,
+                                                                              constraint_tolerance=1e-06),
+                                         replica_mixing_scheme='swap-all',
+                                         hybrid_factory=htf,
+                                         online_analysis_interval=None)
                 hss.setup(n_states=36, temperature=300 * unit.kelvin, t_max=300 * unit.kelvin,
                           storage_file=reporter, minimisation_steps=0, endstates=True)
```
jchodera, line 168: I think we _may_ need to minimize a bit here, since things could get pathologically slow if there are NaNs.

perses/tests/test_repex.py
```diff
@@ -54,16 +54,15 @@ def test_RESTCapableHybridTopologyFactory_repex_neutral_mutation():
...
+                                                                          collision_rate=1.0 / unit.picosecond,
+                                                                          n_steps=50,
+                                                                          reassign_velocities=False,
+                                                                          n_restart_attempts=20,
+                                                                          constraint_tolerance=1e-06),
+                                     replica_mixing_scheme='swap-all',
+                                     hybrid_factory=htf,
+                                     online_analysis_interval=None)
             hss.setup(n_states=12, temperature=300 * unit.kelvin, t_max=300 * unit.kelvin,
                       storage_file=reporter, minimisation_steps=0, endstates=True)
```
jchodera, line 67: I think we _may_ need to minimize a bit here, since things could get pathologically slow if there are NaNs.

perses/tests/test_repex.py
```diff
@@ -188,5 +185,218 @@ def test_RESTCapableHybridTopologyFactory_repex_charge_mutation():
...
+            custom_force.setUseLongRangeCorrection(False)
+            nonbonded_force.setUseDispersionCorrection(True)
+
+            # Create a reporter
+            selection = "not water"
+            reporter_file = os.path.join(temp_dir, f"{ligand_A_index}_{ligand_B_index}.nc")
+            reporter = MultiStateReporter(
+                reporter_file,
+                analysis_particle_indices=htf.hybrid_topology.select(selection),
+                checkpoint_interval=10)
```
jchodera, line 252: The `checkpoint_interval` here could be much longer, like 100.

perses/tests/test_repex.py
```diff
@@ -188,5 +185,218 @@ def test_RESTCapableHybridTopologyFactory_repex_charge_mutation():
...
+            fe_setup = RelativeFEPSetup(
+                ligand_input=[guest_1_filename, guest_2_filename],
+                receptor_mol2_filename=host_filename,
+                old_ligand_index=ligand_A_index,
+                new_ligand_index=ligand_B_index,
+                forcefield_files=["amber14/tip3p.xml"],
+                small_molecule_forcefield="gaff-2.11",
+                small_molecule_parameters_cache=resource_filename("perses", os.path.join("data", "host-guest", "cache.json")),
+                phases=phases,
+                solvent_padding=1.7 * unit.nanometers)
```
jchodera, line 335: DO we need this much solvent padding here? Perhaps 1 nm would be sufficient?

perses/tests/test_repex.py
```diff
@@ -188,5 +185,218 @@ def test_RESTCapableHybridTopologyFactory_repex_charge_mutation():
...
+                        constraint_tolerance=1e-06),
+                    replica_mixing_scheme='swap-all',
+                    hybrid_factory=htf,
+                    online_analysis_interval=10)
+
+                sampler.setup(
+                    n_states=n_states,
+                    temperature=300 * unit.kelvin,
+                    storage_file=reporter,
+                    minimisation_steps=0,
```
jchodera, line 381: We probably want some minimization steps here.

perses/tests/test_repex.py
```diff
@@ -188,5 +185,218 @@ def test_RESTCapableHybridTopologyFactory_repex_charge_mutation():
...
+                custom_force.setUseLongRangeCorrection(False)
+                nonbonded_force.setUseDispersionCorrection(True)
+
+                # Create a reporter
+                selection = "not water"
+                reporter_file = os.path.join(temp_dir, f"{ligand_A_index}_{ligand_B_index}_{phase}.nc")
+                reporter = MultiStateReporter(
+                    reporter_file,
+                    analysis_particle_indices=htf.hybrid_topology.select(selection),
+                    checkpoint_interval=10)
```
jchodera, line 362: This checkpoint interval can be 100

perses/tests/test_repex.py
```diff
@@ -188,5 +185,218 @@ def test_RESTCapableHybridTopologyFactory_repex_charge_mutation():
...
+            # Generate topology proposal and positions
+            smiles_filename = resource_filename("perses", os.path.join("data", "test.smi"))
+            fe_setup = RelativeFEPSetup(
+                ligand_input=smiles_filename,
+                old_ligand_index=ligand_A_index,
+                new_ligand_index=ligand_B_index,
+                forcefield_files=["amber14/tip3p.xml"],
+                small_molecule_forcefield="gaff-2.11",
+                phases=["solvent"],
+                solvent_padding=1.7 * unit.nanometers)
```
jchodera, line 229: Could we use a smaller padding, like 1 nm?

perses/tests/test_repex.py
```diff
@@ -188,5 +185,218 @@ def test_RESTCapableHybridTopologyFactory_repex_charge_mutation():
...
+                    mcmc_moves=mcmc.LangevinDynamicsMove(
+                        timestep=4.0 * unit.femtoseconds,
+                        collision_rate=1.0 / unit.picosecond,
+                        n_steps=50,
+                        reassign_velocities=False,
+                        n_restart_attempts=20,
+                        constraint_tolerance=1e-06),
+                    replica_mixing_scheme='swap-all',
+                    hybrid_factory=htf,
+                    online_analysis_interval=10)
```
jchodera, line 375: Should `online_analysis_interval` be on here? And should it be so frequent?

perses/tests/test_repex.py
```diff
@@ -188,5 +185,218 @@ def test_RESTCapableHybridTopologyFactory_repex_charge_mutation():
...
+                mcmc_moves=mcmc.LangevinDynamicsMove(
+                    timestep=4.0 * unit.femtoseconds,
+                    collision_rate=1.0 / unit.picosecond,
+                    n_steps=50,
+                    reassign_velocities=False,
+                    n_restart_attempts=20,
+                    constraint_tolerance=1e-06),
+                replica_mixing_scheme='swap-all',
+                hybrid_factory=htf,
+                online_analysis_interval=10)
```
jchodera, line 265: Should `online_analysis_interval` be on here? And should it be so frequent?

perses/tests/test_repex.py
zhang-ivy: Actually, we don't ever use the positions, so we probably don't need to write to the checkpoint file at all here. 
The energies should be written every iteration regardless of the checkpoint_interval.

perses/tests/test_repex.py
zhang-ivy: `minimisation_steps=0` just means that minimization will run as many steps as needed until the energy is below the specified tolerance

perses/tests/test_repex.py
zhang-ivy: With openmm > 7.7, we cannot use a smaller padding, because otherwise we see this error: https://github.com/choderalab/perses/issues/949

perses/tests/test_repex.py
zhang-ivy: We should actually set it to 0, since we don't need the results of online analysis for the test.

jchodera: Should it be `0` or `None`?

perses/tests/test_repex.py
jchodera: This seems like incorrect behavior: 
* `None` should be interpreted to run until tolerance is reached
* `0` should be interpreted to run 0 steps

zhang-ivy: I think the reason why `minimisation_steps=0` means that it will run as many steps as needed is because this matches the default openmm behavior: http://docs.openmm.org/latest/api-python/generated/openmm.openmm.LocalEnergyMinimizer.html#openmm.openmm.LocalEnergyMinimizer.minimize

Maybe we could change `minimisation_steps` to `max_iterations` so that the matching behavior is clearer.

jchodera: `max_iterations=0` would still imply that the maximum number of iterations should be zero (no minimization).

The OpenMM choice was a poor default---it should have been -1 or some other choice.
Python has `None`, which should be the appropriate "please use the default" choice. 0 should mean 0 steps.

zhang-ivy: Ok. I'll open an issue about this

perses/tests/test_repex.py
ijpulidos: using 100 here for the `checkpoint_interval` worked okay in the preliminary tests

perses/tests/test_repex.py
ijpulidos: This might need to be reconsidered. I think this changed along the way, since I've been using the default value of 1.1 nm for the `PointMutationExecutor` with the openmm nightly build and it is working correctly. I'll make tests to try to come up with a good value.

perses/tests/test_repex.py
ijpulidos: As far as I can tell, we don't really need it, I think we can just leave it as the default value of 200.

ijpulidos:
While debugging the performance of this test a solution to issue #1085 was required, I updated the description of this PR to reflect this.

perses/tests/test_repex.py
ijpulidos: Resolved by https://github.com/choderalab/perses/pull/1065/files#r947881930

perses/tests/test_repex.py
ijpulidos: The issue is #1086 I'm considering this as "resolved". Feel free to re-open otherwise.

perses/tests/test_repex.py
zhang-ivy: @ijpulidos : Were you able to reproduce the `NonbondedForce: The cutoff distance cannot be greater than half the periodic box size` with openmm 7.7?

perses/rjmc/geometry.py
```diff
@@ -450,8 +451,6 @@ def _logp_propose(self, top_proposal, old_positions, beta, new_positions=None, d
         if self._storage:
             self._storage.write_object("{}_proposal_order".format(direction), proposal_order_tool, iteration=self.nproposed)
 
-        platform_name = 'CUDA'
```
zhang-ivy, line 453: @ijpulidos : I don't think we want to make this fix here. I think we want to open a separate PR to fix this, as both the protein mutation and small molecule tests need this fix. Then we'll want to open/merge a PR that speeds up the protein mutation tests, then finally we can merge this PR #1065.

perses/tests/test_repex.py
```diff
@@ -80,7 +80,7 @@ def test_RESTCapableHybridTopologyFactory_repex_neutral_mutation():
 
         DDG = abs(data['ala-thr']['free_energy'] - data['thr-ala']['free_energy'] * -1)
         dDDG = np.sqrt(data['ala-thr']['error'] ** 2 + data['thr-ala']['error'] ** 2)
-        assert DDG < 6 * dDDG, f"DDG ({DDG}) is greater than 6 * dDDG ({6  * dDDG})"
+        assert DDG < 3 * dDDG, f"DDG ({DDG}) is greater than 3 * dDDG ({3  * dDDG})"
```
zhang-ivy, line 83: @ijpulidos : Why did you change this from 6 to 3? I think we want to keep it as 6, based on these results: 
https://github.com/choderalab/perses/issues/1044#issuecomment-1157006442
https://github.com/choderalab/perses/issues/1044#issuecomment-1167913833

perses/tests/test_repex.py
zhang-ivy: @ijpulidos I think #1086 is a different issue and it hasn't been resolved yet. Can you clarify what you mean?

perses/tests/test_repex.py
ijpulidos: I see, I remember discussing about making it 3*dDDG. So, would this be 6*dDDG for the neutral test and 3*dDDG for the charge mutation test or should we just keep 6 for both?

perses/rjmc/geometry.py
ijpulidos: Fair enough, I agree that helps keeping track of things. It was just easier for me to test the consequences of this fix in this same branch, but I agree we should probably want a different one for this fix for bookkeeping.

perses/tests/test_repex.py
zhang-ivy: I think we'd want to keep 6 for both. I also was thinking you could open a separate PR for speeding up the neutral mutation tests, although its probably fine to make them here -- you'd just need to update the description and title of this PR -- up to you.

ijpulidos: Yeah, I think it is better to make them here, since we are changing the same file, this could avoid potential merge conflicts. I'll update the description.

zhang-ivy:
Notes from discussion with Iván today: 

- We are seeing that small molecule (CCC->CCCC) neutral test fails -- i.e. there is a large discrepancy between forward and reverse transformations. 
```
n_iterations: 1000, DDG (2.0556889259223965), 6 * dDDG (1.8660036814494445)
n_iterations: 2000, DDG (1.9424681339279566), 6 * dDDG (1.504043152162713)
n_iterations: 3000, DDG (2.3012075072431486), 6 * dDDG (1.3496196586095943)
n_iterations: 4000, DDG (2.105869914487293), 6 * dDDG (1.1557338251054274)
n_iterations: 5000, DDG (2.0091207197835246), 6 * dDDG (0.9972790553988209)
n_iterations: 6000, DDG (2.1447482384645653), 6 * dDDG (0.9375579903514)
n_iterations: 7000, DDG (1.8431263594498581), 6 * dDDG (0.8343179445910711)
n_iterations: 8000, DDG (1.9820627460818656), 6 * dDDG (0.7713827653749263)
n_iterations: 9000, DDG (1.9790420115873495), 6 * dDDG (0.758837900156935)
n_iterations: 10000, DDG (2.32585423139), 6 * dDDG (0.720859505508521)
```
- We checked whether the energy validation tests are passing. They are failing, but only by ~0.2 kT. When the dispersion correction is off, the test passes. I think its ok to proceed despite the small discrepancy. I investigated this in detail and wrote up notes here #1098 
- We also checked the yank analysis notebook for these transformations, and found that the statistical inefficiency is super large:
image (15)

Any idea why the statistical inefficiency is so large for this transformation? 

When we look at the same plot for a tyk2 ligand solvent phase transformation, we do not see such a large statistical inefficiency: 
image (14)

I wonder if we should refrain from using CCC->CCCC and instead use a tyk2 ligand solvent phase transformation instead?

zhang-ivy:
We should choose a transformation with fewer rotatable bonds in one state but still has a large free energy difference, like paracetamol to phenol

ijpulidos:
@zhang-ivy Run the neutral transformation test with the Vanilla HTF and the results for the DDG and dDDG are as follow:
```
n_iterations: 1000, DDG (4.379856483835027), 6 * dDDG (6.427969308929397)
n_iterations: 2000, DDG (5.104396507573654), 6 * dDDG (5.103675724302308)
n_iterations: 3000, DDG (5.05622797342096), 6 * dDDG (4.150472778157009)
n_iterations: 4000, DDG (5.661439924925766), 6 * dDDG (3.9788164082347923)
n_iterations: 5000, DDG (4.924333536929069), 6 * dDDG (3.4960583717742804)
n_iterations: 6000, DDG (5.675488856850542), 6 * dDDG (3.3866050177001448)
n_iterations: 7000, DDG (5.928197475123568), 6 * dDDG (3.018527704140613)
n_iterations: 8000, DDG (5.141353236248172), 6 * dDDG (2.786667888092715)
n_iterations: 9000, DDG (5.0989841273643535), 6 * dDDG (2.5409050042675574)
n_iterations: 10000, DDG (5.524543006009203), 6 * dDDG (2.5493728345032727)
```
The equilibration plot is the following
image

zhang-ivy:
@ijpulidos : Can you try generating the same plot with openmmtools 0.21.4? Just want to see if those results also yield a large statistical inefficiency.

zhang-ivy:
@ijpulidos : Did you run phenol to paracetamol with the rest htf as well?

ijpulidos:
> @ijpulidos : Did you run phenol to paracetamol with the rest htf as well?

RestHTF gives better results for the DDG vs. dDDG comparison
```
n_iterations: 1000, DDG (0.7389286664047461), 6 * dDDG (4.349966796166879)
n_iterations: 2000, DDG (1.4476000767035657), 6 * dDDG (3.35301957925458)
n_iterations: 3000, DDG (1.0050371228929293), 6 * dDDG (2.877359027173167)
n_iterations: 4000, DDG (0.5892713335768889), 6 * dDDG (2.4970385474617656)
n_iterations: 5000, DDG (0.08935214968687433), 6 * dDDG (2.2207918586467272)
n_iterations: 6000, DDG (0.7799885336397239), 6 * dDDG (2.0505836836904154)
n_iterations: 7000, DDG (0.17155683937092192), 6 * dDDG (1.90492611578378)
n_iterations: 8000, DDG (0.06495242783663002), 6 * dDDG (1.7822284655187324)
n_iterations: 9000, DDG (0.294242113389501), 6 * dDDG (1.692400658436612)
n_iterations: 10000, DDG (0.05825571585386058), 6 * dDDG (1.6147951072659277)
```
The equilibration plot for it is as follows
image

I found a "strange" thing happening with the mapper, which may explain the differences. Left is RestHTF, right is Vanilla HTF
image

ijpulidos:
> I found a "strange" thing happening with the mapper, which may explain the differences. Left is RestHTF, right is Vanilla HTF 

I don't think this is a bug, it's just that the `debug.png` file getting generated [here](https://github.com/choderalab/perses/blob/cd51c649d45204f7e0ee88787eedb2dfc24ea3a7/perses/rjmc/atom_mapping.py#L933) is not the best mapping but just whatever was last in the `atom_mappings` set. We probably do want to export the best mapping instead, though.

zhang-ivy:
Notes from dev sync today:
- We should try setting `n_steps` per iteration = 250 instead of 50. This will hopefully cause the subsample rate to decrease and should decrease the number of iterations needed for DDG < 6 * dDDG.
- We still don't know why the test fails with the vanilla htf and not the rest htf -- need to investigate this.

mikemhenry:
I've updated this branch/PR with updated logging + using the new openmm8 rc so we can see how that does.

mikemhenry:
These are the slowest tests 
```
5637.75s call     perses/tests/test_repex.py::test_RESTCapableHybridTopologyFactory_repex_charge_mutation
1613.93s call     perses/tests/test_repex.py::test_RESTCapableHybridTopologyFactory_repex_neutral_mutation
1569.07s call     perses/tests/test_repex.py::test_RESTCapableHybridTopologyFactory_repex_neutral_transformation
1179.05s call     perses/tests/test_repex.py::test_RESTCapableHybridTopologyFactory_repex_charge_transformation
```
Need to fix (don't need to fix in this PR):
`subprocess.run("perses-relative test.yml", shell=True, check=True, env=env)`


`test_RESTCapableHybridTopologyFactory_repex_charge_transformation`

```
FAILED perses/tests/test_relative.py::test_RESTCapableHybridTopologyFactory_energies_GPU - ValueError: Could not locate file "gaff.xml"
FAILED perses/tests/test_relative.py::test_unsampled_endstate_energies_GPU - ValueError: Could not locate file "gaff.xml"
FAILED perses/tests/test_repex.py::test_RESTCapableHybridTopologyFactory_repex_neutral_transformation - AssertionError: DDG (2.151708958621054) is greater than 6 * dDDG (1.1169450391489755)
FAILED perses/tests/test_repex.py::test_RESTCapableHybridTopologyFactory_repex_charge_transformation - AssertionError: DDG (22.989485678075503) is greater than 6 * dDDG (12.23018630199661)
FAILED perses/tests/test_resume.py::test_cli_resume_repex - subprocess.CalledProcessError: Command 'perses-relative test.yml' returned non-zero exit ...
```
```
          assert DDG < 6 * dDDG, f"DDG ({DDG}) is greater than 6 * dDDG ({6 * dDDG})"
E           AssertionError: DDG (22.989485678075503) is greater than 6 * dDDG (12.23018630199661)
E           assert 22.989485678075503 < (6 * 2.038364383666101
```

mikemhenry:
```
FAILED perses/tests/test_repex.py::test_RESTCapableHybridTopologyFactory_repex_neutral_transformation - AssertionError: DDG (1.9296819854921097) is greater than 6 * dDDG (0.8521151431457088)
FAILED perses/tests/test_repex.py::test_RESTCapableHybridTopologyFactory_repex_charge_transformation - AssertionError: DDG (23.089451730916885) is greater than 6 * dDDG (11.498215681031873)
```

zhang-ivy:
@ijpulidos @mikemhenry : The first thing I would do to fix these failing tests is to increase the number of steps per iteration to 250 (as we discussed [earlier](https://github.com/choderalab/perses/pull/1065#issuecomment-1244367915))

If that doesn't fix things, I would increase the number of iterations until the tests pass.

zhang-ivy:
It might be best to run these locally (on lilac or your external GPU) first to make sure you can get the tests to converge with the changes I recommended

ijpulidos:
As far as I've been testing, the protein mutation tests pass with 1000 iterations and 250 steps (`n_steps`). 
Whereas the small molecule transformations fail even with 2000 iterations and 250 steps. The time they took are as follows (these on different GPUs, it is hard to test on same gpu due to availability):
Passing tests
```
19356.48s call     perses/tests/test_repex.py::test_RESTCapableHybridTopologyFactory_repex_charge_mutation
1558.90s call     perses/tests/test_repex.py::test_RESTCapableHybridTopologyFactory_repex_neutral_mutation
```
Failing tests
```
10512.75s call     perses/tests/test_repex.py::test_RESTCapableHybridTopologyFactory_repex_charge_transformation
3911.30s call     perses/tests/test_repex.py::test_RESTCapableHybridTopologyFactory_repex_neutral_transformation
```
That means we are talking about at least 10 hrs for these tests to run.

ijpulidos:
I think we already noticed the issues with the small molecule transformations tests in https://github.com/choderalab/perses/pull/1065#issuecomment-1243797927 I'll try running the tests with the transformation suggested in that thread and check the results.

zhang-ivy:
@ijpulidos : Can you try running the small molecule tests even longer (5 ns)? If they are still failing even with 5000 iterations, there may be a more significant problem here.

zhang-ivy:
Ah actually ignore my above comment, I think you're right that we should change the tests to use phenol to paracetamol. But not sure which transformation to use for the small molecule charge changing tests..

ijpulidos (approved):
I think this is finally ready to be merged. Thanks all for your work on this!

We still need to figure out a better test for the charge small molecule transformation, but this will do for now.

ijpulidos:
Tests failing are unrelated to the changes of this PR, will address these in a different PR.

zhang-ivy:
I can review this later tonight or tomorrow! Please wait to merge. Thanks!On Feb 20, 2023, at 18:26, Iván Pulido ***@***.***> wrote:﻿
Tests failing are unrelated to the changes of this PR, will address these in a different PR.

perses/tests/test_repex.py
```diff
@@ -188,5 +189,229 @@ def test_RESTCapableHybridTopologyFactory_repex_charge_mutation():
                        + data['ala-arg']['error'] ** 2
                        + data['lys-ala']['error'] ** 2
                        + data['ala-lys']['error'] ** 2)
+        assert DDG < 6 * dDDG, f"DDG ({DDG}) is greater than 6 * dDDG ({6 * dDDG})"
+
+
+@pytest.mark.gpu_needed
+def test_RESTCapableHybridTopologyFactory_repex_neutral_transformation():
+    """
+    Run CCC->CCCC and CCCC->CCC repex with the RESTCapableHybridTopologyFactory and make sure that the free energies are
```
zhang-ivy, line 198: If we are going to merge this with phenol and paracetamol, we'll want to update the doc string here.

perses/tests/test_repex.py
```diff
@@ -188,5 +189,229 @@ def test_RESTCapableHybridTopologyFactory_repex_charge_mutation():
...
+
+    data = {}
+    n_iterations = 1000
+    n_states = 12
+    transformations = [(0, 1), (1, 0)]  # Ligand indices to use for transformation
+
+    with enter_temp_directory() as temp_dir:
+        for ligand_A_index, ligand_B_index in transformations:
+            # Generate topology proposal and positions
+            # smiles_filename = resource_filename("perses", os.path.join("data", "test.smi"))
```
zhang-ivy, line 225: probably will want to remove this commented out line as well

perses/tests/test_repex.py
```diff
@@ -188,5 +189,229 @@ def test_RESTCapableHybridTopologyFactory_repex_charge_mutation():
...
+            sampler.sampler_context_cache = cache.ContextCache(capacity=None, time_to_live=None, platform=platform)
+
+            # Run repex
+            sampler.extend(n_iterations)
+
+            # Retrieve free energy
+            reporter.close()
+            reporter = MultiStateReporter(reporter_file)
+            analyzer = MultiStateSamplerAnalyzer(reporter, max_n_iterations=n_iterations)
+            f_ij, df_ij = analyzer.get_free_energy()
```
zhang-ivy, line 298: Ah I don't know why I didn't think of this before -- we may want to try bootstrapping the uncertainty. This will make the uncertainty larger (otherwise, MBAR underestimates the uncertainty) which may make the test pass!

perses/tests/test_repex.py
```diff
@@ -188,5 +189,229 @@ def test_RESTCapableHybridTopologyFactory_repex_charge_mutation():
...
+    from perses.samplers.multistate import HybridRepexSampler
+    from perses.annihilation.relative import RESTCapableHybridTopologyFactory
+
+    from openmmtools.multistate import MultiStateReporter, MultiStateSamplerAnalyzer
+    from openmmtools import cache, utils, mcmc
+    platform = configure_platform(utils.get_fastest_platform().getName())
 
+    data = {}
+    n_iterations = 2000
+    n_states = 12
```
zhang-ivy, line 331: The charge-changing mutation test requires 36 states to pass, so I wonder if we need to increase to 36 states here as well to get the test to pass.

perses/tests/test_repex.py
ijpulidos: Yes, I think we had the same question with @dominicrufa . This may be needed not only for the tests but in general for analysis, I think.

perses/tests/test_repex.py
zhang-ivy: Did you make sure to bootstrap the uncertainties? That might help make the test past. Otherwise, I agree with you, might be better to use tyk2 ligands.

ijpulidos: You are right, this was without bootstrapping. I'll check with bootstrapping.

perses/tests/test_repex.py
```diff
@@ -182,11 +185,243 @@ def test_RESTCapableHybridTopologyFactory_repex_charge_mutation():
...
+                sampler.extend(n_iterations)
+
+                # Retrieve free energy
+                reporter.close()
+                reporter = MultiStateReporter(reporter_file)
+                analyzer = MultiStateSamplerAnalyzer(reporter, max_n_iterations=n_iterations)
+                # f_ij, df_ij = analyzer.get_free_energy()
+                # Extract uncorrelated energy matrix (u_ln) and samples from states (N_l)
+                energy_matrix = analyzer._unbiased_decorrelated_u_ln
+                sampled_states = analyzer._unbiased_decorrelated_N_l
```
zhang-ivy, line 414: we actually want to feed MBAR the unsampled states as well, which is why I normally compute the reduced potential matrix (aka uncorrelated energy matrix) like this: https://github.com/choderalab/perses-barnase-barstar-paper/blob/main/scripts/05_analyze/analysis_tools.py#L104
This is minor and shouldn’t affect the test results, since that unsampled endstate effect should cancel out when computing the DDG. But I think we want to include the unsampled endstates here to make sure that its testing the "normal" pipeline for repex.

perses/tests/test_repex.py
```diff
@@ -182,11 +185,243 @@ def test_RESTCapableHybridTopologyFactory_repex_charge_mutation():
...
+                # Compute free energies with boostrapping using pymbar
+                mbar = pymbar.MBAR(energy_matrix, sampled_states, nbootstraps=200)
+                f_ij, df_ij = mbar.getFreeEnergyDifferences(uncertainty_method="bootstrap", return_theta=False)
+
+                data[f"{ligand_A_index}-{ligand_B_index}_{phase}"] = {'free_energy': f_ij[0, -1], 'error': df_ij[0, -1]}
+
+        forward_DG = data['0-1_complex']['free_energy'] - data['0-1_solvent']['free_energy']
+        reverse_DG = data['1-0_complex']['free_energy'] - data['1-0_solvent']['free_energy']
+        DDG = abs(forward_DG - reverse_DG)
```
zhang-ivy, line 423: We may want to avoid using `DDG` as the variable name — this makes sense for the tests with 2 legs but is may be misleading for tests with 4 legs — this is my mistake.

zhang-ivy: and we'll want to remember to fix the sign here: ` DDG = abs(forward_DG - reverse_DG)`

ijpulidos: Should be already done in https://github.com/choderalab/perses/pull/1065/commits/89771675edd04414c4d9d82edbc1efe1e7624f7f

perses/tests/test_repex.py
ijpulidos: I think you mean decorrelated samples instead of unsampled states? Or what's the difference here? Please check the changes in https://github.com/choderalab/perses/pull/1065/commits/59dbceae8ccc8161a502f07e4de6ef52b122b897 and let me know if this works, thanks!

perses/tests/test_repex.py
```diff
@@ -79,12 +82,17 @@ def test_RESTCapableHybridTopologyFactory_repex_neutral_mutation():
             f_ij, df_ij = analyzer.get_free_energy()
             data[f"{wt_name}-{mutant_name}"] = {'free_energy': f_ij[0, -1], 'error': df_ij[0, -1]}
 
-        DDG = abs(data['ala-thr']['free_energy'] - data['thr-ala']['free_energy'] * -1)
-        dDDG = np.sqrt(data['ala-thr']['error'] ** 2 + data['thr-ala']['error'] ** 2)
-        assert DDG < 6 * dDDG, f"DDG ({DDG}) is greater than 6 * dDDG ({6  * dDDG})"
+        forward_fe = data['ala-thr']['free_energy']
+        reverse_fe = data['thr-ala']['free_energy']  # This should have the inverse sign of the forward fe
+        # they should add up to close to zero
+        forward_reverse_diff = abs(forward_fe + reverse_fe)
```
zhang-ivy, line 88: should we call this `forward_reverse_sum` since its a sum and not a difference? same with the line below it, maybe we should call it `forward_reverse_sum_error`

perses/tests/test_repex.py
```diff
@@ -182,11 +190,249 @@ def test_RESTCapableHybridTopologyFactory_repex_charge_mutation():
...
-              - (data['lys-ala']['free_energy'] + data['ala-lys']['free_energy'])
-        dDDG = np.sqrt(data['arg-ala']['error'] ** 2
-                       + data['ala-arg']['error'] ** 2
-                       + data['lys-ala']['error'] ** 2
-                       + data['ala-lys']['error'] ** 2)
+        forward_fe = data['arg-ala']['free_energy'] + data['ala-lys']['free_energy']
+        forward_fe_error = np.sqrt(data['arg-ala']['error']**2 + data['ala-lys']['error']**2)
+        reverse_fe = data['lys-ala']['free_energy'] + data['ala-arg']['free_energy']
+        reverse_fe_error = np.sqrt(data['lys-ala']['error']**2 + data['ala-arg']['error']**2)
+
```
zhang-ivy, line 197: I don't think these free energies and errors are broken down properly. I think something like this would be better:
```
arg_fe = data['arg-ala']['free_energy'] + data['ala-arg']['free_energy']
lys_fe = data['lys-ala']['free_energy'] + data['ala-lys']['free_energy']
arg_lys_diff = abs(arg_fe - lys_fe)
arg_lys_diff_error = np.sqrt(data['arg-ala']['error'] ** 2
                       + data['ala-arg']['error'] ** 2
                       + data['lys-ala']['error'] ** 2
                       + data['ala-lys']['error'] ** 2)
```
Note that `arg_lys_diff = abs(arg_fe - lys_fe)` should in fact be a difference, because we expect the ARG->ALA->ARG leg to have an equivalent free energy to that of LYS->ALA->LYS.

mikemhenry:
Tests took 6hr30min https://github.com/choderalab/perses/actions/runs/4747425819/jobs/8432377018 
 
with only one failure: 
```
FAILED perses/tests/test_repex.py::test_RESTCapableHybridTopologyFactory_repex_charge_mutation - FileNotFoundError: [Errno 2] No such file or directory: '/actions-runner/_work/perses/perses/perses/data/ala_solvated.cif'
```

ijpulidos:
@mikemhenry that's great news. The failing test is just that we have the inverse order in the transformation, I made the changes but forgot to push them, these will add a few hours to the tests but I think we should be okay. I'll push them in a second.

perses/tests/test_repex.py
```diff
@@ -182,11 +190,249 @@ def test_RESTCapableHybridTopologyFactory_repex_charge_mutation():
...
+                # Run repex
+                sampler.extend(n_iterations)
+
+                # Retrieve free energy
+                reporter.close()
+                reporter = MultiStateReporter(reporter_file)
+                analyzer = MultiStateSamplerAnalyzer(reporter, max_n_iterations=n_iterations)
+                # Extract uncorrelated energy matrix (u_ln) and samples from states (N_l)
+                energy_matrix, sampled_states = analyzer._compute_mbar_decorrelated_energies()
```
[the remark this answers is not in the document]
ijpulidos: Thanks, I think `samples_per_state` is a better name, indeed.

perses/tests/test_repex.py
```diff
@@ -182,11 +190,249 @@ def test_RESTCapableHybridTopologyFactory_repex_charge_mutation():
...
+                # Compute free energies with boostrapping using pymbar
+                mbar = pymbar.MBAR(energy_matrix, sampled_states, nbootstraps=200)
+                f_ij, df_ij = mbar.getFreeEnergyDifferences(uncertainty_method="bootstrap", return_theta=False)
+
+                data[f"{ligand_A_index}-{ligand_B_index}_{phase}"] = {'free_energy': f_ij[0, -1], 'error': df_ij[0, -1]}
+
+        forward_fe = data['0-1_complex']['free_energy'] - data['0-1_solvent']['free_energy']
+        reverse_fe = data['1-0_complex']['free_energy'] - data['1-0_solvent']['free_energy']
+        # NOTE: reverse_DG should have the inverse sign compared to forward_DG, so they should ADD up to close to zero
+        forward_reverse_difference = abs(forward_fe + reverse_fe)
```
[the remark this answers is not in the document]
ijpulidos: Should be solved by changes in https://github.com/choderalab/perses/pull/1065/commits/629b20e4de1ab9fc6fdab4af6bba830e6d4e4aed

perses/tests/test_repex.py
```diff
@@ -182,11 +190,249 @@ def test_RESTCapableHybridTopologyFactory_repex_charge_mutation():
...
+            reporter.close()
+            reporter = MultiStateReporter(reporter_file)
+            analyzer = MultiStateSamplerAnalyzer(reporter, max_n_iterations=n_iterations)
+            f_ij, df_ij = analyzer.get_free_energy()
+            data[f"{ligand_A_index}-{ligand_B_index}"] = {'free_energy': f_ij[0, -1], 'error': df_ij[0, -1]}
+
+        forward_fe = data['0-1']['free_energy']
+        reverse_fe = data['1-0']['free_energy']
+        # Note that DDG_reverse should have the inverse sign compared to DDG_forward
+        forward_reverse_diff = abs(forward_fe + reverse_fe)  # This should ADD up to close to zero
```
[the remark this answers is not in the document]
ijpulidos: Solved by changes in https://github.com/choderalab/perses/pull/1065/commits/629b20e4de1ab9fc6fdab4af6bba830e6d4e4aed

perses/tests/test_repex.py
ijpulidos: Should be solved by changes in https://github.com/choderalab/perses/pull/1065/commits/629b20e4de1ab9fc6fdab4af6bba830e6d4e4aed

perses/tests/test_repex.py
ijpulidos: Solved by changes in https://github.com/choderalab/perses/pull/1065/commits/629b20e4de1ab9fc6fdab4af6bba830e6d4e4aed

perses/tests/test_repex.py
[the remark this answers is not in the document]
ijpulidos: Solved in https://github.com/choderalab/perses/pull/1065/commits/1f7e414463cdbe4c30fe6022c5348fb24496c1ab

.github/workflows/self-hosted-gpu-test.yml
```diff
@@ -22,96 +21,93 @@ jobs:
...
             ]
+
   do-the-job:
     name: Do the job on the runner
     needs: start-runner # required to start the main job when the runner is ready
     runs-on: ${{ needs.start-runner.outputs.label }} # run the job on the newly created runner
+    timeout-minutes: 1200 # 20 hrs
     env:
       TEST_MODE: GPU
       OPENMM: ${{ matrix.cfg.openmm }}
```
mikemhenry, line 48:
```suggestion
      OPENMM: 8.0
```

.github/workflows/self-hosted-gpu-test.yml
```diff
@@ -22,96 +21,93 @@ jobs:
...
+
   do-the-job:
     name: Do the job on the runner
     needs: start-runner # required to start the main job when the runner is ready
     runs-on: ${{ needs.start-runner.outputs.label }} # run the job on the newly created runner
+    timeout-minutes: 1200 # 20 hrs
     env:
       TEST_MODE: GPU
-      OPENMM: ${{ matrix.cfg.openmm }}
+      OPENMM: 8.0
```
mikemhenry, line 48: We will need this for when we merge in https://github.com/choderalab/perses/pull/1186

perses/tests/test_repex.py
zhang-ivy: @ijpulidos : Reminder to remove this commented out line

zhang-ivy:
@ijpulidos : I just reviewed again and your changes look good to me. 

A couple of other things to address: 
- https://github.com/choderalab/perses/pull/1065#issuecomment-1513299921
- https://github.com/choderalab/perses/pull/1065/files#r1174114461

ijpulidos:
> Tests took 6hr30min https://github.com/choderalab/perses/actions/runs/4747425819/jobs/8432377018
> 
> with only one failure:
> 
> ```
> FAILED perses/tests/test_repex.py::test_RESTCapableHybridTopologyFactory_repex_charge_mutation - FileNotFoundError: [Errno 2] No such file or directory: '/actions-runner/_work/perses/perses/perses/data/ala_solvated.cif'
> ```

@mikemhenry Last run of the manually run GPU CI tests was successful https://github.com/choderalab/perses/actions/runs/4766605648/jobs/8473881768#step:9:571

@zhang-ivy thanks for the comments/review. I'll be adressing these as soon as possible.

perses/tests/test_repex.py
```diff
@@ -53,17 +55,18 @@ def test_RESTCapableHybridTopologyFactory_repex_neutral_mutation():
...
-                                                                                   n_steps=50,
-                                                                                   reassign_velocities=False,
-                                                                                   n_restart_attempts=20,
-                                                                                   splitting="V R R R O R R R V",
-                                                                                   constraint_tolerance=1e-06),
-                                                                                   replica_mixing_scheme='swap-all',
-                                    hybrid_factory=htf,
-                                    online_analysis_interval=None)
+                                                                          collision_rate=1.0 / unit.picosecond,
+                                                                          n_steps=250,
```
jchodera, line 62: Would this still work if we reduced `n_steps` to 125?

perses/tests/test_repex.py
```diff
@@ -156,17 +161,18 @@ def test_RESTCapableHybridTopologyFactory_repex_charge_mutation():
...
-                                                                                       n_steps=50,
-                                                                                       reassign_velocities=False,
-                                                                                       n_restart_attempts=20,
-                                                                                       splitting="V R R R O R R R V",
-                                                                                       constraint_tolerance=1e-06),
-                                                                                       replica_mixing_scheme='swap-all',
-                                        hybrid_factory=htf,
-                                        online_analysis_interval=None)
+                                                                              collision_rate=1.0 / unit.picosecond,
+                                                                              n_steps=250,
```
jchodera, line 168: Could we get by with reducing `n_steps` to 125?

perses/tests/test_repex.py
```diff
@@ -182,11 +188,249 @@ def test_RESTCapableHybridTopologyFactory_repex_charge_mutation():
...
+                reporter_file,
+                analysis_particle_indices=htf.hybrid_topology.select(selection),
+                checkpoint_interval=100)
+
+            # Build the hybrid repex sampler
+            sampler = HybridRepexSampler(
+                mcmc_moves=mcmc.LangevinDynamicsMove(
+                    timestep=4.0 * unit.femtoseconds,
+                    collision_rate=1.0 / unit.picosecond,
+                    n_steps=250,
```
jchodera, line 279: Could this still work with `n_steps=125`?

perses/tests/test_repex.py
```diff
@@ -182,11 +188,249 @@ def test_RESTCapableHybridTopologyFactory_repex_charge_mutation():
...
+                    reporter_file,
+                    analysis_particle_indices=htf.hybrid_topology.select(selection),
+                    checkpoint_interval=100)
+
+                # Build the hybrid repex sampler
+                sampler = HybridRepexSampler(
+                    mcmc_moves=mcmc.LangevinDynamicsMove(
+                        timestep=4.0 * unit.femtoseconds,
+                        collision_rate=1.0 / unit.picosecond,
+                        n_steps=250,
```
jchodera, line 394: Could this still work with `n_steps=125`?

jchodera (commented):
The GPU tests took 10 hours (~$5 per run), with 8.5 hours spent in these two tests:
```
16633.61s call     perses/tests/test_repex.py::test_RESTCapableHybridTopologyFactory_repex_charge_mutation
13661.38s call     perses/tests/test_repex.py::test_RESTCapableHybridTopologyFactory_repex_charge_transformation
```
* Are we sure the tests are running on the GPU? We aren't logging debug output that would report the Platform that the tests are running on, but we should probably confirm this.
* Is there any way we can reduce the system size for these tests so that runtime is shortened?
* I've highlighted lines where we might be able to halve the number of steps per iteration, which would cut the runtime in half.

If we can't reduce the cost drastically, we might have to only enable these tests in special circumstances, such as once per day on `main`.

ijpulidos:
Good news, we can use 125 steps per iteration and still have the tests passing, we probably could do a bit of more refining here to lower the walltimes, but we moved from ~10 hours, to ~6.5 hours [in the latest manual run of these tests here](https://github.com/choderalab/perses/actions/runs/4810681119/jobs/8563699893#step:9:571).

perses/tests/test_repex.py
ijpulidos: Changes in https://github.com/choderalab/perses/pull/1065/commits/693f41c45984fa35364005021c7581f46a620382

ijpulidos:
And as we discussed in our dev sync, these tests would only be run on a nightly basis, and maybe we can discuss about running them less frequently if we need to. Monitoring the costs here shouldn't be hard on the AWS dashboard.

.github/workflows/self-hosted-gpu-test.yml
```diff
@@ -2,12 +2,11 @@ name: self-hosted-gpu-test
...
   push:
     branches:
-      - master
       - main
   workflow_dispatch:
   schedule:
-    # nightly tests
-    - cron: "0 0 * * *"
+    # weekly tests
+    - cron: "0 0 * * SUN"
```
mikemhenry, line 9:
```suggestion
    # nightly tests
    - cron: "0 0 * * *"
```

mikemhenry (approved):

PR #1066, opened 2022-07-06
[WIP] Protocol : Nonequilibrium Cycling

dotsdl:
## Description

A nonequilibrium cycling protocol supporting execution with OpenMM, using the protocol structure defined in [`gufe`](https://github.com/OpenFreeEnergy/gufe).

## Motivation and context

This is one of the first protocols to be developed using the `gufe` structure, and is intended to surface deficiencies in that structure.  The end result will be a protocol that can be executed locally (with `Protocol.execute()`) as well as on cluster resources using a suitable Scheduler (e.g. an HPC scheduler like that being developed by OpenFE).

## Immediate needs
- [ ] docstrings on all methods required or defined

## How has this been tested?

- [ ] unit tests of protocol units
- [ ] unit tests of NonEquilibriumCyclingProtocol
- [ ] unit tests of NonEquilibriumCyclingResult

## Change log

```
New features
^^^^^^^^^^^^
- `perses.protocols` submodule added; features `NonEquilibriumCyclingProtocol`

```


zhang-ivy:
Suggestions for names other than `Protocol`: experiment, runner, workflow, pipeline

dotsdl:
Ah, I think I'm still inclined to call this `Protocol`, since in the context of [`gufe`](https://github.com/OpenFreeEnergy/gufe) this defines the procedure by which we execute the simulations necessary to calculate a `\Delta G`. A `Protocol` produces a `ProtocolDAG` from a set of input `ChemicalSystem`s, an atom mapping (if necessary), and potentially any previous result that should be used as a starting point. 

The `ProtocolDAG` is more akin to an individual experiment or workflow in this sense, while the `Protocol` it comes from is more akin to the instructions for performing it.

zhang-ivy:
Thanks for the clarification! I still think it would be better to rename `Protocol` because if gufe is going to replace the perses API, "protocol" means something different in the context of free energy calculations, so its going to be confusing for perses users. However, it seems like I may be the only one that has a problem with the name, so I think its ok if we keep it as is and just make sure the class/name choice is well documented (which I'm sure you are planning to do anyway).

ijpulidos:
I don't have a strong opinion here, there are valid points from all sides. I guess that as long as we make this clear from the docstrings and existing documentation, this should be fine. We may even want to put something like "Not to be confused with... " as a first line in the docstring.

I wonder if aliasing the `gufe` classes names is a sane option for perses codebase. I don't really know enough about design patterns around this.

dotsdl:
> ...because if gufe is going to replace the perses API, "protocol" means something different in the context of free energy calculations, so its going to be confusing for perses users.

@zhang-ivy can you define here what "protocol" means in the context of free energy calculations? I don't know that there's a universally-accepted definition in the field, but if there is strong consensus I'd like to know what it is so we can act accordingly.

zhang-ivy:
When I hear "protocol" in the context of alchemical free energy calculations, I think of the alchemical protocol, which tells us how we should transform the WT (or lambda = 0 endstate) into the mutant (or lambda = 1 endstate) residue. In practice, this can be thought of as a list of lambda values. The protocols I've been using are linear (e.g. if there are 4 alchemical states, the protocol would be [0, 0.25, 0.5, 1.0] and 0 would be mean that the WT residue's nonbonded interactions are fully on and the mutant residue's nonbonded interactions are fully off, 0.25 would mean that the WT residue's nonbonded interactions are turned 75% on and the mutant nonbonded interactions are turned 25% on, and so on), but they can be non-linear with a varying number of alchemical states. 

The alchemical protocol used is critical to the calculation because it determines how well our states overlap, which in turn affects the bias and error of the calculation.

perses/protocols/nonequilibrium_cycling.py
```diff
@@ -11,22 +11,39 @@
...
+    Context,
+    execute
 )
+
 #from gufe.protocols.settings import ...
 
 
-class InitializeUnit(ProtocolUnit):
-    def _execute(self, dependency_results):
+class GenerateHybridTopology(ProtocolUnit):
```
zhang-ivy, line 21: Just a small naming note -- I think you will want this to be called something like `GenerateHybridTopologyFactory` as  the class that will need to be generated is called `HybridTopologyFactory`: https://github.com/choderalab/perses/blob/main/perses/annihilation/relative.py#L17

zhang-ivy: And I think all of the variable names below that reference a "hybrid topology" should really be called "hybrid topology factory" or "htf"

ijpulidos: Thanks for the feedback, and yes, this are just place-holder names that were aimed at helping me get the structure of the `gufe` objects (protocols and units). The names themselves of these (namely the `GenerateHybridTopology`) won't end up being that, it was just to illustrate the example.

zhang-ivy: Note that the `HybridTopologyFactory` generates more than just a hybrid topology -- it also (importantly) generates a hybrid system, which is why I think its important to use `HybridTopologyFactory` instead of `HybridTopology` in this PR

jchodera: Perhaps we could call this a `HybridAlchemicalFactory`, since it's supposed to take the real systems and generate a hybrid alchemical system? We could do that now or wait until we refactor, unify, and put this in `openmmtools`.

dotsdl:
@ijpulidos can we sync up on this early this week? I'd like to be in a position to run this protocol locally, ideally by 11/08. Is there anything blocking this for you?

mikemhenry:
Added example notebook here [examples/NEQ_cycling_using_gufe_objects/NEQ_cycling_from_gufe_objs.ipynb](https://github.com/choderalab/perses/pull/1066/files#diff-0d224809db6f63815527c0731d8cdb1d92577c2516d1d982333fa5395ce35308) it is messy, but a start on the notebook that we will want to show people how to use the code in the PR.

perses/protocols/nonequilibrium_cycling.py
ijpulidos: This has been modified in the code. Resolving.

perses/protocols/nonequilibrium_cycling.py
```diff
@@ -0,0 +1,452 @@
...
+        for i in range(n_bootstraps):
+            subsample_indices_forward = np.random.choice(len(forward), forward)
+            subsample_indices_reverse = np.random.choice(len(reverse), reverse)
+            forward_chosen = forward[subsample_indices_forward]
+            reverse_chosen = reverse[subsample_indices_reverse]
+            dg, ddg = pymbar.bar.BAR(forward_chosen, reverse_chosen)
+            all_dgs.append(dg)
+            # all_ddgs.append(ddg)
+
+        return all_dgs
```
ijpulidos, line 385: @jchodera We need your input here. Especially, when computing the free energy uncertainty we are doing it via the computing the standard deviation of the bootstrapped DeltaG distribution. Does that seem most appropriate?

jchodera: For nonequilibrium cycling, the forward and reverse work values are paired since they come from the same trajectory. Each trajectory can be assumed to be the unit to bootstrap on. SO I'd change this to:
```python
        # Check to make sure forward and reverse work values match in length
        assert len(forward) == len(reverse), "Forward and reverse work values are not paired"
        ntraj = len(forward)
        for i in range(n_bootstraps):
            # Sample trajectory indices with replacement
            indices = np.random.choice(np.arange(ntraj), size=[ntraj], replace=True)
            # Compute and store BAR estimates
            dg, ddg = pymbar.bar.BAR(forward[indices], reverse[indices])
            all_dgs.append(dg)
```
Also, is it critical to return a list here, rather than a numpy array? Why not return a numpy array, especially since you know the number of bootstrap samples you will generate anyway.

jchodera:
@dotsdl @ijpulidos : Does the object model for free energy results support confidence intervals or bootstrap samples for free energy estimates, or only standard errors?

perses/protocols/nonequilibrium_cycling.py
jchodera: For `get_estimate`, there is no need to bootstrap---you should just return the MLE. Please replace with this:
```python
    def get_estimate(self):
        """
        Get a free energy estimate.

        Returns
        -------
        free_energy: float
            Free energy estimate in units of kT.
        """
        free_energy, error = pymbar.bar.BAR(forward_work, reverse_work)
        return free_energy
```

jchodera: Also, you don't want to bootstrap every time you call `get_uncertainty`. I'd cache that result after you compute it the first time, since it's expensive.

I'm also not quite sure why we split up the accessors for free energy and uncertainty. Wouldn't we want to compute them both at the same time, once, and then cache them?

ijpulidos: > Also, you don't want to bootstrap every time you call `get_uncertainty`. I'd cache that result after you compute it the first time, since it's expensive. You may also want to cache the results of `get_free_energy` since BAR incurs some computation time.

In this case we are using a cache via the `@lru_cache()` decorator in https://github.com/choderalab/perses/pull/1066/files#diff-2d23d1987b5a674a614a5af699d2b64c5befb5c1c2a0b487200749cf5569b042R360

jchodera: Great!

jchodera:
For **testing**:
* We want to run alchemical transformations on very simple small systems, such as a small molecule transformation in vacuum or a tiny solvent box.
* @irfan noted he has a series of simple benzene transformations he could share
* Failing that, we could go back to our simple transformations (e.g. mutation in alanine dipeptide, changing substituents of benzene)
* Ideally, we would carry out a null transformation in vacuum (or solvent if needed) where we transform the molecule into the same molecule, but we don't map all atoms. For example, fluorobenzene to fluorobenzene:
```
[c:1]1(F)[c:2][c:3][c:4][c:5][c:6]1>>[c:1]1[c:2][c:3][c:4](F)[c:5][c:6]1
```
image
* 20 independent trajectories of 1 cycle each should generate 20 forward/reverse work pairs, and this should be enough to estimate a free energy. 
* We can keep the cycling trajectories short: 100 ps/segment (where a segment is one of eqA, neqAB, eqB, neqBA), so 400 ps/cycle

For **production**:
* We likely want to recommend 100 independent trajectories of 1 cycle each, 1 ns/segment

ijpulidos:
> 
>     * @irfan noted he has a series of simple benzene transformations he could share
> 
The benzene transformations are in https://github.com/OpenFreeEnergy/openfe-benchmarks/blob/main/openfe_benchmarks/benzenes.py

perses/protocols/nonequilibrium_cycling.py
ijpulidos: Thanks so much for the improvements suggested. They have been addressed in commit https://github.com/choderalab/perses/pull/1066/commits/c17c7eb8c06bf4242cbc52b4a27ee8ccb4f417a6

perses/protocols/nonequilibrium_cycling.py
```diff
@@ -0,0 +1,655 @@
...
+    result_cls = NonEquilibriumCyclingProtocolResult
+
+    def __init__(self, settings: ProtocolSettings):
+        super().__init__(settings)
+
+    @classmethod
+    def _default_settings(cls) -> ProtocolSettings:
+        from perses.protocols import settings
+        non_eq_settings = settings.NonEqCyclingSettings()
+        return non_eq_settings
```
richardjgowers, line 610: @dotsdl @ijpulidos I think this is meant to be `gufe.Settings` (the big object that contains the `ProtocolSettings` rather than just the `ProtocolSettings`.  This way, if you're very lazy, you can just do `Protocol(settings=Protocol.default_settings())` rather than `Protocol(settings=gufe.Settings(protocol_settings=Protocol.default_settings()))`

ijpulidos: Yes, I think that makes sense. Good catch! Thanks.

ijpulidos: This has been implemented in https://github.com/choderalab/perses/pull/1066/commits/9044b3c268be7f469283f570da9fb2b31046a38b

dotsdl:
I'm fixing up the settings bits to match latest `gufe`.

dotsdl:
@ijpulidos it looks like the CI may be stuck? Also, running tests locally I now get:

```
> pytest --pdb perses/tests/protocols
============================================================================================= test session starts =============================================================================================
platform linux -- Python 3.9.16, pytest-7.2.2, pluggy-1.0.0
rootdir: /home/david/Library/choderalab/perses
plugins: anyio-3.6.2
collected 6 items

perses/tests/protocols/test_nonequilibrium_cycling.py E
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> captured log >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
INFO     utils.openeye:openeye.py:434 molecule benzene                      does not have unique atom names. Generating now...
INFO     utils.openeye:openeye.py:434 molecule toluene                      does not have unique atom names. Generating now...
INFO     relative_setup:relative_setup.py:167 			_hmass: 3 Da.

INFO     relative_setup:relative_setup.py:198 No receptor specified. Assuming non-complex phase.
INFO     relative_setup:relative_setup.py:209 	setting both molecule oemol titles to 'MOL'.
INFO     relative_setup:relative_setup.py:224 Created mdtraj topologies for both ligands.
INFO     relative_setup:relative_setup.py:492 Setting non bonded method to PME
INFO     relative_setup:relative_setup.py:251 set MonteCarloBarostat because pressure was specified as 1.0 atm atmospheres
INFO     relative_setup:relative_setup.py:288 PME tolerance: 0.00025
INFO     relative_setup:relative_setup.py:313 successfully created SystemGenerator to create ligand systems
INFO     relative_setup:relative_setup.py:315 executing SmallMoleculeSetProposalEngine...
INFO     relative_setup:relative_setup.py:329 instantiating FFAllAngleGeometryEngine...
INFO     relative_setup:relative_setup.py:737 Detected solvent...
INFO     relative_setup:relative_setup.py:739 assgning noCutoff to nonbonded_method
INFO     relative_setup:relative_setup.py:743 Removing barostat for vacuum phase
INFO     relative_setup:relative_setup.py:745 Setting nonbondedMethod to NoCutoff for vacuum phase
INFO     relative_setup:relative_setup.py:747 calling SystemGenerator to create ligand systems.
INFO     relative_setup:relative_setup.py:750 No complex or solvent leg, so performing topology proposal for vacuum leg
INFO     relative_setup:relative_setup.py:1055 	Skipping solvation of vacuum perturbation
INFO     relative_setup:relative_setup.py:1086 	parameterizing...
INFO     openmmforcefields.generators.template_generators:template_generators.py:280 Requested to generate parameters for residue <Residue 0 (MOL) of chain 0>
INFO     openmmforcefields.generators.template_generators:template_generators.py:1402 Generating a residue template for [H]c1c(c(c(c(c1[H])[H])[H])[H])[H] using openff-2.0.0
INFO     relative_setup:relative_setup.py:1088 	System parameterized
INFO     relative_setup:relative_setup.py:1095 Both trajectory_directory and trajectory_prefix need to be provided to save .pdb
INFO     proposal_generator:topology_proposal.py:2053 creating current receptor topology by removing small molecule from current topology...
INFO     proposal_generator:topology_proposal.py:2060 small molecule start index: 0
INFO     proposal_generator:topology_proposal.py:2061 small molecule has 12 atoms.
INFO     proposal_generator:topology_proposal.py:2065 old alchemical atom indices: range(0, 12)
INFO     proposal_generator:topology_proposal.py:2076 proposed mol detected with smiles [H]c1c(c(c(c(c1[H])[H])C([H])([H])[H])[H])[H] and logp_proposal of 0.0
INFO     proposal_generator:topology_proposal.py:2079 conducting proposal from [H]c1c(c(c(c(c1[H])[H])[H])[H])[H] to [H]c1c(c(c(c(c1[H])[H])C([H])([H])[H])[H])[H]...
INFO     proposal_generator:topology_proposal.py:2082 building new topology with proposed molecule and current receptor topology...
INFO     proposal_generator:topology_proposal.py:2303 	setting proposed oemol title to MOL
INFO     proposal_generator:topology_proposal.py:2305 	creating mol topology from oemol...
INFO     proposal_generator:topology_proposal.py:2308 	appending current receptor topology to new mol topology...
INFO     proposal_generator:topology_proposal.py:2087 new molecule has a start index of 0 and 15 atoms.
INFO     proposal_generator:topology_proposal.py:2090 proceeding to build the new system from the new topology...
INFO     openmmforcefields.generators.template_generators:template_generators.py:280 Requested to generate parameters for residue <Residue 0 (MOL) of chain 1>
INFO     openmmforcefields.generators.template_generators:template_generators.py:1402 Generating a residue template for [H]c1c(c(c(c(c1[H])[H])C([H])([H])[H])[H])[H] using openff-2.0.0
INFO     proposal_generator:topology_proposal.py:2099 determining atom map between old and new molecules...
INFO     proposal_generator:topology_proposal.py:2127 atom map is pre-determined as {0: 4, 1: 5, 2: 6, 3: 7, 4: 8, 5: 9, 6: 10, 7: 11, 8: 12, 9: 13, 11: 14}
INFO     proposal_generator:topology_proposal.py:2175 Proposed transformation would delete 2 atoms and create 5 atoms.
INFO     relative_setup:relative_setup.py:771 conducting geometry proposal...
INFO     geometry:geometry.py:218 propose: performing forward proposal
INFO     geometry:geometry.py:235 propose: unique new atoms detected; proceeding to _logp_propose...
INFO     geometry:geometry.py:371 Conducting forward proposal...
INFO     geometry:geometry.py:380 Computing proposal order with NetworkX...
INFO     geometry:geometry.py:386 number of atoms to be placed: 5
INFO     geometry:geometry.py:387 Atom index proposal order is [14, 12, 3, 13, 10]
INFO     geometry:geometry.py:388 omitted_bonds: []
INFO     geometry:geometry.py:392 direction of proposal is forward; creating atoms_with_positions and new positions from old system/topology...
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> traceback >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

self = <perses.tests.protocols.test_nonequilibrium_cycling.TestNonEquilibriumCycling object at 0x7f7f19ab62b0>, protocol_short = <NonEquilibriumCyclingProtocol-5cacbf0b458e69c14f23ad4f16d18807>
benzene_vacuum_system = ChemicalSystem(name=, components={'ligand': SmallMoleculeComponent(name=benzene)})
toluene_vacuum_system = ChemicalSystem(name=, components={'ligand': SmallMoleculeComponent(name=toluene)})
mapping_benzene_toluene = LigandAtomMapping(componentA=SmallMoleculeComponent(name=benzene), componentB=SmallMoleculeComponent(name=toluene), componentA_to_componentB={4: 0, 5: 1, 6: 2, 7: 3, 8: 4, 9: 5, 10: 6, 11: 7, 12: 8, 13: 9, 14: 11}, annotations={})
tmpdir = local('/tmp/pytest-of-david/pytest-5/test_dag_execute0')

    @pytest.fixture
    def protocol_dag_result(self, protocol_short, benzene_vacuum_system, toluene_vacuum_system,
                            mapping_benzene_toluene, tmpdir):
        dag = protocol_short.create(
            stateA=benzene_vacuum_system, stateB=toluene_vacuum_system, name="Short vacuum transformation",
            mapping=mapping_benzene_toluene
        )

        with tmpdir.as_cwd():

            shared=Path('shared')
            shared.mkdir()

            scratch=Path('scratch')
            scratch.mkdir()

>           dagresult: ProtocolDAGResult = execute_DAG(dag, shared_basedir=shared, scratch_basedir=scratch)

perses/tests/protocols/test_nonequilibrium_cycling.py:37:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../../.conda/envs/alchemiscale-lib-docker-2023.04.05/lib/python3.9/site-packages/gufe/protocols/protocoldag.py:415: in execute_DAG
    result = unit.execute(
../../../.conda/envs/alchemiscale-lib-docker-2023.04.05/lib/python3.9/site-packages/gufe/protocols/protocolunit.py:286: in execute
    outputs = self._execute(context, **inputs)
perses/protocols/nonequilibrium_cycling.py:238: in _execute
    fe_setup = RelativeFEPSetup(
perses/app/relative_setup.py:349: in __init__
    self._setup_phases(
perses/app/relative_setup.py:513: in _setup_phases
    self._setup_vacuum_phase(atom_map=atom_map)
perses/app/relative_setup.py:776: in _setup_vacuum_phase
    self._vacuum_positions_new, self._vacuum_logp_proposal = self._geometry_engine.propose(
perses/rjmc/geometry.py:236: in propose
    logp_proposal, new_positions, rjmc_info, atoms_with_positions_reduced_potential, final_context_reduced_potential, neglected_angle_terms, omitted_terms = self._logp_propose(top_proposal, current_positions, beta, direction='forward', validate_energy_bookkeeping = validate_energy_bookkeeping)
perses/rjmc/geometry.py:397: in _logp_propose
    new_positions = self._copy_positions(atoms_with_positions, top_proposal, old_positions)
perses/rjmc/geometry.py:847: in _copy_positions
    new_positions[atom.idx] = current_positions[old_index]
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = Quantity(value=array([[2.5978    , 0.5327    , 0.4779    ],
       [2.6395001 , 0.5074    , 0.3499    ],
       [2.734...    [2.78209   , 0.80598   , 0.53863   ],
       [2.61874   , 0.65720004, 0.64958   ]], dtype=float32), unit=nanometer)
key = 12

    def __getitem__(self, key):
        """
        Keep the same units on contained elements.
        """
>       assert not is_quantity(self._value[key])
E       IndexError: index 12 is out of bounds for axis 0 with size 12

../../../.conda/envs/alchemiscale-lib-docker-2023.04.05/lib/python3.9/site-packages/openmm/unit/quantity.py:735: IndexError

```

dotsdl:
Thoughts on how to proceed?

ijpulidos:
@dotsdl We still haven't refactored the tests to handle the new `RelativeFEPSetup` class, I agree we should start working on this soon. Since, there are also merge conflicts I think that's why the CI is stuck.

ijpulidos:
@dotsdl Oh, looking at your error traceback the problem must be related with how we inverted the mapping in the protocol but we haven't done so in the test fixtures. Should be an easy fix.

mikemhenry:
Now that we are doing the refactor, we can merge this in once tests pass (feel free to remove tests that don't make sense anymore, but I do what the new stuff tested)

ijpulidos:
Tests will fail with this one, I don't really know why they aren't being run. But I expect we would need to refactor some of the tests since the API for the `RelativeFEPSetup` object has changed.

Does it make sense that we try to merge this one as it is (after we resolve the merging conflicts) and work on fixing/refactoring the tests for it in a separate PR? Such that we can review these changes independently and it doesn't get too overwhelming (this is already a huge PR).

ijpulidos:
Oh right, I think tests are not running because of the merge conflict, @mikemhenry already mentioned that in the past.

ijpulidos:
We could just mark the failing tests to be skiped as `old_api` or similar. And fix the tests when we work on the refactor.

ijpulidos:
The activity on this protocol has been migrated to https://github.com/choderalab/feflow. Closing.

Issue #1079, opened 2022-08-02
Support python 3.10


ijpulidos:
Now that `openeye-toolkits` seems to be supporting python 3.10 we should add this version to our config matrix in the CI workflow.

mikemhenry:
Fixed in https://github.com/choderalab/perses/pull/1080

PR #1080, opened 2022-08-02
add python 3.10 to CI

mikemhenry:
## Motivation and context

Resolves #???

## Change log

```

```

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -23,7 +23,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8, 3.9]
+        python-version: [3.8, 3.9, "3.10"]
         openmm: ["nightly", "7.6", "7.7"]
 
     env:
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -25,6 +25,9 @@ jobs:
       matrix:
         python-version: [3.8, 3.9, "3.10"]
         openmm: ["nightly", "7.6", "7.7"]
+        exclude:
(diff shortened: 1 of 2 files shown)


mikemhenry:
Resolves #1079

mikemhenry:
@ijpulidos do we want to drop 3.8 from the matrix?

ijpulidos:
> @ijpulidos do we want to drop 3.8 from the matrix?

That's a good question, I know this was never official but I think we are somewhat following [Numpy's deprecation policy](https://numpy.org/neps/nep-0029-deprecation_policy.html). According to that we should be supporting 3.8+ up to Jan 2023, as far as I understand.

ijpulidos:
Seems like we would need to drop support for Openmm 7.6 if we want to add support to python 3.10, judging from https://github.com/choderalab/perses/runs/7641867028?check_suite_focus=true#step:5:53

mikemhenry:
I was thinking that I would just remove that from the matrix since I don't think we need to stop testing on openmm 7.6, unless there is no scope to use that version with perses. 

@zhang-ivy @dominicrufa 

Do either of you use openmm 7.6?

zhang-ivy:
Nope, I don't use it!

mikemhenry:
@ijpulidos ready for review!

zhang-ivy:
@mikemhenry @ijpulidos : Do we need to add 3.10 to CI for openmmtools as well?

mikemhenry:
@zhang-ivy good idea https://github.com/choderalab/openmmtools/pull/619

ijpulidos (approved):
Looks good to me! LGTM.

PR #1087, opened 2022-08-19
add docker files

mikemhenry:
## Description

These are the docker files I've been using to build images. 

## Motivation and context

I didn't want to be a single point of failure if we needed to re-build some images, I've just kept these local.

## How has this been tested?

@ijpulidos :)

## Change log

I don't think this warrants a change log entry.


mikemhenry:
How about we keep the two you described, I add a readme on how to build them, and then make a folder that keeps all of the extras if we need them?

jchodera (approved):
I think it's best to merge these if possible. Even if we don't use them all, they're good examples.

ijpulidos (approved):
This should be good to merge.

PR #1091, opened 2022-08-22
Cleaning up platform and context usage in geometry engine

### LINKED ISSUE #1085: CUDA platform hardcoded in geometry engine
When playing with different tests and platforms I realized that perses currently *requires* a CUDA platform to work. Specifically,  `FFAllAngleGeometryEngine._log_propose` method has the platform hardcoded in https://github.com/choderalab/perses/blob/87bc1d00c6c087b0dc13b38f83f5e2c4d13e012d/perses/rjmc/geometry.py#L453

My guess is that this is a bug and we actually want to allow different platforms to be used in this case. Uless there's a real reason why CUDA is required for this method.

ijpulidos:
## Description
By default the `CUDA` platform was hardcoded in the geometry engine. Generating contexts that were not handled correctly. This PR is aimed towards making `CPU` to be the default platform for the geometry engine and cleans explicitly created contexts for a better memory management.

## Motivation and context

Resolves #1085 

## Change log

```
``CUDA`` platform was hardcoded in geometry engine, generating performance issues by not clearing openmm contexts correctly. Fixed by defaulting to using the faster ``CPU`` platform (for the geometry engine) and explicitly deleting context variables after they are used.
```

### DIFF
--- a/perses/rjmc/geometry.py
+++ b/perses/rjmc/geometry.py
@@ -732,6 +732,9 @@ class FFAllAngleGeometryEngine(GeometryEngine):
         added_energy_components = compute_potential_components(mod_context)
         print(f"added energy components: {added_energy_components}")
 
+        # Explicitly clean up context memory allocation
+        del mod_context
+
         return modified_reduced_potential_energy
 
 
--- a/perses/rjmc/geometry.py
+++ b/perses/rjmc/geometry.py
@@ -325,7 +325,8 @@ class FFAllAngleGeometryEngine(GeometryEngine):
         pdbfile.flush()
         pdbfile.write('ENDMDL\n')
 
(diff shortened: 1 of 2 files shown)


mikemhenry (approved):

mikemhenry:
Not sure why RTD isn't working, builds have been working just fine: https://readthedocs.org/projects/perses/builds/
I will merge this in any way and we can check on RTD later.

perses/rjmc/geometry.py
```diff
@@ -733,6 +732,9 @@ def _corrected_reduced_potential(self, growth_system_generator, positions, platf
         added_energy_components = compute_potential_components(mod_context)
```
zhang-ivy, line 732: Instead of changing the `DEFAULT_PLATFORM` in tests/utils.py, we want to change this line to 

```
added_energy_components = compute_potential_components(mod_context, platform=platform)
```

PR #1095, opened 2022-08-29
Feat/fix ssl gpu error

mikemhenry:
## Motivation and context

Resolves #???

## Change log

```

```

### DIFF
--- a/.github/workflows/self-hosted-gpu-test.yml
+++ b/.github/workflows/self-hosted-gpu-test.yml
@@ -55,17 +55,19 @@ jobs:
         shell: bash -l {0}
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


mikemhenry:
before merging this we will need to change the yaml to run only on merge and not on every commit to the PR

mikemhenry:
Updating the image didn't fix it, but I will keep that as-is since it shouldn't hurt things.

mikemhenry:
@ijpulidos this is ready for review! I removed the bit where I run the GPU tests on commits to this PR, but you can see that in the previous run we have a green check and that includes the GPU tests

ijpulidos (approved):
This looks good to me, LGTM!

PR #1099, opened 2022-09-01
Pin to older openff-toolkit until we get openmmtools fixed

mikemhenry:
## Motivation and context

Resolves #???

## Change log

```

```


mikemhenry:
@ijpulidos I'm going to get this merged in, it trims down testing deps + pins to the older openff-toolkit until we get openmmtools sorted.

PR #1100, opened 2022-09-06
Accept multiple ligand input files for `RelativeFEPSetup`

### LINKED ISSUE #1093: `RelativeFEPSetup` cannot deal with list of ligand input files
In PR #1073 we included a way to deal with files from different cloud providers using `cloudpathlib`. But when doing this we generated a bug where if we pass a list of filenames for the `RelativeFEPSetup` object as the `ligand_input` parameter, we end up with an error like the following:
```python
  File "/home/user/workdir/debugging/perses/perses-new-issue/test_RelativeFEPSetup_API.py", line 19, in <module>
    fe_setup = RelativeFEPSetup(
  File "/home/user/miniconda3/envs/perses-1092-omm-nightly-test/lib/python3.9/site-packages/perses/app/relative_setup.py", line 165, in __init__
    self._ligand_input = AnyPath(ligand_input)
  File "/home/user/miniconda3/envs/perses-1092-omm-nightly-test/lib/python3.9/site-packages/cloudpathlib/anypath.py", line 38, in __new__
    raise AnyPathTypeError(
cloudpathlib.exceptions.AnyPathTypeError: Invalid input for both CloudPath and Path. CloudPath exception: InvalidPrefixError("Path ['./a1.sybyl.mol2', './a2.sybyl.mol2'] does not begin with a known prefix ['az://', 's3://', 'gs://'].") Path exception: TypeError('expected str, bytes or os.PathLike object, not list')
(perses-1092-omm-nightly-test) 
```
This is caused by trying to convert the list to an `cloudpathlib.AnyPath` object [here](https://github.com/choderalab/perses/blob/c96e65572b26cfdbce2d3ecaec7dae708e01568c/perses/app/relative_setup.py#L164-L165).

A quick fix would be to just convert the given `input_ligand` (or even `self._input_ligand`) to an `AnyPath` object after we check it isn't a list in [line 242](https://github.com/choderalab/perses/blob/c96e65572b26cfdbce2d3ecaec7dae708e01568c/perses/app/relative_setup.py#L242). 

A longer term solution would be to refactor how we are dealing with this and have special methods for each case (list or str path) instead of having the huge if-else blocks we currently have.

ijpulidos:
## Description

`RelativeFEPSetup` now checks wether the `ligand_input` parameters is a list or a string and handles both cases accordingly. This bug was introduced when we changed the behavior of reading files in #1073 .

Added test to check for this behavior.

## Motivation and context

Resolves #1093 

## How has this been tested?

Wrote test that checks the length of the `._ligand_input` private attribute for the `RelativeFEPSetup` object.

## Change log

There is no need for a changelog entry here, since the conflicting changes never made it to a release. That is, we are fixing the conflicting changes in the same release.

```

```


mikemhenry:
`E       openmm.OpenMMException: NonbondedForce: The cutoff distance cannot be greater than half the periodic box size.`
Looks like a padding issue

ijpulidos:
> `E openmm.OpenMMException: NonbondedForce: The cutoff distance cannot be greater than half the periodic box size.` Looks like a padding issue

Yes, it is likely that we need to increase the padding here, I just have to double check with previous tests (for example those in #1065 ), because I think I run those successfully with the openmm nightly build at the time.

mikemhenry:
This is the minimal amount of padding to get the test to pass.

mikemhenry (approved):

Issue #1078, opened 2022-07-28
Incorrect symmetry in mapping of TNSK2 ligand pair?


IAlibay:
Hi,

I'm running through some of the new ligands from https://github.com/openforcefield/protein-ligand-benchmark/pull/52 and I noticed a rather odd mapping for at least one of the edges.

For the transformation of `lig_5d` to `lig_5p` of the TNSK2 set ([edge.zip](https://github.com/choderalab/perses/files/9209410/edge.zip)), it appears that the default mapping generated by `perses.rjmc.atom_mapping.AtomMapperfor` has the wrong symmetry for the central ring, and the tert-butanol?

Here's the 2D mapping:
![atom_mapping](https://user-images.githubusercontent.com/12460125/181504362-48a91293-9963-486f-b65b-3f0a1d2d9d57.png)

And a 3D render of the mapping (7 A x direction translation in positions, matching sphere colors indicate a set of mapped atoms):
<img width="500" alt="image" src="https://user-images.githubusercontent.com/12460125/181504569-d9448874-b260-4a73-a15f-af377c36a955.png">

Here's a gist notebook with the code details: https://gist.github.com/IAlibay/089f614665ab3a10d076990ba8cd8024

Relevant library versions:
  - perses version 0.10.1
  - openeye-toolkits 2022.1.1
  - openff-tookit 0.10.6

jchodera:
@ialibay: Thanks for catching this! A few things could be going on here:
1. We may not be enumerating all symmetry-equivalent valid atom mappings, so only the flipped middle ring mapping get enumerated
2. We may not be using a small enough threshold for atom correspondence such that the flipped middle ring mapping is permitted as a geometrically valid mapping
3. I can't recall if we are scoring by geometry or simply filtering by it. I seem to recall we are only _filtering_ by it and then scoring by number of matched atoms. If so, we may need to score by RMS deviation of mapped atoms as well.

I think the key to debugging this will be to have the atom mapper print out all _valid_ mappings and the scores for these mappings, and then see which of the above cases might be going on here. That will help us identify how to fix it.

If it can't easily be fixed, we may need to rethink the strategy of enumerating all valid mappings via graph matchings and then filtering/scoring them by geometry.

@IAlibay : This 3D visualization is beautiful! It would be fantastic if this ended up in a module that we could use to automatically generate 3D atom mapping visualizations one could explore interactively or non-interactively.

IAlibay:
@jchodera 

[...]

PR #1103, opened 2022-09-09
set aromatic draw style to `OEAromaticStyle_Circle` in atom mapper

### LINKED ISSUE #1082: Change aromaticity drawing style
https://github.com/choderalab/perses/blob/87bc1d00c6c087b0dc13b38f83f5e2c4d13e012d/perses/rjmc/atom_mapping.py#L283-L286

https://docs.eyesopen.com/toolkits/python/depicttk/OEDepictClasses/OE2DMolDisplayOptions.html#OEDepict::OE2DMolDisplayOptions::SetAromaticStyle

mikemhenry:
## Motivation and context

Resolves #1082 

## Change log

```
Set aromatic draw style to `OEAromaticStyle_Circle` in atom mapper rendering
```


ijpulidos:
Thanks! I was trying to see how the visualization changes using the code in https://gist.github.com/IAlibay/089f614665ab3a10d076990ba8cd8024 from issue #1078 , but I don't really see a difference. What do we expect to be different here? As far as I remember that notebook/issue was the cause of this feature request.

mikemhenry:
I think our answer is somewhere here:
https://github.com/choderalab/perses/blob/main/perses/rjmc/atom_mapping.py#L305-L310

Is calling `oedepict.OERenderMolecule(oeimage, display)` enough to get all the `opts` we configured, or do we need to pass a return value to another function?

jchodera (approved):
This looks good to me!

jchodera:
This failure looks unrelated, so merging.

PR #1104, opened 2022-09-09
document setting ionic strengh in example

### LINKED ISSUE #1089: Make sure ionic_strength is available as an option in CLI, and update example to specify this
We want to make sure that users of the new CLI can specify `ionic_strength` and make sure the new CLI example we have uses this explicitly.

mikemhenry:
## Motivation and context

Resolves #1089 

## Change log

```
Document setting `ionic_strength` in `examples/new-cli/template.yaml`
```


ijpulidos (approved):
LGTM!

PR #1105, opened 2022-09-12
Speed up RTD env generation

mikemhenry:
## Motivation and context

Resolves #???

## Change log

```

```


mikemhenry:
This should speed up our RTD builds!

mikemhenry:
I was surprised this didn't work since it is the same as it is here? https://github.com/choderalab/openmmtools/blob/main/readthedocs.yml maybe I am missing something @ijpulidos

mikemhenry:
@ijpulidos should be good to go now!

ijpulidos (approved):
LGTM!

PR #1125, opened 2022-11-08
Handle old constrained to new unconstrained transformation, and vice versa

### LINKED ISSUE #1112: Constrained C-H bond mapped onto unconstrained C-F bond
I just came across this mapping from ligand 0 to 4. The bond between atoms 10 and 22 (constrained C-H) is mapped onto an unconstrained C-F, which should not happen.
image

Receptor PDB, ligands SDF, and template are attached.
[bug.zip](https://github.com/choderalab/perses/files/9699669/bug.zip)

ijpulidos:
## Description

This handles demapping old constrained atoms that become unconstrained in the new ligand, or vice versa.

This does it by directly checking if the atom changes the constrained state during the transformation and deletes de index from the mapping.

This also handles rendering the atom mapping correctly after correction.

## Motivation and context

Resolves #1112 

## Change log

```
Ligands with atoms changing constrained status were not being handled by mapping proposal. Atoms in bonds that change constrained/unconstrained to unconstrained/constrained during the transformation are now _demapped_.
```


perses/rjmc/topology_proposal.py
```diff
@@ -2344,6 +2350,8 @@ def _constraint_repairs(atom_map, old_system, new_system, old_topology, new_topo
         """
         Given an adjusted atom map (corresponding to the true indices of the new: old atoms in their respective systems), iterate through all of the
         atoms in the map that are hydrogen and check if the constraint length changes; if so, we do not map.
+
+        Returns the new atom map and also the deleted atom indices.
```
jchodera, line 2354: I know you didn't write the original docstring, but whenever we encounter a non-compliant docstring and have to work to figure out what the arguments are, we should fix the docstring.

The docstring should clearly document parameters and returns.

I'm not quite sure if the `{old|new}_{system|topology}` refer to the ENTIRE system (e.g. with solvent and protein) or one just representing the alchemical region---do you know?

It's also weird to me that we are passing a dict atom map, rather than an `AtomMap` object---we probably need to refactor that in future.

```
Parameters:
------------
atom_map : dict of int : int
    The atom mapping to repair; will not be modified.
    atom_map[old_index] refers to the new index of the corresponding atom in the new system
old_system : openmm.System
    The old OpenMM System containing parameters
new_system : openmm.System
    The old OpenMM System containing parameters
old_topology : openmm.app.Topology
    Old Topology
new_topology : openmm.app.Topology
    New Topology

Returns
--------
atom_map : dict of int : int 
    New atom map where some mapped atoms have been removed
deleted_new_atoms : list of int
    List of new_system atoms that were removed from the mapping
```
Also, why not return the deleted `old -> new` mappings as a `dict` rather than just the deleted new atoms?

perses/rjmc/topology_proposal.py
```diff
@@ -2366,18 +2374,23 @@ def _constraint_repairs(atom_map, old_system, new_system, old_topology, new_topo
             elif atom2 in new_hydrogens:
                 new_constraints[atom2] = length
 
-         #iterate through the atom indices in the new_to_old map, check bonds for pairs, and remove appropriate matches
+        # iterate through the atom indices in the new_to_old map, check bonds for pairs, and remove appropriate matches
         to_delete = []
```
jchodera, line 2378: Suggest we instead return the old -> new atom mappings that were removed from the `atom_map`, or potentially even return an `AtomMap` object that contains only the removed mapped atoms.

perses/rjmc/topology_proposal.py
```diff
@@ -2366,18 +2374,23 @@ def _constraint_repairs(atom_map, old_system, new_system, old_topology, new_topo
...
+            elif old_index in old_constraints.keys() and new_index not in new_constraints.keys():
+                to_delete.append(new_index)
+            elif old_index not in old_constraints.keys() and new_index in new_constraints.keys():
+                to_delete.append(new_index)
 
         for idx in to_delete:
             del atom_map[idx]
 
-        return atom_map
+        return atom_map, to_delete
```
jchodera, line 2393: It may not even be necessary to return the `to_delete` if we can just take the difference in atom maps.

perses/tests/test_topology_proposal.py
```diff
@@ -72,6 +74,57 @@ def test_small_molecule_proposals():
...
+def test_small_molecule_constraint_repair_mapping():
+    """
+    Tests that a constrained edge is demapped after transformed to an unconstrained one.
+
+    This is done by checking that the new_to_old atom map doesn't have the atom that otherwise would be included.
+
+    For the ligands in the test this atom is the ONLY Fluorine atom involved in the transformation.
+    """
+    # TODO: we could try testing a pyridine to a fluoropyridine transformation with using the given geometries
+    from perses.utils.openeye import createOEMolFromSDF
```
jchodera, line 87: This could be simplified to
```
# Load molecules
from openff.toolkit.topology import Molecule
off_molecules = Molecule.from_files(sdf_path)
# Change molecule names to `MOL`
[molecule.name = 'MOL' for molecule in off_molecules]
```   
This should get rid of the need for OpenEye molecules entirely (likes 93-105)

perses/tests/test_topology_proposal.py
```diff
@@ -72,6 +74,57 @@ def test_small_molecule_proposals():
...
+    system_generator = SystemGenerator(forcefields=forcefield_files, barostat=barostat,
+                                       forcefield_kwargs=forcefield_kwargs,
+                                       nonperiodic_forcefield_kwargs=nonperiodic_forcefield_kwargs,
+                                       small_molecule_forcefield=small_molecule_forcefield, molecules=off_molecules,
+                                       cache=None)
+    # Propose topology
+    proposal_engine = SmallMoleculeSetProposalEngine(molecules, system_generator)
+    proposal_engine.use_given_geometries = True  # Force to use the given geometries in the propose
+    proposal_engine.given_geometries_tolerance = 0.4 * unit.angstrom
+    initial_system, initial_positions, initial_topology = OEMol_to_omm_ff(molecules[0], system_generator)
```
jchodera, line 116: You can use `off_molecules[0].to_openeye()` instead of carrying around the OpenEye `OEMol` here:
```python
    initial_system, initial_positions, initial_topology = OEMol_to_omm_ff(off_molecules[0].to_openeye(), system_generator)
```

perses/tests/test_topology_proposal.py
```diff
@@ -72,6 +74,57 @@ def test_small_molecule_proposals():
...
+    # NOTE: We need to run the propose method for constraints repairs to happen
+    proposal = proposal_engine.propose(initial_system, initial_topology)
+    # Find index for the fluorine atom is in the new molecule
+    for atom_idx, oe_atom in enumerate(list(ligand_b.GetAtoms())):
```
jchodera, line 120: This could be
```python
atoms = { atom.name : atom_index for atom_index, atom in enumerate(off_molecules[1].atoms) }
fluorine_index = atoms['F1']
```

perses/rjmc/topology_proposal.py
```diff
@@ -2144,12 +2144,18 @@ def propose(self,
 
         # Explicitly de-map any atoms involved in constrained bonds that undergo length changes
         _logger.debug(f"mapped {len(adjusted_atom_map)} before constraint repairs")
-        adjusted_atom_map = SmallMoleculeSetProposalEngine._constraint_repairs(adjusted_atom_map, current_system, new_system, current_topology, new_topology)
+        adjusted_atom_map, deleted_indices = SmallMoleculeSetProposalEngine._constraint_repairs(adjusted_atom_map, current_system, new_system, current_topology, new_topology)
         _logger.debug(f"mapped {len(adjusted_atom_map)} after constraint repairs")
+        # Also adjust small molecule atom map accordingly to deleted indices
```
jchodera, line 2149: When we refactor this to use `AtomMap` objects, we shouldn't need to keep these extra utility attributes around---we can just use the repaired `AtomMap` and call `map.render_image(filename)`. But we can punt that to a future cleanup if we capture the cleanup task as an issue.

jchodera (changes requested):
Thanks for putting this together! I've left some feedback that boils down to:
1. Let's capture an issue to clean up the code here to use and store `AtomMap` objects instead of int:int `dict` objects in perses 0.11.0
2. You can streamline the test using OpenFF `Molecule` objects as I show
3. We need to document the parameters/returns---I've tried to suggest how we might do this

perses/rjmc/topology_proposal.py
ijpulidos: I guess we could convert them to sets and get the difference, new - old, but I'm not 100% sure this is safe in the long term, especially if we want to use `AtomMapping` objects instead of dictionaries or sets.

perses/rjmc/topology_proposal.py
ijpulidos: Yes I agree, but we would still need some way to access the inner object, either to some attribute or property. But definitely, we should try to refactor this to use `AtomMapping`.

perses/tests/test_topology_proposal.py
ijpulidos: I don't really understand why, but when I try that I get the following traceback https://gist.github.com/ijpulidos/c192994384f61a1843111105db332cfa This was the reason that I decided to use OE molecules, because I did want this to use off molecules in the first place.

ijpulidos: My guess is that there must be something getting lost when converting from openff's `Molecule` to `OEMol`

perses/tests/test_topology_proposal.py
ijpulidos: Using openff `Molecule`s can make it cleaner here, it would be a bit more intricate than that, since the aton names are not filled automatically when creating openff `Molecule` objects. We instead could use the `atom.element.symbol` or `atom.element.name` attributes, for `openff-toolkit` < 0.11. For >=0.11 we could use `atom.symbol` directly.

perses/rjmc/topology_proposal.py
ijpulidos: should be addressed now in commit https://github.com/choderalab/perses/pull/1125/commits/4f6b782b03c0e20d07c34a0c8c3d5176da7076fb

perses/rjmc/topology_proposal.py
ijpulidos: Closing. Issue created in https://github.com/choderalab/perses/issues/1129

perses/rjmc/topology_proposal.py
ijpulidos: Issue created in https://github.com/choderalab/perses/issues/1129 . Resolving conversation.

perses/tests/test_topology_proposal.py
ijpulidos: So far `OEMol` is still needed, as per https://github.com/choderalab/perses/pull/1125#discussion_r1017131327

perses/tests/test_topology_proposal.py
jchodera: Yikes! That sounds like something we should chase down separately. I'll look into this.

perses/tests/test_topology_proposal.py
jchodera: > Using openff Molecules can make it cleaner here, it would be a bit more intricate than that, since the atom names are not filled automatically when creating openff Molecule objects

I totally forgot that SDF files don't contain atom names! Good point here.

jchodera (approved):
Thanks so much! I'll debug the OEMol/Molecule issue separately!

Issue #1127, opened 2022-11-08
Update to openff 0.11 toolkit


jchodera:
Updating to the openff 0.11 toolkit release causes some tests to fail because the atom mapper no longer works.

Here is a minimal reproducing example:
```
pytest -v perses/tests/test_atom_mapping.py::TestAtomMapper::test_atom_mappings_moonshot
```
end of traceback:
```
E               ValueError: Cannot construct openff.toolkit.topology.Molecule from <openeye.oechem.OEMol; proxy of <Swig Object of type 'OEMolWrapper *' at 0x2b63467e8150> >
E               No registered toolkits can provide the capability "from_object" for args "(<openeye.oechem.OEMol; proxy of <Swig Object of type 'OEMolWrapper *' at 0x2b63467e8150> >,)" and kwargs "{'allow_undefined_stereo': True, '_cls': <class 'openff.toolkit.topology.molecule.Molecule'>}"
E               Available toolkits are: [ToolkitWrapper around OpenEye Toolkit version 2022.1.1, ToolkitWrapper around The RDKit version 2022.09.1, ToolkitWrapper around AmberTools version 22.0, ToolkitWrapper around Built-in Toolkit version None]
E                ToolkitWrapper around OpenEye Toolkit version 2022.1.1 <class 'openff.toolkit.utils.exceptions.RadicalsNotSupportedError'> : The OpenFF Toolkit does not currently support parsing molecules with radicals. Found 1 radical electrons on molecule [c]1[c][c]c([c][c]1)[C][C][N]c2[c][c][c]n[c]2.
E                ToolkitWrapper around The RDKit version 2022.09.1 <class 'NotImplementedError'> : Cannot create Molecule from <class 'openeye.oechem.OEMol'> object

../../../../miniconda/envs/perses-dev/lib/python3.10/site-packages/openff/toolkit/topology/molecule.py:1007: ValueError
```
The fix will likely require reworking `perses/rjmc/atom_mapping.py` to avoid instantiating `openff.toolkit.topology.Molecule` from an OpenEye `OEMol` that represents a molecule substructure.

The root cause is that the openff-toolkit 0.11 makes it impossible to convert fragments of molecules between OpenFF Molecule, OEMol, and RDMol now due to this PR:
https://github.com/openforcefield/openff-toolkit/pull/1236

I’ve opened an issue here to discuss whether this regression can be rolled back:
https://github.com/openforcefield/openff-toolkit/issues/1457

PR #1128, opened 2022-11-08
Support openff-toolkit 0.11

jchodera:
## Description

This PR addresses #1127.

**THIS IS A WORK IN PROGRESS**

It's not yet clear whether we will remain compatible with older openff-toolkit versions, but we will attempt to if possible.

## Motivation and context

The openff-toolkit 0.11 included API-breaking changes, such as https://github.com/openforcefield/openff-toolkit/issues/1457

## How has this been tested?

- [ ] Implement tests

## Change log

```
* Updated to support openff-toolkit 0.11, which included API-breaking changes.
```


jchodera:
@ijpulidos : You'll want to just keep chipping away at the test failures with
```
 pytest -v -x perses/tests/test_atom_mapping.py
```
(which stops after one test failure)

ijpulidos:
Tests are passing, I still need to test that the ultimate purpose is achieved, which is running this in an environment with `openff-toolkit>=0.11` and maybe we want to double check that the mappings for our benchmarks (tyk2 or others) stay the same.

jchodera:
Woo! Thanks so much!

I bet you can just update the GitHub Actions workflow to see if it works with the latest openff-toolkit.

ijpulidos:
@jchodera We are actually already testing the latest `openff-toolkit` version (0.12.1 to date) in our CI https://github.com/choderalab/perses/actions/runs/4738937597/jobs/8414096774?pr=1128#step:7:276 

So this should be good to go!

ijpulidos:
Sadly, the openmm8 tests are still failing due to the subnormals treatment in the `openeye` package. But these fails are unrelated to the changes in this PR.

mikemhenry:
This openmm 7.7 testing, is it a flaky one? https://github.com/choderalab/perses/actions/runs/4738937597/jobs/8414096664?pr=1128#step:9:853

devtools/conda-envs/test_env.yaml
```diff
@@ -24,7 +24,7 @@ dependencies:
   - numexpr
   - numpy >=1.14
   - openeye-toolkits
-  - openff-toolkit <0.11.0
+  - openff-toolkit >=0.11.0 # updated to latest
```
mikemhenry, line 27: We will need to make sure we also do this on conda-forge

ijpulidos:
> This openmm 7.7 testing, is it a flaky one? https://github.com/choderalab/perses/actions/runs/4738937597/jobs/8414096664?pr=1128#step:9:853

Oh I did notice that as well, good question. I'm pretty sure that if we re-run it will pass, but you are correct, I wonder if this is related to the changes in this PR. Any ideas how to check this?

perses/tests/test_atom_mapping.py
```diff
@@ -21,8 +21,8 @@ class TestAtomMapping(unittest.TestCase):
     """Test AtomMapping object."""
     def setUp(self):
         """Create useful common objects for testing."""
-        self.old_mol = Molecule.from_smiles('[C:0]([H:1])([H:2])([H:3])[C:4]([H:5])([H:6])([H:7])') # ethane
-        self.new_mol = Molecule.from_smiles('[C:0]([H:1])([H:2])([H:3])[C:4]([H:5])([H:6])[O:7][H:8]') # ethanol
+        self.old_mol = Molecule.from_mapped_smiles('[C:1]([H:2])([H:3])([H:4])[C:5]([H:6])([H:7])([H:8])') # ethane
```
mikemhenry, line 24: So does the mapping index start at `1` instead of `0` now?

mikemhenry (approved):

mikemhenry:
> > This openmm 7.7 testing, is it a flaky one? https://github.com/choderalab/perses/actions/runs/4738937597/jobs/8414096664?pr=1128#step:9:853
> 
> Oh I did notice that as well, good question. I'm pretty sure that if we re-run it will pass, but you are correct, I wonder if this is related to the changes in this PR. Any ideas how to check this?

I will try re-running it, I think it is unlikely that this PR broke that one

perses/tests/test_atom_mapping.py
j-wags: Yup - It's 1-indexed! https://docs.openforcefield.org/projects/toolkit/en/stable/api/generated/openff.toolkit.topology.Molecule.html?highlight=from_mapped_smiles#openff.toolkit.topology.Molecule.from_mapped_smiles

I will note that `from_mapped_smiles` guarantees that the output molecule will have the atoms INDEXED in the map order, but the output molecule won't have anything in `mol.properties['atom_map']`. On the other hand, running boring old `from_smiles` with a mapped smiles doesn't guarantee anything about the output atom indexing, but does populate the appropriate mapping in `molecule.properties['atom_map']`.

For testing purposes (where you need atom 3 to be a hydrogen, for example) it's usually correct to use `from_mapped_smiles`. But since I know perses uses atom maps for other purposes, I wanted to double check that this test doesn't need anything in the `molecule.properties['atom_map']` dict :-)

ijpulidos: I did a search for `atom_map` being used as a key for anything in perses and I didn't find any matches. I think we should be safe that way. Thank you!

devtools/conda-envs/test_env.yaml
ijpulidos: I think this works with 0.10.7 as well, but we probably want to support 0.11 or higher either way, because the protocols with the gufe objects will need that once we merge that into main. Thanks!

PR #1133, opened 2022-11-22
Not reassigning velocities on resume for small molecule pipeline

### LINKED ISSUE #1114: Small pipeline workflow is reassigning velocities by default on resume
The default behavior for the small molecule pipeline is to reassign velocities when resuming, as shown in https://github.com/choderalab/perses/blob/2ff4a7bb631a93716b64f2bd11dc68a2864402cd/perses/app/setup_relative_calculation.py#L713

Since we now are able to resume velocities correctly, PR https://github.com/choderalab/openmmtools/pull/602 , shouldn't we just make the `reassign_velocities = False` instead? This should probably improve both performance and accuracy of the simulations, when resuming.

ijpulidos:
## Description

The small molecule pipeline (CLI) was reassigning velocities by default on resume. Changing this behavior to avoid equilibration issues when resuming simulations now that it is supported upstream https://github.com/choderalab/openmmtools/pull/602.

## Motivation and context

Resolves #1114 

## Change log

```
Now that upstream ``openmmtools`` is storing velocities on checkpoint, the small molecule transformation pipeline does not reassign velocities on resume. Instead, it is reading them from the checkpoint file.
```

### DIFF
--- a/perses/app/setup_relative_calculation.py
+++ b/perses/app/setup_relative_calculation.py
@@ -694,7 +694,7 @@ def run_setup(setup_options, serialize_systems=True, build_samplers=True):
                         timestep=timestep,
                         collision_rate=1.0 / unit.picosecond,
                         n_steps=n_steps_per_move_application,
-                        reassign_velocities=True,
+                        reassign_velocities=False,
                         n_restart_attempts=20, constraint_tolerance=1e-06),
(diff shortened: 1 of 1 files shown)


mikemhenry:
So this just changes the default? Looks okay, can you check to see why pint is missing?

mikemhenry:
Looks like it is failing on scheduled CI, so not an issue for this PR

mikemhenry (approved):

PR #1138, opened 2023-01-06
Using the faster `LangevinDynamicsMove`

### LINKED ISSUE #1137: CLI repex should use `LangevinDynamicsMove`
It appears that the perses CLI uses [`LangevinSplittingDynamicsMove`](https://github.com/choderalab/perses/blob/1dc9b869bfb67963736bea23e5a87d26a28fa3dc/perses/app/setup_relative_calculation.py#L709) for hybrid repex, rather than the much faster [`LangevinDynamicsMove`](https://github.com/choderalab/openmmtools/blob/main/openmmtools/mcmc.py#L1018). We should change this to use `LangevinDynamicsMove`.

ijpulidos:
## Description

We should be using the faster [`LangevinDynamicsMove`](https://github.com/choderalab/openmmtools/blob/2e64980d86cf087f69e282927eebf2b41b73a3ea/openmmtools/mcmc.py#L1018) by default in our CLI workflow.

Tests and test systems were also updated to reflect the changes.

## Motivation and context

Increase performance by using the faster `LangevinMiddleIntegrator`.

Resolves #1137 

## Change log

```
CLI workflow for replica exchange now uses the faster ``LangevinMiddleIntegrator`` via the ``LangevinDynamicsMove``. Tests were updated to reflect the changes. 
```


jchodera (approved):
Thanks!

ijpulidos:
I just wanted to make sure that the new integrator wouldn't change anything in our benchmarks, just me being cautious. We get the same results with our `tyk2` benchmark (this one hasn't fully run the 5ns that's why it has somewhat big error bars, but it should be okay)

image

In terms of the performance gain, I was able to obtain about a 20% increase for the solvent phase and about a 9% increase for the complex phase.

PR #1142, opened 2023-01-13
Remove basicConfig

dwhswenson:
## Description

Removed `basicConfig` from being run on import, which causes problems for users, including one that is actually causing Python processes to crash for some users on `import perses.rjmc`.

* 7efc92931d9a9c64677a7872767baeb7e6b3583d : This is the minimal fix for the issue we're having
* d95a5fd30bfff2f4a90e5771382de339feb06af4 : I went ahead and removed similar issues from the rest of the library. I did not change things that looked like a script. For the CLI and anything with a `__main__` guard, I moved the `basicConfig` to only run on execution.

One I didn't change is in `tests/test_geometry_engine.py`. I'm not sure why that logging setting is there, but that looked like a Chesterton's fence.

I commented out the instance in `utils/url_utils.py` because maybe someone means to make that into a proper formatted version?

Places that still have `logging.basicConfig`:

```
benchmarks/run_benchmarks.py:22:logging.basicConfig(
perses/app/cli.py:91:    logging.basicConfig(
perses/app/setup_relative_calculation.py:1194:    logging.basicConfig(
perses/tests/test_geometry_engine.py:1012:    logging.basicConfig(level=logging.INFO)
perses/utils/url_utils.py:13:# logging.basicConfig(
```

## Motivation and context

Importing `perses.rjmc` in a fresh environment leads to a bunch of numba debug output. This seems to actually crash the Python process for some users (I haven't been able to reproduce, but have had 2 independent reports). This debug dump occurs on compile of numba functions, so after initial import, the compiled versions in `__pycache__` mean you don't get it again.

There are several places in `perses` with the line `logging.basicConfig(level=logging.NOTSET)`. This is the cause of the debug output, and I do not think it is the intended behavior:

1. The value of `logging.NOTSET` is `0`, meaning it gives the most verbose logging (more than `logging.DEBUG`). So what you probably want is `logging.basicConfig()`, taking the default behavior (which is `logging.WARNING`). This is why we're getting drowned in numba logging on first import. Doing this in `basicConfig` means that it is set for the root logger, so it affects all packages.

```python
In [1]: import logging
   ...: logger = logging.getLogger("foo")

In [2]: logger.debug("Speak into the void")

In [3]: logging.basicConfig(level=logging.NOTSET)
   ...: logger.debug("Now it's noisy")
DEBUG:foo:Now it's noisy
```

2. There is no need to call `logging.basicConfig` multiple times during import. Only the first call actually does anything; if there's already a handler attached to the root logger, the `basicConfig` doesn't do anything. In general, you probably only want it at the application (user-facing) level, not at the library (developer-facing) level. And you probably don't want it to be part of the imports, because you get things like this, which is not what a user would expect:

```python
In [1]: import perses.rjmc
   ...: import logging
WARNING:root:Warning: importing 'simtk.openmm' is deprecated.  Import 'openmm' instead.

In [2]: logging.basicConfig(level=logging.INFO)
   ...: logging.debug("this should not print, right?")
DEBUG:root:this should not print, right?
```

## How has this been tested?

Waiting on your CI? :laughing: 

## Change log

```
Removed use of `logging.basicConfig`, which was leading to excessive numba debug output on first import.
```

### DIFF
--- a/perses/analysis/visualization.py
+++ b/perses/analysis/visualization.py
@@ -50,7 +50,6 @@ def _check_mpy():
 ################################################################################
 
 import logging
-logging.basicConfig(level=logging.NOTSET)
 
 _logger = logging.getLogger("visualization")
 _logger.setLevel(logging.INFO)
--- a/perses/annihilation/lambda_protocol.py
+++ b/perses/annihilation/lambda_protocol.py
@@ -4,7 +4,6 @@ import logging
 import copy
 from openmmtools.alchemy import AlchemicalState
 
-logging.basicConfig(level=logging.NOTSET)
 _logger = logging.getLogger("lambda_protocol")
 _logger.setLevel(logging.INFO)
 
--- a/perses/annihilation/relative.py
+++ b/perses/annihilation/relative.py
@@ -9,7 +9,6 @@ InteractionGroup = enum.Enum("InteractionGroup", ['unique_old', 'unique_new', 'c
 
 #######LOGGING#############################
 import logging
-logging.basicConfig(level = logging.NOTSET)
 _logger = logging.getLogger("relative")
 _logger.setLevel(logging.INFO)
 ###########################################
--- a/perses/annihilation/rest.py
+++ b/perses/annihilation/rest.py
@@ -11,7 +11,6 @@ from perses.annihilation.relative import HybridTopologyFactory
 
 #######LOGGING#############################
 import logging
-logging.basicConfig(level = logging.NOTSET)
 _logger = logging.getLogger("REST")
 _logger.setLevel(logging.INFO)
 ###########################################
--- a/perses/app/cli.py
+++ b/perses/app/cli.py
@@ -9,11 +9,6 @@ import openmmtools.utils
 
 # Setting logging level config
 LOGLEVEL = os.environ.get("LOGLEVEL", "DEBUG").upper()
-logging.basicConfig(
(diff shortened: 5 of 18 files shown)


mikemhenry:
This is great @dwhswenson :tada:

dwhswenson:
looks like some CI issues currently... don't think that I broke that -- but I would recommend checking that this passes before merging. I could have done something stupid like leave off a `)`

ijpulidos:
@dwhswenson This is a great contribution, thank you for taking the time to fix this mess. 

In terms of the CI, I believe this is just an issue with your fork not having the OpenEye github secrets to be able to run them. I will just merge this one into another branch in perses repo and create a new PR with that one. I hope that works for you.

dwhswenson:
> In terms of the CI, I believe this is just an issue with your fork not having the OpenEye github secrets to be able to run them. I will just merge this one into another branch in perses repo and create a new PR with that one. I hope that works for you.

Yep, that works. Another thing to be aware of is the [`pull_request_target`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) trigger. This uses the workflows (therefore, safely uses the secrets) of the target branch. It was designed so that secrets could be used from forks. The downside is that, when updating workflows, you won't see the changes (since it runs the unchanged version).

ijpulidos:
Closing. Superseded by #1149

PR #1148, opened 2023-01-25
switch to using openmm rc

mikemhenry:
Just a quick and dirty way to get the rc check going, need to fix the automated check. Since Nightly hasn't been updated in a while (at least last time I checked) I think it is okay


jchodera:
Looks like this cut the test time down by a few minutes as well!

mikemhenry:
Taken care of by testing dev builds + our automatic RC checker

PR #1151, opened 2023-01-26
resolves issue #1150

### LINKED ISSUE #1150: Fix resume tests
Some of the tests in `test_resume.py` use the old and deprecated CLI. We need to fix them to use the new CLI interface.

`subprocess.run("perses-relative test.yml", shell=True, check=True, env=env)`

_Originally posted by @mikemhenry in https://github.com/choderalab/perses/issues/1065#issuecomment-1405304205_

mikemhenry:
## Description

Fix issue with test using old CLI

## Motivation and context

Resolves #1150 

## How has this been tested?

tested locally and on lilac, this test is skipped on CI

## Change log

```

```

### DIFF
--- a/perses/tests/test_resume.py
+++ b/perses/tests/test_resume.py
@@ -68,13 +68,13 @@ def test_cli_resume_repex():
             yaml.dump(y_doc, outfile)
 
         env = os.environ.copy()
-        subprocess.run("perses-relative test.yml", shell=True, check=True, env=env)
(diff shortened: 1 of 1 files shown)


ijpulidos (approved):
Thanks! We probably want to use the `click` capabilities to have isolated filesystems and using an instance of the `CliRunner` but this works for now. LGTM!

PR #1163, opened 2023-02-22
fix issue with logic, add debug statements

### LINKED ISSUE #1157: Serialized XML objects in small molecule pipeline have the wrong extension
When we serialize objects in https://github.com/choderalab/perses/blob/4e36a6b1f9f5588dc06c5eda2f3f037d1183f007/perses/app/setup_relative_calculation.py#L733-L735 

They are supposed to be gzipped files, but while inspecting these files one can easily tell they are directly the XML files. As in:

```
❯ file *
complex-hybrid-system.gz: XML 1.0 document, ASCII text
complex-new-system.gz:    XML 1.0 document, ASCII text
complex-old-system.gz:    XML 1.0 document, ASCII text
solvent-hybrid-system.gz: XML 1.0 document, ASCII text
solvent-new-system.gz:    XML 1.0 document, ASCII text
solvent-old-system.gz:    XML 1.0 document, ASCII text
```

Instead of the expected `"gzip compressed data"`.

mikemhenry:
## Description
There was a bug in the logic where if we saved a file with a `.gz` extension, we would save the file compressed, then overwrite the file with an uncompressed version.

## Motivation and context

Since this commit: 1145a7405f4beb0de10ab80aa80f6e01aa1a8794 we have not actually been compressing our xml files.

Resolves #1157 

## How has this been tested?

Tested locally. I could be convinced to write a test, but with the refactoring coming up, I think will prefer to use a different method for this. 

## Change log

```
Resolves #1157 objects serialized with utils.data.serialize now will be compressed with bzip2 or gzip depending on file name (.gz and .bz2, respectively)
```

### DIFF
--- a/perses/utils/data.py
+++ b/perses/utils/data.py
@@ -4,6 +4,8 @@ Tools for managing datafiles in perses
 
 """
 
+import logging
+
 __author__ = 'John D. Chodera'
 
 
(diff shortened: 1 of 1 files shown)


ijpulidos (approved):
Oh wow, great catch! I totally missed that when I reviewed it. Thanks! LGTM.

PR #1145, opened 2023-01-18
Add reproducible version of COVID Moonshot examples anyone can run as an example

jchodera:
This PR adds examples from the COVID Moonshot that anyone should be able to run.

This is currently a work in progress, but includes:
* `moonshot-mainseries/` - relative binding free energy calculations primary synthetic series milestones COVID Moonshot 
* `moonshot-sulfonamides/` - relative binding free energy calculations of a congeneric sulfonamide series from the COVID Moonshot from a single structure

PR #1164, opened 2023-02-22
Avoid testing non-examples in CI

ijpulidos:
## Description

From #1145 our examples directory was populated with many scripts that are not necessarily ready-to-run examples.

These changes avoid trying to run these as part of our CI, solving errors on running tests.

## How has this been tested?

Tested locally running `test_examples.py` tests.

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -23,11 +23,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


mikemhenry (approved):
This looks great and easy to maintain!

mikemhenry:
Looks like 3.11 needs numba support still, looks like they are close! https://github.com/numba/numba/issues/8304

PR #1165, opened 2023-02-28
only run dev with newest python version

mikemhenry:
## Description

To save some compute time/carbon, we are only going to test `dev` with the latest python version.

## Motivation and context

Resolves #???

## Change log

```

```

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -23,8 +23,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8, 3.9, "3.10",]
-        openmm: ["8.0", "7.7",]
+        python-version: [3.8, 3.9, "3.10"]
+        openmm: ["8.0", "7.7"]
         include:
           - openmm: "8.0.0dev3"
           - python-version: "3.10"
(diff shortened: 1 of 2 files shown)


ijpulidos:
In this case we are not using the `dev` versions from OpenMM, since we currently rely on the `matrix.openmm` option to install them in https://github.com/choderalab/perses/blob/c7971dbea0162f31a6771a0f35f55ae3efe21552/.github/workflows/CI.yaml#L55

ijpulidos:
I just double checked that this is indeed the case for example in https://github.com/choderalab/perses/actions/runs/4295333336/jobs/7485595343#step:7:236

ijpulidos:
Oh, but according to documentation using `include` should work for this, exactly as you proposed it. Not sure why it isn't working as expected then.

mikemhenry:
I think I fixed it, the problem (I think) was the lack of telling micromamba to reinstall if the package is in the env. It would be faster and better to use this kind of syntax to install openmm:
```
extra-specs: |
  python=${{ matrix.python-version }}
```

but that would then crash on "dev" but would work for the other openmm version strings... but I have an idea!

ijpulidos (approved):
This looks good, thanks!

I think eventually we do want to think about not having to manually change the openmm dev version every time there's a new build.

Issue #1161, opened 2023-02-21
Mapping/parameterization issue


viktorbelay:
I am trying to use the Perses CLI to calculate free energies for nucleotide binding to the IP3 receptor and am using ATP <-> AMP as a test run. I've been able to successfully run Perses for the Tyk2 example (https://github.com/choderalab/perses/tree/main/examples/protein-ligand/cli), but run into an error when trying it for my system:

```
2023-02-06 15:54:36,881:(22.83s):openff.toolkit.typing.engines.smirnoff.parameters:Finding matches for ToolkitAM1BCCHandler
2023-02-06 15:54:36,881:(0.00s):openff.toolkit.typing.engines.smirnoff.parameters:0 matches identified
2023-02-06 15:54:40,299:(3.42s):proposal_generator:determining atom map between old and new molecules...
2023-02-06 15:54:40,299:(0.00s):proposal_generator:the atom map is not specified; proceeding to generate an atom map...
2023-02-06 15:54:40,324:(0.02s):proposal_generator:Using MCSS to derive mapping...
Traceback (most recent call last):
  File "/home/belayv/miniconda3/envs/perses-viktor/bin/perses-cli", line 10, in <module>
    sys.exit(cli())
  File "/home/belayv/miniconda3/envs/perses-viktor/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/home/belayv/miniconda3/envs/perses-viktor/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/home/belayv/miniconda3/envs/perses-viktor/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/belayv/miniconda3/envs/perses-viktor/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/home/belayv/miniconda3/envs/perses-viktor/lib/python3.10/site-packages/perses/app/cli.py", line 108, in cli
    run(yaml_filename=yaml, override_string=override)
  File "/home/belayv/miniconda3/envs/perses-viktor/lib/python3.10/site-packages/perses/app/setup_relative_calculation.py", line 812, in run
    setup_dict = run_setup(setup_options)
[...]

PR #1167, opened 2023-03-01
[DNM] hack to allow undefined stero when mapping

mikemhenry:
Wanted to have this bounce off our CI to make sure this doesn't cause problems elsewhere.


mikemhenry:
See https://github.com/choderalab/perses/issues/1161 for more details

mikemhenry:
@ijpulidos has a PR coming in that will do this in a cleaner way.

PR #1169, opened 2023-03-15
use `False` for reassign velocities in `ProteinMutationExecutor` docstring example.

### LINKED ISSUE #1168: Update docstring example for PointMutationExecutor
Just noticed that the docstring for the PointMutationExecutor contains `reassign_velocities=True`. It should contain `reassign_velocities=False`.

https://github.com/choderalab/perses/blob/e1db2061eec75cceba9f29fa3e3d0a26a8156ed2/perses/app/relative_point_mutation_setup.py#L89-L97

ijpulidos:
## Description

Minor fix in docstring to make the example use `reassign_velocities=False` parameter.

Resolves #1168

## Change log

```
Fix documentation string for `ProteinMutationExecutor`, using `False` for `reassign_velocities` parameter.
```


zhang-ivy:
Thanks for opening this PR. What do you think of keeping the line, but changing it to `False`?  I know `False` is default, but might be good to have it explicitly written for clarity? I don't have a strong opinion on this, so can keep the PR as is if you feel strongly.

ijpulidos:
@zhang-ivy yeah, I guess we can leave the explicit line there, it's a docstring example so it shouldn't be a problem. Thanks for the feedback.

zhang-ivy (approved):

zhang-ivy:
@ijpulidos : Looks good! A couple of the non-OpenMM 8 tests are still failing though?

PR #1170, opened 2023-03-16
[DNM] test new openmmtools version

mikemhenry:
## Motivation and context

Resolves #???

## Change log

```

```


ijpulidos:
This doesn't seem to be doing what we want it to do https://github.com/choderalab/perses/actions/runs/4434092394/jobs/7779743567?pr=1170#step:7:236 Maybe we need to remove the `openmmtools` entry from the `test_env.yaml`  and only leave the `pip` option?

mikemhenry:
It looks okay, I can run the perses tests locally as well with the PR #1065 to double check

mikemhenry:
@ijpulidos you are correct, lets see if faf3e9c8 helps elucidate this

PR #1173, opened 2023-03-27
Support pymbar 4

mikemhenry:
## Motivation and context

Resolves #???

## Change log

```
Added support for pymbar 4 (pymbar 3 is also still supported)
```


mikemhenry:
I will re-run this with pymbar 4 to make sure it still works

mikemhenry:
Pretty sure that the tests we expect to pass are, and the ones we expect to fail are

mikemhenry:
So ready for review @ijpulidos

ijpulidos (commented):
As far as I can tell, there is still one place where we want to change the import

 https://github.com/choderalab/perses/blob/648ad3e11c93261ae6853276ce44a3800c21fb16/perses/tests/test_repex.py#L3

After changing that it seems that this should be ready to go.

mikemhenry:
@ijpulidos good catch! ready for re-reivew

perses/tests/test_repex.py
```diff
@@ -416,7 +416,7 @@ def test_RESTCapableHybridTopologyFactory_repex_charge_transformation():
                 # Extract uncorrelated energy matrix (u_ln) and samples from states (N_l)
                 energy_matrix, samples_per_state = analyzer._compute_mbar_decorrelated_energies()
                 # Compute free energies with boostrapping using pymbar
-                mbar = pymbar.MBAR(energy_matrix, samples_per_state, nbootstraps=200)
+                mbar = _pymbar_bar(energy_matrix, samples_per_state, nbootstraps=200)
```
ijpulidos, line 419: Wait, this is MBAR, I think you had it right from the beginning, this hasn't change in pymbar 4 so we should be using it as we had it before without problems! Apologies for the confusion,

ijpulidos (commented):
We need to change `test_repex.py` back to how it was, sorry!

devtools/conda-envs/test_env.yaml
```diff
@@ -47,3 +47,5 @@ dependencies:
   - sphinx
   - tinydb
   - tqdm
+  - pip:
+    - git+https://github.com/choderalab/openmmtools.git
```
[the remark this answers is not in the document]
mikemhenry: Yes good catch!

mikemhenry, line 51:
```suggestion
```

ijpulidos (approved):
LGTM!

PR #1174, opened 2023-03-28
use python executable from env

mikemhenry:
## Motivation and context

Resolves #???

## Change log

```
Ensure tests use python executable used to invoke tests. 
```


ijpulidos:
As discussed, the problem here is that we end up having an [old version of `ambertools` in our environment (20.9)](https://github.com/choderalab/perses/actions/runs/4550979162/jobs/8024534545?pr=1174#step:7:51), and this version has a vendored old implementation of `parmed`, which seems to be causing the problem for us.

devtools/conda-envs/test_env.yaml
```diff
@@ -30,11 +30,11 @@ dependencies:
   - openmmforcefields >=0.9.0
   - openmmtools >=0.21.5 # may need to sort out ambermini/ambertools/parmed dependencies
   - openmoltools # may need to sort out ambermini/ambertools/parmed dependencies (we don't want ambertools)
-  - parmed # may need to sort out ambermini/ambertools/parmed dependencies
+  - parmed ==3.4.4 # may need to sort out ambermini/ambertools/parmed dependencies
   - pdbfixer
   - pip
   - progressbar2
-  - pymbar <4
+  - pymbar < 4
```
mikemhenry, line 37:
```suggestion
  - pymbar <4
```

ijpulidos (approved):
LGTM!

PR #1176, opened 2023-04-05
print perses version on startup

mikemhenry:
## Motivation and context

Resolves #???

## How has this been tested?

Tested locally, should see output in CI 

## Change log

```
perses-cli now prints version on startup.
```


ijpulidos (approved):
LGTM!

PR #1178, opened 2023-04-07
add caching

mikemhenry:
Going to see if I can get caching to work, this should speed up CI even more.


mikemhenry:
27 seconds now to setup the env when we get a cache hit https://github.com/choderalab/perses/actions/runs/4636885133/jobs/8206700017?pr=1178#step:5:77

ijpulidos (approved):
The times when it hits a cache are great. I checked some of the diffs in the environments compared to our current CI and I don't see differences that should be problematic for us. Great work! LGTM.

PR #1185, opened 2023-04-18
change default to none

### LINKED ISSUE #1179: Make solute-only trajectory writing optional
Typical calculations generate NetCDF files that take up way too much space.

For example, a typical SARS-CoV-2 Mpro calculation will generate a `positions` variable in the complex NetCDF file that consumes (18 replicas) (5000 iterations/replica) (10000 atoms) (3 dimensions/atom) (4 bytes/dimension) = 10GB of data.

Instead, we should default to setting `atom_selection = None` (or perhaps an empty `list`, depending on what the multistate sampler expects) and suggesting our examples remove `atom_selection: not water` from the input YAML.

We can still extract some useful analysis out of the checkpoints as needed.

mikemhenry:
## Motivation and context

Resolves #1179 

## Change log

```

```


mikemhenry:
@ijpulidos feel free to pick up this PR

mikemhenry:
We will want to make sure we have a test where the default is used, or a test where we set the atom selection to none

ijpulidos:
> We will want to make sure we have a test where the default is used, or a test where we set the atom selection to none

In the fixtures for the tests we have the two cases, testing the default (default is now `none` as per discussions in the dev sync), and another one that runs with the `not water` selection.

This does change the default behavior, but I don't think it deserves a new full minor release (it doesn't technically break the API). We should of course document it in the release notes.

ijpulidos (approved):
Looks good to me :)

PR #1186, opened 2023-04-19
Skip failing openmm 8 tests

mikemhenry:
## Motivation and context

Resolves #???

## Change log

```

```


mikemhenry:
Git history looks a bit weird but the diff is correct

mikemhenry:
fixed git history

mikemhenry:
Ah so I need to also skip if it is a dev build:
https://github.com/choderalab/perses/actions/runs/4747449763/jobs/8432372995#step:9:2099

But it looks like an extra test is failing on the dev build...

ijpulidos (approved):
Thanks! Looks good to me. Hopefully we will eventually get the openmm8 subnormals thing worked out and we don't have to skip tests.

lhmzhou (approved):

PR #1188, opened 2023-04-21
Fix RESTTopologyFactory test

zhang-ivy:
## Description

Fixed bug in test that checks to see if the non-rest and REST energy match. Should be checking that the absolute value in the difference in energies is < 1.

## Motivation and context

Resolves #???

## Change log

```

```


ijpulidos (approved):
Looks good to me, thanks!

PR #1197, opened 2023-05-26
Maximum temperature yaml parameter for REST with CLI

### LINKED ISSUE #1195: REST max temperature from CLI
Currently perses REST capabilities are not being used in the small molecule pipeline from CLI. One quick way to do this is passing a configuration option from the YAML file to be used as `t_max` when setting up the sampler in https://github.com/choderalab/perses/blob/53c61b8bfddd6f815d2e4523af5d8c77eba030d8/perses/app/setup_relative_calculation.py#L698

This would allow sampling using REST in the small molecule pipeline, which might be something that we would like to have and experiment with in the near future.

ijpulidos:
## Description

Adds handling of max temperature parameter from input yaml file with CLI.

## Motivation and context

We need to be able to specify the maximum temperature from input YAML file to be able to benefit from using REST in the small molecule simulations pipeline.

Resolves #1195 

## How has this been tested?

Need to come up with a way of testing that REST scaling is being done from the serialized objects. I still don't know what would be the best way to accomplish this.

## Change log

```
Enabling ``max_temperature`` parameter in input yaml file for CLI. This makes REST available for the small molecule pipeline using ``perses-cli``.
```

### DIFF
--- a/perses/annihilation/relative.py
+++ b/perses/annihilation/relative.py
@@ -2833,14 +2833,7 @@ class RESTCapableHybridTopologyFactory(HybridTopologyFactory):
     _new_system_exceptions : dict of key: tuple of ints, value: list of floats
         key: new system indices of the atoms in the exception, value: chargeProd (units of the proton charge squared), sigma (nm), and epsilon (kJ/mol) for the exception
     """
-    
-    # Constants copied from: https://github.com/openmm/openmm/blob/master/platforms/reference/include/SimTKOpenMMRealType.h#L89. These will be imported directly once we have addresssed https://github.com/choderalab/openmmtools/issues/522
-    M_PI = 3.14159265358979323846
-    E_CHARGE = (1.602176634e-19)
-    AVOGADRO = (6.02214076e23)
-    EPSILON0 = (1e-6*8.8541878128e-12/(E_CHARGE*E_CHARGE*AVOGADRO))
-    ONE_4PI_EPS0 = (1/(4*M_PI*EPSILON0))
-    #from openmmtools.constants import ONE_4PI_EPS0
+    from openmmtools.constants import ONE_4PI_EPS0
 
     _default_electrostatics_expression_list = [
 
--- a/examples/new-cli/template.yaml
+++ b/examples/new-cli/template.yaml
@@ -36,6 +36,8 @@ pressure: 1 # atmsopheres
 temperature: 300 # kelvin
 timestep: 4 # femtoseconds
 ionic_strength: 0.15 # molar
+# Max temperature for REST scaling (optional)
+max_temperature: 600 # Kelvin
 
 # Atom mapping specification
 atom_expression:
@@ -78,3 +80,7 @@ phases:
 # Use geometry-derived mapping
 use_given_geometries: true
 given_geometries_tolerance: 0.4 # angstroms
+
+# Realtime analysis frequency
+offline-freq: 100
+
--- a/examples/new-cli/template.yaml
+++ b/examples/new-cli/template.yaml
@@ -84,3 +84,5 @@ given_geometries_tolerance: 0.4 # angstroms
 # Realtime analysis frequency
 offline-freq: 100
 
+# Advanced (optional) -- Specify HTF (HybridTopologyFactory or RESTCapableHybridTopologyFactory)
+# hybrid_topology_factory: RESTCapableHybridTopologyFactory
--- a/perses/app/setup_relative_calculation.py
+++ b/perses/app/setup_relative_calculation.py
@@ -3,7 +3,7 @@ import numpy as np
 import pickle
 import os
 import sys
-import simtk.unit as unit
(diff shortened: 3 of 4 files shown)


ijpulidos:
I figured that a selection of `is_protein` was hardcoded when defining the REST region, but if I try to overcome this I'm just faced with the following issue

```python
  File "/home/user/miniconda3/envs/perses-dev/lib/python3.9/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/home/user/miniconda3/envs/perses-dev/lib/python3.9/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/user/miniconda3/envs/perses-dev/lib/python3.9/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/home/user/workdir/repos/perses/perses/app/cli.py", line 109, in cli
    run(yaml_filename=yaml, override_string=override)
  File "/home/user/workdir/repos/perses/perses/app/setup_relative_calculation.py", line 820, in run
    setup_dict = run_setup(setup_options)
  File "/home/user/workdir/repos/perses/perses/app/setup_relative_calculation.py", line 639, in run_setup
    _validate_endstate_energies_for_htf(htf, top_prop, phase,
  File "/home/user/workdir/repos/perses/perses/app/setup_relative_calculation.py", line 1140, in _validate_endstate_energies_for_htf
    zero_state_error, one_state_error = validate_endstate_energies(topology_proposal,
  File "/home/user/workdir/repos/perses/perses/dispersed/utils.py", line 898, in validate_endstate_energies
    nonalch_zero, nonalch_one, alch_zero, alch_one = generate_endpoint_thermodynamic_states(hybrid_system, top_proposal, repartitioned_endstate)
  File "/home/user/workdir/repos/perses/perses/dispersed/utils.py", line 613, in generate_endpoint_thermodynamic_states
    check_system(system)
  File "/home/user/workdir/repos/perses/perses/dispersed/utils.py", line 64, in check_system
    force = forces['PeriodicTorsionForce']
KeyError: 'PeriodicTorsionForce'
```

For which I don't have a good clue of what could be causing it.

perses/annihilation/relative.py
```diff
@@ -3220,8 +3213,10 @@ def _generate_rest_region(self):
 
         # Retrieve neighboring atoms within self._rest_radius nm of the query atoms
         traj = md.Trajectory(np.array(self._hybrid_positions), self._hybrid_topology)
-        solute_atoms = list(traj.topology.select("is_protein"))
+        solute_atoms = list(traj.topology.select("not water"))
```
zhang-ivy, line 3216: I don't think this is the right way to change this function to work for small molecules. Specifying "not water" will include solvent ions, which we definitely don't want to include as part of the REST region.

We first need to determine what the desired behavior is. The current behavior of this function for small molecules is to automatically include all small molecule atoms in the REST region, and also include protein atoms within X angstroms (where X is user specified) of the small molecule. I think we discussed that you would run some experiments testing this approach, though this will require finding transformations where REST is known to improve sampling / speed up convergence.

One worry I have with this approach is that including all small molecule atoms in the REST region is not the best practices approach (Schrodinger papers typically choose the alchemically-changing (aka unique old and new) atoms and some nearby protein atoms) and there may be risk of small molecule unbinding. Therefore, I think the desired approach should be to modify these lines:

```python

        # Retrieve the residue index of the residue that is being alchemified based on the first unique old atom
        atom_index = list(self._atom_classes['unique_old_atoms'])[0]
        hybrid_topology_atoms = list(self._hybrid_topology.atoms)
        alchemical_residue_index = hybrid_topology_atoms[atom_index].residue.index
        
       # Retrieve indices of all atoms in the alchemical residue
        hybrid_topology_residues = list(self._hybrid_topology.residues)
        mutated_res = hybrid_topology_residues[alchemical_residue_index]
        query_indices = [atom.index for atom in mutated_res.atoms]
        _logger.info(f"Generating rest_region with query indices: {query_indices}")
```

such that for small molecule transformations only, `query_indices` is all unique old/new atoms. This may requiring introducing an argument to the function like `is_small_molecule` that tells the function whether its dealing with a small molecule transformation.

ijpulidos: One of the things that we noticed is that the `solute_atoms` after the selection is empty for `solvent` phase. But we probably want to have the option to do REST scaling in solvent phase as well, such that the neighboring part is not needed, but the molecule atoms are still scaled according to the REST protocol.

zhang-ivy: Notes from discussion: 
- Double check that "is protein" mdtraj selection only contains protein atoms (and does not interpret small molecule atoms as protein atoms). If indeed "is protein" only corresponds to protein atoms, the way that the function is written will fail to include small molecule atoms in the `rest_atoms_all`. Will need to make sure the relevant small molecule atoms are included in this.

PR #1200, opened 2023-06-01
Improve docker building

mikemhenry:
## Motivation and context

Resolves #???

## Change log

```

```


ijpulidos (approved):
This is going to be very useful, thanks! LGTM

ijpulidos:
I see running the examples test is failing. We probably need to find a better way to run/test the examples. When I first created this test it was done in a hacky way.

mikemhenry:
Okay if this passes then I need to remove the trigger that builds on a push to the PR branch and we should be good to merge

mikemhenry:
Ah yes, I forgot I need to make a perses release with the examples removed for this to pass because I use the conda-forge package to create this image.

PR #1201, opened 2023-06-02
drop 3.8, add 3.11

mikemhenry:
## Motivation and context

Resolves #???

## Change log

```

```


mikemhenry:
Will need to refine the matrix to add 3.11 since openmm doesn't have a 3.11 build for 7.7

PR #1202, opened 2023-06-05
Allow specifying solvent model in input YAML file

### LINKED ISSUE #1192: [CLI] no entry point for non-tip3p solvation model
As far as I can tell, there's no way to pass an entry that isn't 'tip3p' to `_solvate_system` in `perses.app.relative_setup.py`:

https://github.com/choderalab/perses/blob/8e3e423a388553fadd9d3f9c0342ffe3e3b4e4c0/perses/app/relative_setup.py#L815

This leads to a template matching `ValueError` when trying to use tip4pew (i.e. `amber/tip4pew_standard.xml`).

ijpulidos:
## Description

These changes allow the specification of the solvent model through the input YAML file for the small molecule pipeline in perses. This is intended to be used via the perses cli.

## Motivation and context

Some users have expressed the desire to use different solvent models for perses simulations.

Resolves #1192 

## Change log

```
Users can now specify solvent model for simulations using the ``solvent_model`` field in the input YAML file.
```

### DIFF
--- a/examples/new-cli/template.yaml
+++ b/examples/new-cli/template.yaml
@@ -78,3 +78,6 @@ phases:
 # Use geometry-derived mapping
 use_given_geometries: true
 given_geometries_tolerance: 0.4 # angstroms
+
+# Solvent model
+solvent_model: "tip3p"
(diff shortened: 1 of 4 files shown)


ijpulidos:
Note that this allows whatever Openmm's `Modeller` allows for solvent model, since it's being directly used by it.

mikemhenry:
Can you add a quick test that uses a different solvent model with an override? It doesn't have to run more than a step

mikemhenry (approved):

PR #1210, opened 2023-07-10
Store initial and final topologies for all phases -- small molecule pipeline

### LINKED ISSUE #1154: Overhaul perses writing of topology information
Currently, perses generates the following topology information on setup:
* `out-complex.pdb` : a single system (old? new?) in complex
* `out-solvent.pdb` : a single system (old? new?) in solvent
* `out-hybrid_factory.npy.npz` : a really slow to deserialize set of way more information than we need to write a trajectory

Additionally, we have two other issues:
* the `HybridTopologyFactory` contains a `htf._hybrid_topology` object that does not contain all bonds for the new atoms
* the [`analysis_particle_indices`](https://github.com/choderalab/perses/blob/e536cfd7f61bdfdd28f315e4b0b559aba5ef5208/perses/app/setup_relative_calculation.py#L660-L670) used to slice out only non-water atoms to write to the NetCDF file has atom indices out of order because it stacks them as `| environment, core, and unique old atoms | unique new atoms | counterions |` while `hybrid_topology` originally has them in the order `| environment, core, and unique old atoms | water | counterions | unique new atoms |`

I propose we restructure this so we have:
```
models/ : organize all PDB files here
  {complex,solvent}-{old,new}.pdb : all atoms 
  {complex,solvent}-solute-{old,new}.pdb : solute atoms 
```
and ensure the atoms in the NetCDF trajectories (checkpoint, standard) are written in the same order as the atoms in the PDB files.
Ideally, we could later write replica trajectories as XTC files directly instead of using the NetCDF file, though extracting coordinates doesn't take a huge amount of time.

We can do this for the new `Protocol` version, where we hopefully have a way to package these files in a more sane way.

ijpulidos:
## Description

These changes make serialization of initial (old) and final (new) topologies for complex and solvent phases for the small molecule pipeline, adding the `RelativeFEPSetup._serialize_topologies` method which is called at the end of the constructor.

Note that while these changes do NOT break the API, it is a change in the behavior since now the stored objects are different.

## Motivation and context

For extracting trajectories for debugging or visualization purposes, it is important to have the topologies available in PDB for the different phases and the initial and final states of the transformation.

Resolves #1154 

## How has this been tested?

Test written, need to check locally that the files make sense.

## Change log

```
Storage of topologies for initial and final states of the transformation, for all phases of the simulation. Useful for extracting trajectories.
```


ijpulidos:
Tested locally and visualizing the different generated PDB files for an example seem to work as expected.

perses/app/relative_setup.py
```diff
@@ -922,6 +922,60 @@ def _handle_charge_changes(self, topology_proposal, new_positions):
...
+                                "positions": old_positions},
+                f"{phase}_new": {"topology": topology_proposal.new_topology,
+                                "positions": new_positions},
+            }
+            # Store in "models" subdir -- create if doesn't exist
+            models_path = f"{self._trajectory_directory}/models"
+            if not os.path.exists(models_path):
+                os.makedirs(models_path)
+            for phase_key, phase_data in topologies_to_store.items():
+                pdb_filename = AnyPath(f"{models_path}/{self._trajectory_prefix}_{phase_key}.pdb")
```
mikemhenry, line 969: Thanks for keeping this an `AnyPath`

perses/app/relative_setup.py
```diff
@@ -922,6 +922,62 @@ def _handle_charge_changes(self, topology_proposal, new_positions):
             # modify the topology proposal
             modify_atom_classes(new_water_indices_to_ionize, topology_proposal)
 
+    def _store_phase_topologies(self, phase: str):
+        """
+        Stores topologies and positions in PDB file for the given the phase, for both the initial (old) and final (new)
+        states. Stores both solvent and solute whenever possible (only "solute" in vacuum).
+
+        Generates two PDB files, one for each state (old/new).
```
zhang-ivy, line 930: Given that the RCSB is phasing out PDB files in favor of mmCIF files, should we allow the user to control which file format is used?

perses/tests/test_relative_setup.py
```diff
@@ -260,47 +258,6 @@ def warn(*args, **kwargs):
                 found_tla = True
         assert found_tla == True, 'Spectator TLA not in old topology'
 
-
-def test_host_guest_deterministic_geometries():
```
zhang-ivy, line 264: Why is this test being deleted?

zhang-ivy:
Left some comments, but otherwise looks good!

perses/tests/test_relative_setup.py
ijpulidos: It basically got replaced by the test at https://github.com/choderalab/perses/blob/499b39ea3b61d4b21377d65343bd7afbe10517bc/perses/tests/test_relative_setup.py#L368

The diff between the two is shown in the following screenshot 
perses_tests_diff

You can see only the forcefield was changed between the tests (since we want to be testing openff rather than gaff, preferably), and the lines to test the storage pdbs were added.

perses/app/relative_setup.py
ijpulidos: I don't have a strong opinion about this, but I would advocate on just sticking with one to avoid unnecessary complexity, as possible. If we feel we should be trying to push for PDBx/mmcif files, then I'm okay with that. Assuming `openmm.PDBFile` can handle that correctly.

ijpulidos: Or `openmm.PDBxFile` in that case, I guess.

zhang-ivy: Ok, lets use PDB files for now, but we will want to keep in mind that CIF files are eventually going to replace PDB files, so may want to open an issue about this (for the repo as a whole, not just the issue)

ijpulidos: Yes, I agree. Created the issue in https://github.com/choderalab/perses/issues/1221

mikemhenry (approved):

PR #1212, opened 2023-07-11
run CI on new 0.10.x branch

mikemhenry:
## Motivation and context

Resolves #???

## Change log

```

```

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -5,7 +5,7 @@ on:
   pull_request:
     branches:
       - "main"
-      - "0.10.x"
+      - "0.10.*"
   schedule:
     # nightly tests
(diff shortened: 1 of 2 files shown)


ijpulidos (approved):
Looks good, thanks!

ijpulidos:
Wait, is it actually `0.10.x` or `0.10.*`?

ijpulidos:
I think it needs to be `0.10.*` instead, check https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpull_requestpull_request_targetbranchesbranches-ignore

ijpulidos (changes requested):
Change requested

mikemhenry:
Ooops, yes typo in the name (should we use x instead of * for the branch name?)

PR #1214, opened 2023-07-20
Remove example testing

mikemhenry:
## Motivation and context

Resolves #???

## Change log

```

```


ijpulidos (approved):
This should be okay, since we are already using the example code in our CI tests. So we are not really losing anything here, as far as I can tell. LGTM! Thanks!

PR #1227, opened 2023-08-22
Realtime analysis interval to default to checkpoint interval

### LINKED ISSUE #1220: Make `offline-freq` be the same as `checkpoint_interval` by default
If not specified in the template file it might be good to make `offline-freq` default to being the same as the `checkpoint_interval` instead of [10](https://github.com/choderalab/perses/blob/ae5580173e321cc85014a2a780080b9b546f18bf/perses/app/setup_relative_calculation.py#L151-L153) as this will cause an error in openmmtools if `checkpoint_interval`  is larger than 10 which most of the time it is.

ijpulidos:
## Description

Changes the default for the `offline-freq` input parameter to match `checkpoint_interval`.

## Motivation and context

Resolves #1220 

## How has this been tested?

Tested locally with CLI. Input yaml file in examples adjusted for it.

## Change log

```
Realtime analysis interval, ``offline-freq``, is now defaulting to ``checkpoint_interval`` when not specified in input parameter file.
```


mikemhenry:
Thank you for this!

mikemhenry (approved):

PR #1230, opened 2023-08-23
Fix/issue 1194

### LINKED ISSUE #1194: [HybridTopologyFactory] `CustomNonbondedForce` cutoff doesn't match `NonBondedForce`
If two endstate systems are built with a non-1.0 nm cutoff, the HybridTopologyFactory will set the `NonBondedForce` to that cutoff value, but will leave `CustomNonbondedForce` to its default value of 1.0.

It's unclear if this is an intended feature or a bug. It looks like RestCapableHybridTopologyFactory does add a cutoff to custom_force, but it would be great to also be able to do this in the base class (if possible).

## How to reproduce

A quick way to reproduce is to manually force an `'nonbondedCutoff': 0.9` entry to the SystemGenerator calls in `relative_setup`.

mikemhenry:
## Motivation and context

Resolves #1194 

## Change log

```
Correctly set cutoff for `CustomNonbondedForce` in `HybridTopologyFactory` when using a non-default cutoff distance.
```

### DIFF
--- a/perses/annihilation/relative.py
+++ b/perses/annihilation/relative.py
@@ -758,22 +758,20 @@ class HybridTopologyFactory(object):
         # Create a CustomNonbondedForce to handle alchemically interpolated nonbonded parameters.
         # Select functional form based on nonbonded method.
         # TODO: check _nonbonded_custom_ewald and _nonbonded_custom_cutoff since they take arguments that are never used...
+        r_cutoff = self._old_system_forces['NonbondedForce'].getCutoffDistance()
+        sterics_energy_expression = self._nonbonded_custom(self._softcore_LJ_v2)
         if self._nonbonded_method in [openmm.NonbondedForce.NoCutoff]:
             _logger.info("\t_add_nonbonded_force_terms: nonbonded_method is NoCutoff")
             sterics_energy_expression = self._nonbonded_custom(self._softcore_LJ_v2)
         elif self._nonbonded_method in [openmm.NonbondedForce.CutoffPeriodic, openmm.NonbondedForce.CutoffNonPeriodic]:
             _logger.info("\t_add_nonbonded_force_terms: nonbonded_method is Cutoff(Periodic or NonPeriodic)")
             epsilon_solvent = self._old_system_forces['NonbondedForce'].getReactionFieldDielectric()
-            r_cutoff = self._old_system_forces['NonbondedForce'].getCutoffDistance()
-            sterics_energy_expression = self._nonbonded_custom(self._softcore_LJ_v2)
             standard_nonbonded_force.setReactionFieldDielectric(epsilon_solvent)
             standard_nonbonded_force.setCutoffDistance(r_cutoff)
         elif self._nonbonded_method in [openmm.NonbondedForce.PME, openmm.NonbondedForce.Ewald]:
             _logger.info("\t_add_nonbonded_force_terms: nonbonded_method is PME or Ewald")
             [alpha_ewald, nx, ny, nz] = self._old_system_forces['NonbondedForce'].getPMEParameters()
             delta = self._old_system_forces['NonbondedForce'].getEwaldErrorTolerance()
-            r_cutoff = self._old_system_forces['NonbondedForce'].getCutoffDistance()
-            sterics_energy_expression = self._nonbonded_custom(self._softcore_LJ_v2)
             standard_nonbonded_force.setPMEParameters(alpha_ewald, nx, ny, nz)
             standard_nonbonded_force.setEwaldErrorTolerance(delta)
             standard_nonbonded_force.setCutoffDistance(r_cutoff)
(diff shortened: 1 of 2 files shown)


ijpulidos (changes requested):
The changes look good.

I think we do require a test for this, I guess it would involve extracting the custom nonbonded forces from the openmm system and manually checking its cutoff distance. 

After that we should also check the effect of this on a full tyk2 benchmark calculation, just to be on the safe side.

perses/annihilation/relative.py
```diff
@@ -800,8 +800,10 @@ def _add_nonbonded_force_terms(self, add_custom_sterics_force=True):
         sterics_custom_nonbonded_force = openmm.CustomNonbondedForce(total_sterics_energy)
         if self._softcore_LJ_v2:
             sterics_custom_nonbonded_force.addGlobalParameter("softcore_alpha", self._softcore_LJ_v2_alpha)
+            sterics_custom_nonbonded_force.setCutoffDistance(self.r_cutoff)
```
ijpulidos, line 803: I just realized that this should probably exist right after line 800 where the object is instantiated in a single statement, instead of being inside the conditional.

ijpulidos: Also, I don't think there's a instance attribute for `r_cutoff`, we probably want to use the one defined in lines 767 or 775, that in turn don't need to exist separately, these could be computed outside that conditional.

ijpulidos: Oh yeah, that's why the CI tests aren't passing https://github.com/choderalab/perses/actions/runs/5955395011/job/16153949168?pr=1230

mikemhenry: yes much better!

mikemhenry (commented):
Looks good! Thanks!

ijpulidos (approved):
LGTM!

ijpulidos:
Actually, I first want to run a tyk2 validation simulation with these changes before merging them

ijpulidos:
@zhang-ivy Since you have worked with the topology factories in detail, can you please take a look into these changes? Just to make sure we didn't miss anything. Thanks!

perses/annihilation/relative.py
```diff
@@ -758,22 +758,20 @@ def _add_nonbonded_force_terms(self, add_custom_sterics_force=True):
...
             standard_nonbonded_force.setCutoffDistance(r_cutoff)
         elif self._nonbonded_method in [openmm.NonbondedForce.PME, openmm.NonbondedForce.Ewald]:
             _logger.info("\t_add_nonbonded_force_terms: nonbonded_method is PME or Ewald")
             [alpha_ewald, nx, ny, nz] = self._old_system_forces['NonbondedForce'].getPMEParameters()
             delta = self._old_system_forces['NonbondedForce'].getEwaldErrorTolerance()
-            r_cutoff = self._old_system_forces['NonbondedForce'].getCutoffDistance()
-            sterics_energy_expression = self._nonbonded_custom(self._softcore_LJ_v2)
             standard_nonbonded_force.setPMEParameters(alpha_ewald, nx, ny, nz)
             standard_nonbonded_force.setEwaldErrorTolerance(delta)
             standard_nonbonded_force.setCutoffDistance(r_cutoff)
```
zhang-ivy, line 777: Noting that the cutoff distance was set properly for the `standard_nonbonded_force` but not the `sterics_custom_nonbonded_force`

perses/tests/test_relative.py
```diff
@@ -1219,3 +1219,38 @@ def test_unsampled_endstate_energies_GPU():
...
+        hybrid_system_forces = hybrid_factory.hybrid_system.getForces()
+        old_nonbonded_forces = [force for force in old_system_forces if isinstance(force, NonbondedForce)]
+        hybrid_custom_nonbonded_forces = [force for force in hybrid_system_forces if isinstance(force, CustomNonbondedForce)]
+        # Modify the cutoff for nonbonded forces in the OLD system (!)
+        for force in old_nonbonded_forces:
+            force.setCutoffDistance(force.getCutoffDistance() + 1*unit.nanometer)
+            # Assert that the nb cutoff distance is different compared to the custom nb forces
+            for custom_force in hybrid_custom_nonbonded_forces:
+                assert custom_force.getCutoffDistance() != \
+                       force.getCutoffDistance(), "Expected different cutoff distances between NB and custom NB forces."
```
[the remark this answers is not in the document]
zhang-ivy: Seems a bit confusing/unnecessary to check each custom force with respect to each standard nonbonded force.

perses/annihilation/relative.py
zhang-ivy: Nothing to change here, just making a note.

zhang-ivy:
Looks good to me! Just added some non-blocking suggestions for the test. Feel free to ignore if you disagree.

mikemhenry:
Thanks!

perses/tests/test_relative.py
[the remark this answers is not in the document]
ijpulidos: I did it in this way to make it general in case we change the test system to something that has more than one nonbonded force and also more than one custom nonbonded forces. The assumption here is that all of them should share the same cutoff.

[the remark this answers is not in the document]
ijpulidos: I understand what you mean that the HTF only has one of them (and should only have one of them), but if in the future we want to inherit from this test class, say for example to test other kind of HTFs that may have more than one of the nonbonded/custom forces, then we can use the same code to test them. I hope that makes sense.

ijpulidos:
Thanks for the review! Waiting for the tyk2 benchmark results to confirm that these changes are not changing anything else along the way.

ijpulidos:
Results seem correct (unaffected by these changes), we could use this for the release as well.

plot_absolute
plot_relative

PR #1235, opened 2023-09-12
0.10.x merge into main

mikemhenry:
## Motivation and context

Resolves #???

## Change log

```

```


mikemhenry:
Getting some interesting errors from pymbar
`E           AttributeError: 'MBAR' object has no attribute 'getFreeEnergyDifferences'`

https://github.com/choderalab/perses/actions/runs/6160980084/job/16719308140#step:9:818

I had thought we fixed that in openmmtools, it is pulling in the latest version

ijpulidos:
@mikemhenry it should be fixed, it should catch the attribute error and handle that in https://github.com/choderalab/openmmtools/blob/main/openmmtools/multistate/multistateanalyzer.py#L1924

mikemhenry:
Looks like the real error was https://github.com/choderalab/perses/actions/runs/6160980084/job/16719308140#step:9:977

ijpulidos:
It seems like we do need to use pymbar 3.1.1 for GPU tests to pass

devtools/conda-envs/test_env.yaml
```diff
@@ -30,11 +30,11 @@ dependencies:
   - openmmforcefields >=0.9.0
   - openmmtools >=0.23 # may need to sort out ambermini/ambertools/parmed dependencies
   - openmoltools # may need to sort out ambermini/ambertools/parmed dependencies (we don't want ambertools)
-  - parmed # may need to sort out ambermini/ambertools/parmed dependencies
+  - parmed
   - pdbfixer
   - pip
   - progressbar2
-  - pymbar <4
+  - pymbar
```
ijpulidos, line 37: Do we want to pin the pymbar version here as well or only in the conda-forge recipe?

ijpulidos (approved):
Just a comment, but not a blocker. This looks good to me!

devtools/conda-envs/test_env.yaml
mikemhenry: I rather do it in the conda-forge recipe since if we don't have a pin in the yaml we use for tests, it is really easy then in CI to make a matrix testing different versions or like the way we do it now, since there is no pin, cpu tests get pymbar 4, but in our GPU CI yaml, we use the extra build args to then pin pymbar<4. If we put it in the env yaml, we would get an error because micomamba doesn't know what to do when the yaml says >3 but the cli command says <4

PR #1237, opened 2023-09-28
Test new open ffmast fix

mikemhenry:
@ijpulidos do you remember which test(s) we disabled because of this bug?

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -77,6 +77,7 @@ jobs:
         shell: bash -l {0}
         run: |
           python -m pip install --no-deps -v .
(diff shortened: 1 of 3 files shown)


ijpulidos:
Yes, I'll make the changes in this branch/PR such that we run the tests that were failing with Openmm8. Thanks!

mikemhenry:
Do we want to merge this in so we can keep these tests running? Part of me is thinking that it could be good to keep these tests running, even if it is using an unreleased beta of the openeye-toolkit.

ijpulidos:
I think it could be worth it to have this in a separate flow (yet another one haha). Is that a reasonable way to do it?

mikemhenry:
We could add it to the matrix (but just do it for one python version and on linux). I am trying to think of the risks of just keeping it as is.

ijpulidos:
I don't see any immediate risks with this. I agree with adding it to the matrix (single python+linux).

PR #1238, opened 2023-10-15
Fix incorrect implementation of logsumexp

badisa:
* Off by a_n.max() for the final value, needs to be added back in after the call to log
* Use the scipy.special.logsumexp implementation (used elsewhere) rather than re-implement

Doesn't look like `MultiTargetDesign` is used, so feel free to close this PR. Just happened to see the bug when reading the code. 

## Description

Noticed that a custom implementation of logsumexp was incorrect, as it didn't add back in `a_n.max()`.

## Motivation and context

Incorrect implementation, will produce non-normalized log probabilities for the class `MultiTargetDesign`

No associated case

## How has this been tested?

Tested by the following python script
```
import numpy as np
from numpy.typing import NDArray


def old_log_sum_exp(vals: NDArray):
    return np.log(np.sum(np.exp(vals - vals.max())))


rng = np.random.default_rng(2023)

log_prob = rng.normal(size=1000, loc=1000)

assert not np.all(np.isclose(np.sum(np.exp(log_prob)), 1.0))

# Verify that the logsumexp from scipy produces a normalized log probability
norm_log = log_prob - logsumexp(log_prob)
print("Sum of probabilities (scipy.special.logsumexp)", np.sum(np.exp(norm_log)))
# Sum of probabilities (scipy.special.logsumexp) 0.9999999999999626
assert np.all(np.isclose(np.sum(np.exp(norm_log)), 1.0))

bad_norm_log = log_prob - old_log_sum_exp(log_prob)
print("Sum of probabilities (old logsumexp)", np.sum(np.exp(bad_norm_log)))
# Sum of probabilities (old logsumexp) inf
assert np.all(np.isclose(np.sum(np.exp(bad_norm_log)), 1.0)), np.sum(
    np.exp(bad_norm_log) 
)  # Fails

```

## Change log

```
Fix log probability normalization in `MultiTargetDesign`
```

### DIFF
--- a/perses/samplers/samplers.py
+++ b/perses/samplers/samplers.py
@@ -18,6 +18,7 @@ __author__ = 'John D. Chodera'
 import mdtraj as md
 import numpy as np
 import time
+from scipy.special import logsumexp
 from openmmtools.states import SamplerState, ThermodynamicState
 
 from perses.annihilation.ncmc_switching import NCMCEngine
@@ -49,7 +50,7 @@ def log_sum_exp(a_n):
 
     """
     a_n = np.array(list(a_n.values()))
-    return np.log( np.sum( np.exp(a_n - a_n.max() ) ) )
+    return logsumexp(a_n)
 
 
 ################################################################################


mikemhenry:
@badisa Thank you for your contribution! Even if we don't end up really using this function, it would be good for it to be correct!

badisa:
> @badisa Thank you for your contribution! Even if we don't end up really using this function, it would be good for it to be correct!

Looks like the tests failed due to being unable to get the OE license, let me know if I need to do anything to resolve that.

mikemhenry:
Yah this is an issue with it being a fork, I will need to make this PR under a branch attached to the main repo, I can do that right now

mikemhenry:
This "tricks" github into running the tests here as well as https://github.com/choderalab/perses/pull/1239 since it is the same commit

mikemhenry:
@ijpulidos This PR has more context, do you mind reviewing this one?

mikemhenry:
The code is the same btw @ijpulidos

ijpulidos (approved):
Thanks for the fix, this looks good to me. I agree with the test and the remark that `MultiTargetDesign` doesn't seem to be used in the code base, so it should be fairly safe to merge this into main.

PR #1239, opened 2023-10-19
Fix incorrect implementation of logsumexp

mikemhenry:
* Off by a_n.max() for the final value, needs to be added back in after the call to log
* Use the scipy.special.logsumexp rather than reimplement

## Motivation and context

Resolves #???

## Change log

```

```


ijpulidos (approved):
LGTM!

PR #1240, opened 2023-11-02
[DNM] test openmm 8.1

mikemhenry:
## Motivation and context

Resolves #???

## Change log

```

```


mikemhenry:
Looks like we don't have any issues with 8.1

PR #1242, opened 2023-11-16
[DNM] Test openmm 8.1 release candidate

ijpulidos:
## Description
PR to test Openmm 8.1 release candidate. Do not merge.

PR #1243, opened 2023-11-16
[DNM] Test/openmm 8.1.0rc1

mikemhenry:


mikemhenry:
GPU test here https://github.com/choderalab/perses/actions/runs/6894390986

ijpulidos:
Thanks! No idea why the changes in #1242 didn't work, I'll be closing that.

With regards the GPU, I don't think the 8.1 rc is getting picked up by it, as per https://github.com/choderalab/perses/actions/runs/6894390986/job/18756206563#step:7:287

mikemhenry:
Haha that is my bad, I didn't even update the yaml file for the gpu jobs :facepalm:

mikemhenry:
Okay it is running on the GPU now https://github.com/choderalab/perses/actions/runs/6896218977/job/18761970375#step:7:314

mikemhenry:
(ignore the CPU tests right now, they passed already so its okay I made a mistake)

mikemhenry:
@ijpulidos okay looks like 2 errors on the GPU tests:
https://github.com/choderalab/perses/actions/runs/6896894039/job/18764058862

One error looks like an API bug, the other though might be real:

```
FAILED perses/tests/test_repex.py::test_RESTCapableHybridTopologyFactory_repex_charge_mutation - pymbar.utils.ParameterError: Warning: Should have \sum_n W_nk = 1. Actual column sum for state 0 was 275.917719. 38 other columns have similar problems. 
This generally indicates the free energies are not converged.
```

image
The tests work on main quite reliably, so we should investigate this IMHO, the first thing I will do is try and re-run the test to see if it happens again

mikemhenry:
re-running things here: https://github.com/choderalab/perses/actions/runs/6896894039

ijpulidos:
@mikemhenry it's actually something we have encountered before https://github.com/choderalab/pymbar/issues/419 I'm still not sure why is this happening, but I think it's only showing with pymbar4

mikemhenry:
We really need to track down that pymbar bug

PR #1249, opened 2025-01-15
Test perses with new versions of core deps

mikemhenry:
## Description

We want to make sure the newest versions of openmm, openmmtools, and pymbar work with perses since the perses atom mapper is used by downstream tools, like `openfe`.

## Motivation and context

Resolves #???

## Change log

```

```


ijpulidos:
We should probably just remove all the reference to `nose` in the tests and see what fails. I suspect `nose` is there just for historical purposes and we are not really using it.

